PR #108, opened 2022-02-06
Enable run on macOS

### LINKED ISSUE #89: [BUG] Unable to run opensearch-benchmark on OSX
**Describe the bug**
I followed https://github.com/opensearch-project/OpenSearch-Benchmark/blob/main/DEVELOPER_GUIDE.md to run the benchmark locally and it failed with following error (logs attached):

```
[ERROR] Cannot execute_test. This test_execution ended with a fatal crash.
```

**To Reproduce**
- Checkout opensearch-benchmark on OSX.
- Setup all the prereq in Developer Guide linked above.
- execute `make prereq` 
- execute `make install`
- `source .venv/bin/activate`
- Run the benchmark test. sample command: ```opensearch-benchmark execute_test --workload=eventdata --target-hosts=endpoint --pipeline=benchmark-only --workload-params="number_of_replicas:1" --client-options="use_ssl:true,basic_auth_user:USERNAME,basic_auth_password:PASSWORD,verify_certs:false" --kill-running-processes```

**Expected behavior**
Benchmark test should execute successfully.

**Logs**
[osx-log.txt](https://github.com/opensearch-project/opensearch-benchmark/files/7717058/osx-log.txt)

**Additional info**
- I verified that there are no listener on 1900 port using `lsof -P -n -r 1 -i :1900`
- I also tried restarting Mac but it didn't help.
- I see the following error in the logs . Not sure if there is any alternative.

 ```
2021-12-15 04:29:48,233 -not-actor-/PID:35422 osbenchmark.actor ERROR Could not initialize internal actor system.
  File "/Volumes/ws/opensearch-benchmark/osbenchmark/actor.py", line 254, in bootstrap_actor_system
    return thespian.actors.ActorSystem(system_base,
  File "/Volumes/ws/opensearch-benchmark/.venv/lib/python3.8/site-packages/thespian/actors.py", line 635, in __init__
    systemBase = self._startupActorSys(
  File "/Volumes/ws/opensearch-benchmark/.venv/lib/python3.8/site-packages/thespian/actors.py", line 676, in _startupActorSys
    systemBase = sbc(self, logDefs=logDefs)
  File "/Volumes/ws/opensearch-benchmark/.venv/lib/python3.8/site-packages/thespian/system/multiprocTCPBase.py", line 28, in __init__
    super(ActorSystemBase, self).__init__(system, logDefs)
  File "/Volumes/ws/opensearch-benchmark/.venv/lib/python3.8/site-packages/thespian/system/multiprocCommon.py", line 86, in __init__
    super(multiprocessCommon, self).__init__(system, logDefs)
  File "/Volumes/ws/opensearch-benchmark/.venv/lib/python3.8/site-packages/thespian/system/systemBase.py", line 326, in __init__
    self._startAdmin(self.adminAddr,
  File "/Volumes/ws/opensearch-benchmark/.venv/lib/python3.8/site-packages/thespian/system/multiprocCommon.py", line 114, in _startAdmin
    raise InvalidActorAddress(adminAddr,
thespian.actors.InvalidActorAddress: ActorAddr-(T|:1900) is not a valid ActorSystem admin```

sharp-pixel:
### Description
Enables run on macOS.

Note 1: due to OpenSearch 1.0.1 not being available for Apple Silicon, the integration tests fail on a Mac M1. There is not much that I can do here.

Note 2: this PR also fixes the templates to run on Amazon OpenSearch Service with OpenSearch 1.1, as the _doc is now an error.
 
### Issues Resolved
#89 
 
### Check List


dblock:
Let's also add Darwin to CI?

sharp-pixel:
Anything I can do from my side?

dblock:
Yes @sharp-pixel! Add `macos-latest` to https://github.com/opensearch-project/opensearch-benchmark/blob/main/.github/workflows/main.yml as part of this PR?

sharp-pixel:
Added

.github/workflows/main.yml
```diff
@@ -12,6 +12,17 @@ jobs:
         run: bash .ci/build.sh build
         env:
           BENCHMARK_HOME: env.GITHUB_WORKSPACE
+  Unit-Tests-MacOS:
```
dblock, line 15: Since this is identical to the above, can we use a matrix? Example: https://github.com/opensearch-project/opensearch-build/blob/main/.github/workflows/python-tests.yml#L8

dblock (changes requested):
How about this?

.github/workflows/main.yml
sharp-pixel: Thanks for pointing it out, I corrected the config

sharp-pixel:
Hum, time based unit tests... interesting.
Python time.sleep's implementation on macOS seems not to be accurate enough: https://stackoverflow.com/questions/1133857/how-accurate-is-pythons-time-sleep
Also these tests would be highly dependent of the underlying system.
Now the problem is that is the system's time.sleep is not accurate enough and this is used for the actual benchmarking then the results would be unreliable, and it would be impossible to compare benchmarks run from different environments.

A suggestion would be to remove these tests on the build system, but run them at start of an actual benchmark to validate that the system running it is within accuracy limits. Open for discussion.

dblock:
Generally, I recommend disabling unpredictable/unreliable tests and opening an issue to replace/fix them to be 100% reliable.

engechas (changes requested):
Based on the mac CI unit tests, there's other tests that are time specific and will need to be disabled as well

FAILED tests/client_test.py::RequestContextManagerTests::test_propagates_nested_context
FAILED tests/worker_coordinator/runner_test.py::CompositeTests::test_adds_request_timings

Rest LGTM

engechas (dismissed):

weifenh (approved):

sharp-pixel:
The `_doc` will be a problem for Elasticsearch < v6.7, and between 6.7 and 7.x, we would need to set `include_type_name=false` explicitely. This is the default after Elasticsearch 7.0 and for OpenSearch.

sharp-pixel:
I we want to still support older versions, we would need two versions of the templates, check version and use the correct ones.

engechas:
It looks like `include_type_name` based on version was accounted for [here](https://github.com/opensearch-project/opensearch-benchmark/blob/ba96b9a262a303a739e39296ec676e6908c21905/osbenchmark/metrics.py#L68-L72) but doesn't work with OpenSearch metrics stores since the version is reported as 1.

I see two options: 
1. Remove `_doc`, forcing metrics stores to be ES7+ or OpenSearch
2. Refactor the logic [here](https://github.com/opensearch-project/opensearch-benchmark/blob/ba96b9a262a303a739e39296ec676e6908c21905/osbenchmark/metrics.py#L68-L72) to send `include_type_name=true` for OpenSearch metrics stores

Either option seems fine to me. Curious to hear others' thoughts

sharp-pixel:
> It looks like `include_type_name` based on version was accounted for [here](https://github.com/opensearch-project/opensearch-benchmark/blob/ba96b9a262a303a739e39296ec676e6908c21905/osbenchmark/metrics.py#L68-L72) but doesn't work with OpenSearch metrics stores since the version is reported as 1.
> 
> I see two options:
> 
>     1. Remove `_doc`, forcing metrics stores to be ES7+ or OpenSearch
> 
>     2. Refactor the logic [here](https://github.com/opensearch-project/opensearch-benchmark/blob/ba96b9a262a303a739e39296ec676e6908c21905/osbenchmark/metrics.py#L68-L72) to send `include_type_name=true` for OpenSearch metrics stores
> 
> 
> Either option seems fine to me. Curious to hear others' thoughts

I made some changes:
- I restored the `_doc` in the templates for now.
- I added a new `_cluster_distribution` field in `OsClient` that holds the distribution (either 'opensearch' or 'elasticsearch'), that way we can differentiate versions for OpenSearch from versions for Elasticsearch.

osbenchmark/metrics.py
```diff
@@ -52,20 +52,23 @@ def __init__(self, client, cluster_version=None):
...
         self.logger = logging.getLogger(__name__)
         self._cluster_version = cluster_version
+        self._cluster_distribution = None
 
     # TODO #653: Remove version-specific support for metrics stores before 7.0.0.
     def probe_version(self):
         info = self.guarded(self._client.info)
         try:
             self._cluster_version = versions.components(info["version"]["number"])
+            self._cluster_distribution = info["version"]["distribution"]
```
engechas, line 62: If I curl an ES cluster it doesn't return a distribution in the version
```
"version" : {
    "number" : "7.7.0",
```
but it looks like all OpenSearch clusters do (tested Amazon OpenSearch Service and self-managed)
```
"version" : {
    "distribution" : "opensearch",
    "number" : "1.1.0",
```

The code as is will throw a KeyError for Elasticsearch clusters, can we refactor to the following?
```
def probe_version(self):
    info = self.guarded(self._client.info)
    try:
        self._cluster_version = versions.components(info["version"]["number"])
    except BaseException:
        msg = "Could not determine version of metrics cluster"
        self.logger.exception(msg)
        raise exceptions.BenchmarkError(msg)

    try:
        self._cluster_distribution = info["version"]["distribution"]
    except BaseException:
        msg = "Could not determine distribution of metrics cluster, assuming elasticsearch"
        self.logger.exception(msg)
        self._cluster_distribution = "elasticsearch"
```

osbenchmark/metrics.py
```diff
@@ -52,20 +52,23 @@ def __init__(self, client, cluster_version=None):
...
         except BaseException:
             msg = "Could not determine version of metrics cluster"
             self.logger.exception(msg)
             raise exceptions.BenchmarkError(msg)
 
     def put_template(self, name, template):
         # TODO #653: Remove version-specific support for metrics stores before 7.0.0 (also adjust template)
-        if self._cluster_version[0] > 6:
+        if (self._cluster_version[0] > 6 and self._cluster_distribution == 'elasticsearch') or \
+                self._cluster_distribution == 'opensearch':
```
engechas, line 71: Slight nit: single vs double quotes. The rest of this file seems to use double quotes

engechas (changes requested):

osbenchmark/metrics.py
sharp-pixel: I missed that. Just pushed the correction

osbenchmark/metrics.py
sharp-pixel: Corrected

engechas (approved):
LGTM

Integration tests using an OpenSearch metrics store with these changes and the changes in #119 are passing

engechas:
I have created followup issues to break the DCO out into it's own file and to re-evalutate the disabled unit tests: #120 #121

Merging this PR, thanks!

PR #130, opened 2022-02-25
Small fixes to get the integration tests working on Github Actions, adding the tests to pull_requests

### LINKED ISSUE #129: Add Integration Test Workflow to Github Actions
Similar to the unit tests, the integration tests should be run as a Github Actions workflow on pull requests.

engechas:
### Description
This PR fills in the gaps to get the integration tests running on Github Actions. This required installing JDK16, pulling the squid container, and updating the `BENCHMARK_HOME` environment variable.

The Github Actions integration tests run in 2 parallel streams, one for python 3.8 and one for python 3.9. The unit tests have been removed from the tox.ini commands since they are run in a separate Github Actions workflow.

Potential further enhancements:
1. The current containers use Ubuntu, we may want to add a MacOS iteration as well
2. Certain integration tests [pick randomly](https://github.com/opensearch-project/opensearch-benchmark/blob/main/it/distribution_test.py#L28) between using an in-memory or OpenSearch metrics store. This can lead to different results on different runs, even with the same code. IMO we should remove the randomness and pick one or the other types of metrics stores for all currently randomized tests
3. Coverage tool to see how much of the code is actually being checked by these tests
4. Tests for ES metrics stores (large effort, contingent on the provisioning expansion)
5. Add integration tests to the Workloads repo (a simple solution could be each test/challenge run once in `test-mode`)

### Issues Resolved
#129 
 
### Check List

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -67,12 +67,13 @@ function build_it {
   export PATH="$HOME/.pyenv/bin:$PATH"
   export TERM=dumb
   export LC_ALL=en_US.UTF-8
-  export BENCHMARK_HOME=$WORKSPACE
+  export BENCHMARK_HOME="$GITHUB_WORKSPACE"
   export JAVA_PATH="/opt/hostedtoolcache/Java_Adopt_jdk"
   export JAVA_HOME="$JAVA_PATH/15.0.2-7/x64"
   export JAVA8_HOME="$JAVA_PATH/8.0.292-1/x64"
   export JAVA11_HOME="$JAVA_PATH/11.0.11-9/x64"
   export JAVA15_HOME="$JAVA_PATH/15.0.2-7/x64"
+  export JAVA16_HOME="$JAVA_PATH/16.0.2-7/x64"
 
   update_pyenv
   eval "$(pyenv init -)"
@@ -80,11 +81,17 @@ function build_it {
   eval "$(pyenv init --path)"
   eval "$(pyenv virtualenv-init -)"
   pip install opensearch-benchmark
+  docker pull datadog/squid:latest
 
   make prereq
   make install
   make precommit
-  make it
+
+  if [[ "$1" == "Python_3.8" ]]; then
+    make it38
+  elif [[ "$1" == "Python_3.9" ]]; then
+    make it39
+  fi
 }
 
 function license-scan {
@@ -118,7 +125,7 @@ function archive {
 }
 
 if declare -F "$1" > /dev/null; then
-    $1
+    $1 $2
     exit
 else
     echo "Please specify a function to run"
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -125,11 +125,7 @@ function archive {
 }
 
 if declare -F "$1" > /dev/null; then
-    if [ -n "${2-}" ]; then
-      $1 $2
-    else
-      $1
-    fi
+    "$@"
     exit
 else
     echo "Please specify a function to run"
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -38,7 +38,6 @@ setenv =
     # applications behave identically, we also set LANG explicitly.
```
travisbenedict, line 38: Not related to your changes but while you're in here can you you also change `OSBenchmark` to `OpenSearch Benchmark` on line 3. 

I struggled to put this comment on the line I actually wanted...

travisbenedict (approved):
This is great!

tox.ini
engechas: Sure will do once the current run of integ tests finishes (pushing a new commit resets them)

.ci/build.sh
```diff
@@ -118,7 +125,11 @@ function archive {
 }
 
 if declare -F "$1" > /dev/null; then
-    $1
+    if [ -n "${2-}" ]; then
```
travisbenedict, line 128: Can you explain what this is for?

engechas: On [line 25](https://github.com/opensearch-project/opensearch-benchmark/blob/main/.ci/build.sh#L25) `set -u` is run which causes the script to throw an error when a variable is referenced but unset. The unit tests are run with `./build.sh build` whereas the integ tests are run with `./build build_it <py version>`. So for unit tests, `$2` is unset and throws an exception.

`${2-}` gets the value of $2 if set, otherwise returns an empty value. `-n` checks if the provided value is non-null/empty/0.

It's essentially a hacky way to check if a `$2` parameter is set without throwing an exception. This file should be holistically re-evaluated IMO as it's a bit messy (for example, we run the opposite of `set -u` [in](https://github.com/opensearch-project/opensearch-benchmark/blob/main/.ci/build.sh#L91-L92) [multiple](https://github.com/opensearch-project/opensearch-benchmark/blob/main/.ci/build.sh#L39-L40) [functions](https://github.com/opensearch-project/opensearch-benchmark/blob/main/.ci/build.sh#L64-L65)).

gkamat: You can replace both the cases with a single `"$@"` and eliminate the `if` statement.

.ci/build.sh
```diff
@@ -67,24 +67,31 @@ function build_it {
...
   export JAVA15_HOME="$JAVA_PATH/15.0.2-7/x64"
+  export JAVA16_HOME="$JAVA_PATH/16.0.2-7/x64"
 
   update_pyenv
   eval "$(pyenv init -)"
   # ensure pyenv shims are added to PATH, see https://github.com/pyenv/pyenv/issues/1906
   eval "$(pyenv init --path)"
   eval "$(pyenv virtualenv-init -)"
   pip install opensearch-benchmark
+  docker pull datadog/squid:latest
```
gkamat, line 84: Where does this get used?

gkamat (approved):

.ci/build.sh
engechas: That is much cleaner and scales without modifications if we were to add a second parameter to one of the functions. Will change, thanks!

.ci/build.sh
engechas: It is used in the integration tests for testing a proxy user. I am not entirely sure how that fits into the larger scope of OSB

https://github.com/opensearch-project/opensearch-benchmark/blob/main/it/proxy_test.py#L39

Issue #134, opened 2022-02-28
[Builder] Migrate Existing Builder Component Implementations


engechas:
Migrate the existing builder system components to conform to the new interfaces

Parent Issue: #113

Issue #113, opened 2022-02-21
[RFC] System Under Test Provisioning Expansion


engechas:
# Background
OpenSearch Benchmark supports creating a self-managed OpenSearch cluster on the same machine running OpenSearch Benchmark. Clusters can be created by building OpenSearch from source, downloading a distribution, or using a docker image. A variety of cluster settings can be specified such as heap size, JVM flags, and the garbage collection algorithm.

# Proposal
OpenSearch Benchmark’s system under test provisioning capabilities will be expanded to support cloud infrastructure creation. This will allow clusters to be provisioned on infrastructure external to the machine running OpenSearch Benchmark. 

Implementations for self-managed OpenSearch running on EC2 instances and managed Amazon OpenSearch Service #100 are in scope for this project. The existing functionality to provision a cluster on the local machine will continue to be supported.

## Design
There will be no change to the existing `execute_test` workflow. The BuilderActor will continue to be invoked as the first step of the BenchmarkActor’s workflow. The WorkerCoordinatorActor will not receive a request to start the test until the BuilderActor has notified the BenchmarkActor that the cluster is ready. Once the test has completed, the BuilderActor will receive a request to delete the cluster from the BenchmarkActor.

![image (2)](https://user-images.githubusercontent.com/62891993/155000330-9e3f9b30-8e86-4d21-9845-01421430dec8.png)

The Builder System will consist of the core logic required to provision a local cluster running self-managed OpenSearch. Additional plugins can be defined to provision cloud infrastructure when creating a cluster. Eventually these additional plugins will be moved to a separate repository and fetched via a PluginLoader #98. In the short term, any infrastructure-provider plugins will be implemented as modules within OpenSearch Benchmark’s Builder System.

![image (3)](https://user-images.githubusercontent.com/62891993/155002253-21bcb77d-e155-44be-bd50-ea6e84ec137f.png)

The existing NodeBuilderActor class will be removed in favor of handling the provisioning logic within the Core Builder. The Dispatcher will remain an intermediary step between the builder plugins and the BuilderActor in order to prevent the CreateCluster call from blocking the BuilderActor while provisioning is underway. Based on the user input, the Core Builder will determine if a provisioning plugin is required and load it for use by the Builder System.

### Builder System
The Builder System is responsible for orchestrating the logic to create and delete a system under test. It consists of multiple components that are executed sequentially to perform the provisioning logic. The top-level class and the interface into the Builder System is a Builder. 

[...]

PR #162, opened 2022-03-10
Migrate DockerLauncher logic to conform to the new Launcher interface

engechas:
### Description
This PR is part of the larger changes for #134 

Please see #113 for a class diagram of how this PR fits into the larger builder system. The plan for migrating the existing components is to duplicate the logic initially, then remove the initial logic and wire the new subcomponents into the builder system once everything is ready.

This PR adds 2 new implementations of the Launcher interface. The first is a NoOpLauncher, which performs no action when  its methods are invoked. The second is a DockerLauncher, which is a migration of the [existing class here](https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/builder/launcher.py#L37-L105). The logic is a copy/paste with the following modifications:
* Update the public methods to include a `host` parameter
* Add `shell_executor` to the constructor for executing shell commands
* Replace instances of using `process.xyz` with `shell_executor.execute`

Additionally, the unit tests for the original DockerLauncher were copy pasted and reformatted to conform to the new DockerLauncher. For further testing, I wired the new DockerLauncher into the Builder and ran the `test_docker_distribution` test which passed.
 
### Check List
- [ ] Commits are signed per the DCO using --signoff 

### DIFF
--- a/osbenchmark/builder/launchers/docker_launcher.py
+++ b/osbenchmark/builder/launchers/docker_launcher.py
@@ -1,0 +1,75 @@
+import os
+
+from osbenchmark import time, telemetry
+from osbenchmark.builder import cluster
+from osbenchmark.builder.launchers.launcher import Launcher
+from osbenchmark.exceptions import LaunchError, ExecutorError
+
+
+class DockerLauncher(Launcher):
+    # May download a Docker image and that can take some time
+    PROCESS_WAIT_TIMEOUT_SECONDS = 10 * 60
+
+    def __init__(self, pci, executor, clock=time.Clock):
+        super().__init__(executor)
+        self.clock = clock
+
+    def start(self, host, node_configurations):
+        nodes = []
+        for node_configuration in node_configurations:
+            node_name = node_configuration.node_name
+            host_name = node_configuration.ip
+            binary_path = node_configuration.binary_path
+            self.logger.info("Starting node [%s] in Docker.", node_name)
+            self._start_process(host, binary_path)
+            node_telemetry = [
+                # Don't attach any telemetry devices for now but keep the infrastructure in place
+            ]
+            t = telemetry.Telemetry(devices=node_telemetry)
+            node = cluster.Node(0, binary_path, host_name, node_name, t)
+            t.attach_to_node(node)
+            nodes.append(node)
+        return nodes
+
+    def _start_process(self, host, binary_path):
+        compose_cmd = self._docker_compose(binary_path, "up -d")
+
+        try:
+            self.executor.execute(host, compose_cmd)
+        except ExecutorError as e:
+            raise LaunchError("Exception starting OpenSearch Docker container", e)
+
+        container_id = self._get_container_id(binary_path)
+        self._wait_for_healthy_running_container(container_id, DockerLauncher.PROCESS_WAIT_TIMEOUT_SECONDS)
+
+    def _docker_compose(self, compose_config, cmd):
+        return "docker-compose -f {} {}".format(os.path.join(compose_config, "docker-compose.yml"), cmd)
+
+    def _get_container_id(self, compose_config):
+        compose_ps_cmd = self._docker_compose(compose_config, "ps -q")
+        return self.executor.execute(compose_ps_cmd, ouput=True)[0]
+
+    def _wait_for_healthy_running_container(self, container_id, timeout):
+        cmd = 'docker ps -a --filter "id={}" --filter "status=running" --filter "health=healthy" -q'.format(container_id)
+        stop_watch = self.clock.stop_watch()
+        stop_watch.start()
+        while stop_watch.split_time() < timeout:
+            containers = self.executor.execute(cmd, ouput=True)
+            if len(containers) > 0:
+                return
+            time.sleep(0.5)
+        msg = "No healthy running container after {} seconds!".format(timeout)
+        self.logger.error(msg)
+        raise LaunchError(msg)
+
+    def stop(self, host, nodes, metrics_store):
+        self.logger.info("Shutting down [%d] nodes running in Docker on this host.", len(nodes))
+        for node in nodes:
+            self.logger.info("Stopping node [%s].", node.node_name)
+            if metrics_store:
+                telemetry.add_metadata_for_node(metrics_store, node.node_name, node.host_name)
+            node.telemetry.detach_from_node(node, running=True)
+            self.executor.execute(host, self._docker_compose(node.binary_path, "down"))
+            node.telemetry.detach_from_node(node, running=False)
+            if metrics_store:
+                node.telemetry.store_system_metrics(node, metrics_store)
--- a/osbenchmark/builder/launchers/docker_launcher.py
+++ b/osbenchmark/builder/launchers/docker_launcher.py
@@ -1,3 +1,4 @@
+import logging
 import os
 
 from osbenchmark import time, telemetry
@@ -12,6 +13,7 @@ class DockerLauncher(Launcher):
 
     def __init__(self, pci, executor, clock=time.Clock):
         super().__init__(executor)
+        self.logger = logging.getLogger(__name__)
         self.clock = clock
 
     def start(self, host, node_configurations):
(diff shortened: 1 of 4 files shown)


osbenchmark/builder/launchers/docker_launcher.py
```diff
@@ -0,0 +1,77 @@
...
+        try:
+            self.shell_executor.execute(host, compose_cmd)
+        except ExecutorError as e:
+            raise LaunchError("Exception starting OpenSearch Docker container", e)
+
+        container_id = self._get_container_id(host, binary_path)
+        self._wait_for_healthy_running_container(host, container_id, DockerLauncher.PROCESS_WAIT_TIMEOUT_SECONDS)
+
+    def _docker_compose(self, compose_config, cmd):
+        return "docker-compose -f {} {}".format(os.path.join(compose_config, "docker-compose.yml"), cmd)
```
IanHoang, line 48: Nit: ```.format()``` works fine but I think it becomes quite verbose when working with multiple parameters such as this line. Even if this is pre-existing code, it might be worth switching areas to use f-strings and f-strings from here on out for newer snippets of code. f-strings would enhance readability, flexibility, and _slight_ speed over %s and .format(). Again, not necessary, but I think it's worth using f-strings from here on out but I understand if we want to maintain consistency by using .format()

engechas: I didn't put much time into refactoring inherited code unless it was necessary. I am in agreement that much of it could be cleaned up so I'm happy to refactor stuff like this as folks point it out. 

Will refactor this instance

osbenchmark/builder/launchers/docker_launcher.py
```diff
@@ -0,0 +1,77 @@
...
+        self.logger.info("Shutting down [%d] nodes running in Docker on this host.", len(nodes))
+        for node in nodes:
+            self.logger.info("Stopping node [%s].", node.node_name)
+            if metrics_store:
+                telemetry.add_metadata_for_node(metrics_store, node.node_name, node.host_name)
+            node.telemetry.detach_from_node(node, running=True)
+            self.shell_executor.execute(host, self._docker_compose(node.binary_path, "down"))
+            node.telemetry.detach_from_node(node, running=False)
+            if metrics_store:
+                node.telemetry.store_system_metrics(node, metrics_store)
```
IanHoang, line 82: Is this condensable? Might not be worth touching since order probably matters here but if it doesn't, we could move line 72 under line 76.

engechas: I believe order does matter here as the metadata for the node must be processed (Line 72) before detaching the telemetry devices (Lines 73 and 75)

osbenchmark/builder/launchers/docker_launcher.py
```diff
@@ -0,0 +1,82 @@
...
+        raise LaunchError(msg)
+
+    def stop(self, host, nodes, metrics_store):
+        self.logger.info("Shutting down [%d] nodes running in Docker on this host.", len(nodes))
+        for node in nodes:
+            self.logger.info("Stopping node [%s].", node.node_name)
+            if metrics_store:
+                telemetry.add_metadata_for_node(metrics_store, node.node_name, node.host_name)
+            node.telemetry.detach_from_node(node, running=True)
+            self.shell_executor.execute(host, self._docker_compose(node.binary_path, "down"))
```
achitojha, line 79: It seems like we're handling the exception while starting the docker and wrapping it in `LaunchError`, should we do something similar for tear down?

osbenchmark/builder/launchers/launcher.py
```diff
@@ -15,12 +15,13 @@ def start(self, host, node_configurations):
         """
         raise NotImplementedError
 
-    def stop(self, host, nodes):
+    def stop(self, host, nodes, metrics_store):
```
achitojha, line 18: This interface looks confusing to me. As a user of this interface I'm wondering the interface/class asks me to provide metrics_store only for the stop method?

Is there a reason why we shouldn't have the metrics_store as a class level instance. My understanding is that it can't vary with time - so once this class is instantiated, the metrics_store can just live with it all through its lifetime.

osbenchmark/builder/launchers/docker_launcher.py
```diff
@@ -0,0 +1,82 @@
...
+            node_name = node_configuration.node_name
+            host_name = node_configuration.ip
+            binary_path = node_configuration.binary_path
+            self.logger.info("Starting node [%s] in Docker.", node_name)
+            self._start_process(host, binary_path)
+            node_telemetry = [
+                # Don't attach any telemetry devices for now but keep the infrastructure in place
+            ]
+            t = telemetry.Telemetry(devices=node_telemetry)
+            node = cluster.Node(0, binary_path, host_name, node_name, t)
```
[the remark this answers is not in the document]
engechas: It's essentially a placeholder for the DockerLauncher implementation as managing the OpenSearch process is done using docker commands. In the ProcessLauncher, this is used for killing the OpenSearch process after a test is finished. `0` here seems like a bad solution as it's a valid PID, I will change it to `None` to make it clear that it's unused for the DockerLauncher implementation

osbenchmark/builder/launchers/docker_launcher.py
```diff
@@ -0,0 +1,82 @@
...
+    def _wait_for_healthy_running_container(self, host, container_id, timeout):
+        cmd = 'docker ps -a --filter "id={}" --filter "status=running" --filter "health=healthy" -q'.format(container_id)
+        stop_watch = self.clock.stop_watch()
+        stop_watch.start()
+        while stop_watch.split_time() < timeout:
+            containers = self.shell_executor.execute(host, cmd, output=True)
+            if len(containers) > 0:
+                return
+            time.sleep(0.5)
+        msg = "No healthy running container after {} seconds!".format(timeout)
```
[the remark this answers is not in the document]
engechas: I like that, will refactor

osbenchmark/builder/launchers/docker_launcher.py
engechas: I was planning to use a decorator to wrap all of these calls, similar to ExceptionHandlingShellExecutor

osbenchmark/builder/launchers/launcher.py
engechas: `metrics_store` being a parameter always struck me as odd as well. I like the idea to add it to the constructor, will refactor

osbenchmark/builder/launchers/docker_launcher.py
```diff
@@ -0,0 +1,82 @@
...
+import os
+
+from osbenchmark import time, telemetry
+from osbenchmark.builder import cluster
+from osbenchmark.builder.launchers.launcher import Launcher
+from osbenchmark.exceptions import LaunchError, ExecutorError
+
+
+class DockerLauncher(Launcher):
+    # May download a Docker image and that can take some time
```
weifenh, line 11: Might worth adding a 1-line documentation about ``DockerLauncher``, like the ``ProcessLauncher`` below.

osbenchmark/builder/launchers/docker_launcher.py
engechas: That's a good callout, will remove the try/catch on shell executor calls in favor of handling at the top level in the decorator

osbenchmark/builder/launchers/docker_launcher.py
engechas: I'm not sure that I agree with documenting interface implementations. It adds burden when the implementation changes. I'd rather rely on documentation in the interface definition that users can refer to for more context, then have the code in the implementation self-document with well written variable and method names. 

For this case in particular, I think DockerLauncher is relatively self-explanatory

achitojha (approved):

osbenchmark/utils/periodic_waiter.py
```diff
@@ -0,0 +1,19 @@
+from osbenchmark import time
+
+
+class PeriodicWaiter:
```
travisbenedict, line 4: Minor: This seems more like a `Poller` than a `Waiter` to me

travisbenedict (approved):

Issue #132, opened 2022-02-28
[Builder] Create/Refactor Builder Models


engechas:
Refactor the existing builder system models and create new ones to fill in the gaps for supporting the build system expansion.

Parent Issue: #113

PR #163, opened 2022-03-17
Add Provider and Flavor enums, populate their default values in ProvisionConfigInstance

engechas:
### Description
Adds the new Provider and Flavor models for the build system expansion.
#132 

### Check List

### DIFF
--- a/osbenchmark/builder/models/cluster_flavors.py
+++ b/osbenchmark/builder/models/cluster_flavors.py
@@ -1,6 +1,6 @@
 from enum import Enum
 
 
-class Flavor(str, Enum):
+class ClusterFlavor(str, Enum):
     SELF_MANAGED = "self_managed"
     MANAGED = "managed"
--- a/osbenchmark/builder/models/flavors.py
+++ b/osbenchmark/builder/models/flavors.py
@@ -1,0 +1,6 @@
+from enum import Enum
+
+
+class Flavor(str, Enum):
+    SELF_MANAGED = "self_managed"
+    MANAGED = "managed"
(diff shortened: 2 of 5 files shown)


osbenchmark/builder/models/providers.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum
+
+
+class Provider(str, Enum):
```
travisbenedict, line 4: How do you feel about calling this something like `InfraProvider` or `ClusterInfraProvider` instead?

osbenchmark/builder/models/flavors.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum
+
+
+class Flavor(str, Enum):
```
travisbenedict, line 4: Thoughts on calling this `ClusterFlavor` instead?

osbenchmark/builder/models/providers.py
engechas: That's more clear, will change

osbenchmark/builder/models/flavors.py
engechas: That's more clear, will change

travisbenedict (approved):

osbenchmark/builder/models/cluster_flavors.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum
+
+
+class ClusterFlavor(str, Enum):
+    SELF_MANAGED = "self_managed"
+    MANAGED = "managed"
```
achitojha, line 6: Do we need two different enums for ClusterFlavor & ClusterInfraProvider? Can you explain what's the use case supporting the need for two enums.

I was thinking of combining them into one. The `MANAGED` flavor seems too generic for me, similarly `LOCAL` provider probably doesn't add much value.

IanHoang (approved):
LGTM

PR #164, opened 2022-03-18
Create LocalProcessLauncher and ExceptionHandlingLauncher

engechas:
### Description
Migrates the ProcessLauncher to a LocalProcessLauncher. Adds an ExceptionHandlingLauncher to decorate the Launcher implementations with exception handling.

I initially tried to generalize the LocalProcessLauncher to use shell commands instead of helpers such as `io` or `os` modules. This would have allowed the implementation to be used by non-local providers. However, the telemetry devices and `java_resolver` are tightly coupled to evaluating the local machine as well and would have required significant refactoring.

Rather than going this route, I decided that a separate implementation for ExternalProcessLauncher could be created to handle the generalized flow of this class without coupling the logic to the local machine. This seems to the be quickest and cleanest path forwards. If there is shared logic between these implementations, it will be broken out to avoid duplication.

The LocalProcessLauncher is mostly unchanged from the original ProcessLauncher, with the exception of using the executor instead of the previous `subprocess.Popen` call. Additionally, I refactored a few of the methods to make them more modular.

The LocalProcessLauncher was tested by migrating the unit tests from the original ProcessLauncher. Additionally, I wired it into the builder system and ran the `test_tar_distributions` and `test_sources` integration tests.
 
### Issues Resolved
#134 
 
### Check List

### DIFF
--- a/osbenchmark/builder/launchers/local_process_launcher.py
+++ b/osbenchmark/builder/launchers/local_process_launcher.py
@@ -1,11 +1,13 @@
 import logging
 import os
+import subprocess
 
 import psutil
 
 from osbenchmark import time, telemetry
 from osbenchmark.builder import java_resolver, cluster
 from osbenchmark.builder.launchers.launcher import Launcher
+from osbenchmark.exceptions import LaunchError
 from osbenchmark.utils import io, opts
 from osbenchmark.utils.periodic_waiter import PeriodicWaiter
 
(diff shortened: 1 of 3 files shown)


osbenchmark/builder/launchers/local_process_launcher.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+class LocalProcessLauncher(Launcher):
+    PROCESS_WAIT_TIMEOUT_SECONDS = 90
+    PROCESS_WAIT_INTERVAL_SECONDS = 0.5
+
+    def __init__(self, pci, shell_executor, metrics_store, clock=time.Clock):
+        super().__init__(shell_executor)
+        self.logger = logging.getLogger(__name__)
+        self.pci = pci
```
IanHoang, line 22: I think `provision_config_instance` is better than using `pci` since it's more descriptive.

engechas: I've gone back and forth on this in my mind a couple times. I settled on `pci` for brevity since many of the accesses to this object will be something like `self.pci.variables["abcd"]["efgh"]`.

But ultimately clarity > brevity so I will refactor. I have used `pci` in a few of the other PRs as well so I will create a followup PR after this one to replace all of the references in 1 go

osbenchmark/builder/launchers/local_process_launcher.py
```diff
@@ -0,0 +1,187 @@
...
+            self._set_env(env, "PATH", os.path.join(java_home, "bin"), separator=os.pathsep, prepend=True)
+            # This property is the higher priority starting in ES 7.12.0, and is the only supported java home in >=8.0
+            env["OPENSEARCH_JAVA_HOME"] = java_home
+            # TODO remove this when ES <8.0 becomes unsupported by Benchmark
+            env["JAVA_HOME"] = java_home
+            self.logger.info("JAVA HOME: %s", env["JAVA_HOME"])
+        if not env.get("OPENSEARCH_JAVA_OPTS"):
+            env["OPENSEARCH_JAVA_OPTS"] = "-XX:+ExitOnOutOfMemoryError"
+
+        # we just blindly trust telemetry here...
```
IanHoang, line 87: Is this a note or will this be a TODO to refactor this portion so that we don't blindly trust telemetry in the future?

osbenchmark/builder/launchers/local_process_launcher.py
```diff
@@ -0,0 +1,187 @@
...
+        try:
+            pid = self._get_pid_from_file(pid_file_name)
+            return pid != 0
+        except (FileNotFoundError, EOFError):
+            self.logger.info("PID file %s is not ready", pid_file_name)
+            return False
+
+    def _get_pid_from_file(self, pid_file_name):
+        with open(pid_file_name, "rb") as f:
+            buf = f.read()
```
IanHoang, line 130: Nit: what's buf represent?

IanHoang (approved):
Added a few more comments to get some clarity but otherwise, LGTM!

osbenchmark/builder/launchers/local_process_launcher.py
engechas: This is an inherited comment. I don't think it's a TODO so much as calling out that the telemetry devices can set environment variables that may break the launching of OpenSearch. Essentially loosely passing the burden of validating the env vars that a telemetry device specifies to the telemetry device itself.

A better solution would probably be an allowlist, but I don't have enough familiarity with the scope of telemetry environment variables to propose one.

osbenchmark/builder/launchers/local_process_launcher.py
engechas: It's the byte buffer read from the file

achitojha (approved):

travisbenedict (approved):

PR #167, opened 2022-03-24
Create DockerInstaller

engechas:
### Description
Migrates the DockerProvisioner to a DockerInstaller that conforms to the Installer interface. As with the other migrations, the key changes are copying and refactoring the unit tests, substituting shell calls with executor.execute, swap out references to the `cfg` object with references to the `provision_config_instance`, and code cleanup/refactoring.

Aside from the above, one key decision this CR makes is to drop support for multiple nodes per host when building a cluster. This decision may be revisited in the future.
 
#134  
 
### Check List

### DIFF
--- a/osbenchmark/builder/installers/docker_installer.py
+++ b/osbenchmark/builder/installers/docker_installer.py
@@ -7,92 +7,113 @@ from jinja2 import select_autoescape
 
 from osbenchmark import paths
 from osbenchmark.builder.installers.installer import Installer
-from osbenchmark.builder.provisioner import NodeConfiguration
-from osbenchmark.utils import convert, io
+from osbenchmark.builder.models.node import Node
+from osbenchmark.utils import io, console
 
 
 class DockerInstaller(Installer):
-    def __init__(self, pci, executor, node_name, ip, http_port, node_root_dir):
+    def __init__(self, provision_config_instance, executor):
         super().__init__(executor)
         self.logger = logging.getLogger(__name__)
-        self.pci = pci
-
-        #What to do about node info?
-        self.node_name = node_name
-        self.node_ip = ip
-        self.http_port = http_port
-        self.node_root_dir = node_root_dir
-        self.node_log_dir = os.path.join(node_root_dir, "logs", "server")
-        self.heap_dump_dir = os.path.join(node_root_dir, "heapdump")
-        self.binary_path = os.path.join(node_root_dir, "install")
-        # use a random subdirectory to isolate multiple runs because an external (non-root) user cannot clean it up.
-        self.data_paths = [os.path.join(node_root_dir, "data", str(uuid.uuid4()))]
-
-        # How to handle this for remote machines??
-        self.node_root_dir = node_root_dir
-        self.node_log_dir = os.path.join(node_root_dir, "logs", "server")
-        self.heap_dump_dir = os.path.join(node_root_dir, "heapdump")
-        self.binary_path = os.path.join(node_root_dir, "install")
-        # use a random subdirectory to isolate multiple runs because an external (non-root) user cannot clean it up.
-        self.data_paths = [os.path.join(node_root_dir, "data", str(uuid.uuid4()))]
+        self.provision_config_instance = provision_config_instance
 
     def install(self, host, binaries):
-        self._prepare(host)
+        return self._prepare(host)
 
     def _prepare(self, host):
-        # we need to allow other users to write to these directories due to Docker.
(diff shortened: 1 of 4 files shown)


osbenchmark/builder/installers/docker_installer.py
```diff
@@ -0,0 +1,169 @@
...
+
+        self.logger.info("Wiping benchmark candidate installation at [%s].", host.node.binary_path)
+
+        for data_path in host.node.data_paths:
+            self._delete_path(host, data_path)
+
+        self._delete_path(host, host.node.binary_path)
+
+    def _delete_path(self, host, path):
+        self.executor.execute(host, "rm -r " + path)
```
engechas, line 169: This seems dangerous to me. Open to other ideas

osbenchmark/builder/installers/docker_installer.py
```diff
@@ -6,11 +6,11 @@
 from jinja2 import select_autoescape
 
 from osbenchmark.builder.installers.installer import Installer
-from osbenchmark.exceptions import InvalidSyntax, SystemSetupError
 
 
 class DockerInstaller(Installer):
-    def __init__(self, pci, node_name, ip, http_port, node_root_dir, distribution_version, benchmark_root):
+    def __init__(self, pci, executor, node_name, ip, http_port, node_root_dir, distribution_version, benchmark_root, ):
```
weifenh: nit: tailing comma seems preferred for multiple line argument lists here ...

tests/builder/installers/docker_installer_test.py
```diff
@@ -0,0 +1,191 @@
...
+          test: curl -f http://localhost:39200 -u admin:admin --insecure
+          interval: 5s
+          timeout: 2s
+          retries: 10
+
+volumes:
+  opensearch-data1:
+networks:
+  opensearch-net:
+""" % (self.node_data_dir, self.node_log_dir, self.node_heap_dump_dir), docker_cfg)
```
weifenh, line 130: Seems this whole multiple-line string is the same as the one below? I'd read it from a file to keep the source of true in a single place. OK to leave it as is.

osbenchmark/builder/installers/docker_installer.py
engechas: Is this from an intermediary commit? The init signature currently looks like this:
```
def __init__(self, provision_config_instance, executor):
```

osbenchmark/builder/installers/docker_installer.py
weifenh: If I use rm -rf, then would do something to check if the path string has some signatures to be OK to be deleted, including
* not empty
* not a very high level folder like "/" or home folder
* not "*"

osbenchmark/builder/installers/docker_installer.py
weifenh: Ah, ok, ignore it.

weifenh (approved):
A rev is welcomed but LGTM already.

tests/builder/installers/docker_installer_test.py
engechas: It's not quite the same, the second one has additional lines:
```
    cpu_count: 2
    mem_limit: 256m
```

Not a big fan of the file as string check in these tests, maybe moving the strings into files within the test package would be cleaner

osbenchmark/builder/installers/docker_installer.py
engechas: Validating the `path` was one thought I had, but it doesn't seem quite secure enough. May be the best option available though.

osbenchmark/builder/installers/docker_installer.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def _prepare_mounts_from_config_path(self, host, node, config_path):
+        config_path_mounts = {}
+        for root, _, files in os.walk(config_path):
+            config_path_mounts.update(self._prepare_mounts_from_config_path_root(host, node, config_path, root, files))
+
+        return config_path_mounts
+
+    def _prepare_mounts_from_config_path_root(self, host, node, config_path, root, files):
+        env = jinja2.Environment(loader=jinja2.FileSystemLoader(root),
```
travisbenedict, line 79: Not required for this PR but I think we might want to move this Jinja environment setup into its own utils function. I think it's used several times across the codebase and ensuring that each caller passes the autoescape characters seems like a pain.

travisbenedict (approved):

PR #168, opened 2022-03-25
Refactor JvmResolver to use executor, remove unused methods, simplify existing methods

engechas:
### Description
This PR migrates the [jvm.py](https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/utils/jvm.py) file to use an executor for it's calls. This will allow the builder system subcomponents to use this class regardless of the cluster provider.

Many of the methods in the previous class were unused. They have been removed. Additionally, much of the logic within the `resolve_path` call has been simplified.

I initially shied away from changing this class as I thought it was used across the code base. After further inspection, it's only used by the builder system.

### Issues Resolved
#134 
 
### Check List

### DIFF
--- a/osbenchmark/builder/utils/jdk_resolver.py
+++ b/osbenchmark/builder/utils/jdk_resolver.py
@@ -5,11 +5,11 @@ from osbenchmark.exceptions import SystemSetupError
 from osbenchmark.utils import io
 
 
-class JvmResolver:
+class JdkResolver:
     def __init__(self, executor):
         self.executor = executor
 
-    def resolve_path(self, host, majors):
+    def resolve_jdk_path(self, host, majors):
         """
         Resolves the path to the JDK with the provided major version(s). It checks the versions in the same order specified in ``majors``
         and will return the first match. To achieve this, it first checks the major version x in the environment variable ``JAVAx_HOME``
@@ -17,16 +17,16 @@ class JvmResolver:
 
         If no appropriate version is found, a ``SystemSetupError`` is raised.
 
-        :param host: The host on which to resolve the JVM path
+        :param host: The host on which to resolve the JDK path
         :param majors: Either a list of major versions to check or a single version as an ``int``.
         :return: A tuple of (major version, path to Java home directory).
         """
         if isinstance(majors, int):
-            return majors, self._resolve_path(host, [majors])
+            return majors, self._resolve_jdk_path(host, [majors])
         else:
-            return majors, self._resolve_path(host, majors)
+            return majors, self._resolve_jdk_path(host, majors)
 
-    def _resolve_path(self, host, majors):
+    def _resolve_jdk_path(self, host, majors):
         """
         Resolves the path to a JDK with one of the provided major versions.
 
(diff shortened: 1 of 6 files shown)


osbenchmark/builder/utils/jvm_resolver.py
```diff
@@ -0,0 +1,103 @@
...
+
+from osbenchmark.exceptions import SystemSetupError
+from osbenchmark.utils import io
+
+
+class JvmResolver:
+    def __init__(self, executor):
+        self.executor = executor
+
+    def resolve_path(self, host, majors):
```
weifenh, line 12: I'd rename the func name as ``resolve_jdk_path``(same below) in order to be more specific, because there could be other kinds of paths related to jvm... e.g. log path... or can rename the class as JdkResolver? Otherwise LGTM.

engechas: That's a good call, jvm doesn't make a ton of sense here. Will change references of `jvm` to `jdk`

travisbenedict (approved):

travisbenedict (commented):
Is `jvm.py` going to be removed in another PR?

engechas:
> Is `jvm.py` going to be removed in another PR?

Yes. All of the classes being migrated will be removed in one fell swoop once all of the migrations are finished and rigorous E2E testing is performed

osbenchmark/builder/utils/jdk_resolver.py
```diff
@@ -0,0 +1,103 @@
...
+        # are we under the "old" (pre Java 9) or the new (Java 9+) version scheme?
+        if version.startswith("1."):
+            return int(version[2])
+        else:
+            return int(version)
+
+    def _system_property(self, host, java_home, system_property_name):
+        lines = self.executor.execute(host, "{} -XshowSettings:properties -version".format(self._java(java_home)), output=True)
+        # matches e.g. "    java.runtime.version = 1.8.0_121-b13" and captures "1.8.0_121-b13"
+        sys_prop_pattern = re.compile(r".*%s.*=\s?(.*)" % system_property_name)
```
IanHoang, line 83: Nit: Might be worth making this into a constant

osbenchmark/builder/utils/jdk_resolver.py
```diff
@@ -0,0 +1,103 @@
...
+        and will return the first match. To achieve this, it first checks the major version x in the environment variable ``JAVAx_HOME``
+        and falls back to ``JAVA_HOME``. It also ensures that the environment variable points to the right JDK version.
+
+        If no appropriate version is found, a ``SystemSetupError`` is raised.
+
+        :param host: The host on which to resolve the JDK path
+        :param majors: Either a list of major versions to check or a single version as an ``int``.
+        :return: A tuple of (major version, path to Java home directory).
+        """
+        if isinstance(majors, int):
```
IanHoang, line 24: Nit: Are we following pep8? If so, this method should be updated to `is_instance`

IanHoang (approved):
LGTM!! 👍🏻

osbenchmark/builder/utils/jdk_resolver.py
engechas: Not sure tbh, I've been relying on `make precommit` for code styling. Should we be following pep8?

IanHoang: Update: spoke offline, I misunderstood that isinstance() is a built-in function and not in reference to our cluster and AOS terminology.

travisbenedict: This is a python built in function. We can't do anything about the formatting :)

https://docs.python.org/3/library/functions.html#isinstance

PR #171, opened 2022-03-29
OpenSearchInstaller Migration

engechas:
### Description
Migrates the OpenSearchInstaller. This installer is responsible for preparing the binaries and config files of a source or distribution installation of OpenSearch.

Additionally, a lot of shared logic that was previously in the DockerInstaller has been moved to the Installer base class. This simplifies much of the logic in the Docker and OpenSearch installers
 
### Issues Resolved
#134 
 
### Check List
- [ ] Commits are signed per the DCO using --signoff 


osbenchmark/builder/installers/installer.py
```diff
@@ -1,7 +1,10 @@
...
+
 import jinja2
+from jinja2 import select_autoescape
 
 from osbenchmark.exceptions import InvalidSyntax, SystemSetupError
-from osbenchmark.utils import io
+from osbenchmark.utils import io, console
 
 
 class Installer:
```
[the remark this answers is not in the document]
engechas: This question came up in a previous, non-OSB project. The consensus there was not to use ABC but I don't know if I still agree with that. 

I think there is more benefit to using ABC than not, primarily to enforce method implementations. I will convert the builder subcomponent base classes to use ABC unless anyone's opposed.

engechas:
Apologies that the latest commit made the PR quite large. There is no functional change introduced in the latest commit, instead the following refactoring was done:

* Refactor interfaces to use ABC
* After speaking offline with Achit, the shared logic was refactored to use helper classes. This aligns with preferring composition over inheritance
* Added unit tests for the new helper classes

travisbenedict (approved):

osbenchmark/builder/installers/opensearch_installer.py
```diff
@@ -0,0 +1,113 @@
...
+        config_path = os.path.join(node.binary_path, "config")
+        self.logger.info("Deleting pre-bundled OpenSearch configuration at [%s]", config_path)
+        self.path_manager.delete_path(host, config_path)
+
+    def _prepare_config_files(self, host, node, all_node_ips):
+        config_vars = self.get_config_vars(host, node, all_node_ips)
+        self.config_applier.apply_configs(host, node, self.provision_config_instance.config_paths, config_vars)
+
+    def get_config_vars(self, host, node, all_node_ips):
+        provisioner_defaults = {
```
IanHoang, line 87: Nit: I doubt there's going to be another iteration of an installer but how do you feel about adding these keys of provisioner_defaults to an enum? 

Not sure where else provisioner_defaults might be used in the code. It might make this dictionary a more muddled; However, I'm curious if we should start implementing enums throughout OSB to enable reusability of terms and if we were ever to update the naming conventions of variables such as these, it'd be easier. Curious to what others think too.

osbenchmark/builder/installers/opensearch_installer.py
```diff
@@ -0,0 +1,113 @@
...
+            "node_ip": host.address,
+            # this is the IP address that the node will be bound to. Benchmark will bind to the node's IP address (but not to 0.0.0.0). The
+            "network_host": host.address,
+            "http_port": str(node.port),
+            "transport_port": str(node.port + 100),
+            "all_node_ips": "[\"%s\"]" % "\",\"".join(all_node_ips),
+            # at the moment we are strict and enforce that all nodes are master eligible nodes
+            "minimum_master_nodes": len(all_node_ips),
+            "install_root_path": node.binary_path
+        }
```
IanHoang, line 103: Nit: I doubt there's going to be another iteration of an installer but how do you feel about adding these keys of provisioner_defaults to an enum? 

Not sure where else provisioner_defaults might be used in the code. It might make this dictionary a more muddled; However, I'm curious if we should start implementing enums throughout OSB to enable reusability of terms and if we were ever to update the naming conventions of variables such as these, it'd be easier. Curious to what others think too.

IanHoang (approved):
LGTM

achitojha (approved):

osbenchmark/builder/installers/opensearch_installer.py
engechas: Different installers have different keys for their default variables. There's some overlap, but I don't think there's a good solution for a rigid model representing the same.

PR #173, opened 2022-03-31
Plugin installer migration

engechas:
### Description
Migrates the PluginInstaller. Also slight refactors to OpenSearchInstaller to rename `provisioner_defaults` -> `installer_defaults` and make config_vars an optional param for the `install` method. The CompositeInstaller coming in the next PR will use this new parameter to pass in `config_vars` overrides.
 
### Issues Resolved
#134 
 
### Check List


osbenchmark/builder/installers/opensearch_installer.py
```diff
@@ -26,9 +26,10 @@ def __init__(self, provision_config_instance, executor, hook_handler_class=Boots
         self.config_applier = ConfigApplier(executor, self.template_renderer, self.path_manager)
         self.host_cleaner = HostCleaner(self.path_manager)
 
-    def install(self, host, binaries, all_node_ips):
+    # pylint: disable=arguments-differ
+    def install(self, host, binaries, all_node_ips, config_vars=None):
```
travisbenedict, line 30: Should `config_vars` just be added to the `Installer.install` interface? 

Is there a reason that disabling the linter is preferred?

engechas: `config_vars` shouldn't be part of this method's signature at all IMO. An installer should be responsible for owning and defining its own config_vars rather than having some passed into it.

The trouble comes when both OS and plugin(s) are set to be installed. In this case, the config_vars [from all sources must be considered](https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/builder/provisioner.py#L215-L240). Adding `config_vars` to the signature allows the CompositeInstaller to pass in the finalized config_vars from all sources without changing the general flow of the sub-installers. OpenSearchInstaller can be consumed as a standalone Installer implementation or utilized by the CompositeInstaller via the same install method.

I feel that I am missing a design pattern that would eliminate the need to add `config_vars` to this method. I will think on it some more and try to come up with a better solution.

achitojha: I agree with your assessment that `config_vars` shouldn't be a part of this method. Its essentially an implementation detail that the user of this method shouldn't care about. Also, that's why the `Installer` interface shouldn't & doesn't have it.

Please correct me if I'm wrong, but it seems like each installer already just makes a single install call in its lifetime. Constructor is instance specific and can be different for classes implementing the same interface, have you considered having the `config_vars` as an optional constructor param for `OpenSearchInstaller` and making it an class level variable? 

I believe this way the CompositeInstaller can have multiple sub-installers, and not break their general flow. However, if you mean the CompositeInstaller needs results from sub-installers to come up with the final `config_vars`, I'd still recommend the same approach, but using a factory class the generates the OS installer instance after the `config_vars` are ready.

osbenchmark/builder/installers/plugin_installer.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class PluginInstaller(Installer):
+    def __init__(self, plugin, executor, hook_handler_class=BootstrapHookHandler):
+        super().__init__(executor)
+        self.logger = logging.getLogger(__name__)
+        self.plugin = plugin
+        self.hook_handler = hook_handler_class(self.plugin)
+        if self.hook_handler.can_load():
+            self.hook_handler.load()
```
achitojha, line 18: Does this need to happen during the object construction itself? Would it make sense to move this to the install logic?

I'm not fundamentally against it but prefer lightweight constructors. Think of scenarios where these fail and what stack trace they leave - I feel it makes more sense if we say the plugin failed to install /load. Instead if the constructor throws an exception, I suspect there's things like class not found exception or something which isn't a good description of what happened.

osbenchmark/builder/installers/opensearch_installer.py
engechas: The CompositeInstaller does need results from the sub-installers to come up with the final `config_vars`. Specifically, it needs the Node object generated by the OpenSearchInstaller's `install` method.

With the current implementation there's no way to generate the `config_vars` without first instantiating an instance of OpenSearchInstaller and invoking the `install` method to get a Node. I think a further layer of abstraction to separate preparing the installation and applying the config data is the solution here: create new OpenSearchPreparer and PluginPreparer classes with `prepare` and `get_config_vars` methods. 

Then the OpenSearch and PluginInstallers would take in a respective instance of these Preparer classes and perform the whole installation flow (including applying the config) within the install method. The CompositeInstaller would have instances of both OpenSearchPreparer and PluginPreparer, with the flow being: call prepare methods -> call `get_config_vars` methods -> generate final `config_vars` -> apply configs

I plan to make heavy use of the factory pattern when selecting the correct builder components for the `ClusterBuilder` :)

osbenchmark/builder/installers/plugin_installer.py
engechas: It could go in the `install` logic. The rationale for having it here is to fail fast if an exception is thrown when loading the `hook_handler`.

If this logic is moved to the `install` method, then the exception it creates on failure would only be thrown after the infrastructure has been provisioned and the necessary data downloaded and installed on the host. By failing in the constructor, no resources are created or work done for a scenario that couldn't possibly succeed.

Perhaps wrapping the exception with some documentation on the failure would be a solution to the vague description. Given that this `hook_handler` logic is used in multiple Installers, adding that handling into the factory generating Installer instances would be a good approach.

engechas:
After thinking more on it, there's no need to distinguish between a SingularBareInstaller and a CompositeBareInstaller as the flow can be the exact same via looping. Instead, I've opted to create a BareInstaller that handles both scenarios

weifenh (approved):

osbenchmark/builder/installers/bare_installer.py
```diff
@@ -0,0 +1,76 @@
+from osbenchmark.builder.installers.installer import Installer
+from osbenchmark.builder.installers.preparers.plugin_preparer import PluginPreparer
+from osbenchmark.builder.provision_config import BootstrapPhase
+from osbenchmark.builder.utils.config_applier import ConfigApplier
+from osbenchmark.builder.utils.java_home_resolver import JavaHomeResolver
+from osbenchmark.builder.utils.path_manager import PathManager
+from osbenchmark.builder.utils.template_renderer import TemplateRenderer
+
+
+class BareInstaller(Installer):
```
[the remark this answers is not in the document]
engechas: Bare essentially means "not in a container". The original implementation used this name and I carried forward since I can't think of a better one. Open to suggestions if anyone has ideas

osbenchmark/builder/installers/bare_installer.py
```diff
@@ -0,0 +1,76 @@
...
+            env["JAVA_HOME"] = java_home
+
+        config_vars_copy = config_vars.copy()
+        for preparer in self.preparers:
+            preparer.invoke_install_hook(host, BootstrapPhase.post_install, config_vars_copy, env)
+
+    def _get_node(self, preparer_to_node):
+        nodes_list = list(filter(lambda node: node is not None, preparer_to_node.values()))
+
+        assert len(nodes_list) == 1, "Exactly one node must be provisioned per host, but found nodes: {}".format(nodes_list)
```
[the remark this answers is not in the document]
engechas: `preparers_to_nodes` is a map of a `Preparer` to the return value of its `prepare` method, which can be either a `Node` object or `None`. Only 1 `Preparer` should be generating a `Node`, the others are used to install additional components (i.e. plugins) to that same `Node`

Allowing more than one node per host could be implemented in the future, but there doesn't seem to be much justification for supporting it at the moment.

travisbenedict (approved):

PR #176, opened 2022-04-05
Migrate the OpenSearch and PluginDistributionDownloaders

engechas:
### Description
Migrates the Distribution based downloaders to conform to the new Downloader interface.
 
### Issues Resolved
#134 
 
### Check List

### DIFF
--- a/osbenchmark/builder/downloaders/opensearch_distribution_downloader.py
+++ b/osbenchmark/builder/downloaders/opensearch_distribution_downloader.py
@@ -1,0 +1,66 @@
+import logging
+import os.path
+
+from osbenchmark.builder.downloaders.downloader import Downloader
+from osbenchmark.builder.downloaders.repositories.opensearch_distribution_repository_provider import \
+    OpenSearchDistributionRepositoryProvider
+from osbenchmark.builder.utils.path_manager import PathManager
+from osbenchmark.exceptions import ExecutorError
+
+
+class OpenSearchDistributionDownloader(Downloader):
+    BINARY_KEY = "opensearch"
+
+    def __init__(self, provision_config_instance, executor):
+        super().__init__(executor)
+        self.logger = logging.getLogger(__name__)
+        self.provision_config_instance = provision_config_instance
+        self.path_manager = PathManager(executor)
+        self.distribution_repository_provider = OpenSearchDistributionRepositoryProvider(provision_config_instance, executor)
+
+    def download(self, host):
+        binary_path = self._fetch_binary(host)
+        return {OpenSearchDistributionDownloader.BINARY_KEY: binary_path}
+
+    def _fetch_binary(self, host):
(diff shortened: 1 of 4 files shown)


osbenchmark/builder/downloaders/opensearch_distribution_downloader.py
```diff
@@ -0,0 +1,66 @@
...
+    def __init__(self, provision_config_instance, executor):
+        super().__init__(executor)
+        self.logger = logging.getLogger(__name__)
+        self.provision_config_instance = provision_config_instance
+        self.path_manager = PathManager(executor)
+        self.distribution_repository_provider = OpenSearchDistributionRepositoryProvider(provision_config_instance, executor)
+
+    def download(self, host):
+        binary_path = self._fetch_binary(host)
+        return {OpenSearchDistributionDownloader.BINARY_KEY: binary_path}
```
achitojha, line 23: Looked at the interface and understand that it should be a map of various items being downloaded. Do you mind sharing examples of what are the different items that may get downloaded(Understand it'll be OS and associated plugins). Specifically, I'd like to understand how this map would be used by other classes. Do we have scenarios where this could have multiple items?

osbenchmark/builder/downloaders/opensearch_distribution_downloader.py
```diff
@@ -0,0 +1,66 @@
...
+import os.path
+
+from osbenchmark.builder.downloaders.downloader import Downloader
+from osbenchmark.builder.downloaders.repositories.opensearch_distribution_repository_provider import \
+    OpenSearchDistributionRepositoryProvider
+from osbenchmark.builder.utils.path_manager import PathManager
+from osbenchmark.exceptions import ExecutorError
+
+
+class OpenSearchDistributionDownloader(Downloader):
```
achitojha, line 11: Do we only support OpenSearch?

engechas: For now, yes. Supporting ES is a non-MVP "nice to have"

osbenchmark/builder/downloaders/opensearch_distribution_downloader.py
engechas: >Specifically, I'd like to understand how this map would be used by other classes.

This map is used by the installers to prepare the files that were downloaded: https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/builder/installers/preparers/opensearch_preparer.py#L48

>Do we have scenarios where this could have multiple items?

Similar to the BareInstaller, there will be a `CompositeDownloader` fetches the binaries for both OpenSearch and plugins. It will combine the maps returned from the individual Downloaders into a single binaries map for use by the Installer

travisbenedict (approved):

achitojha (approved):

PR #179, opened 2022-04-07
CorePluginSourceDownloader Migration

engechas:
### Description
Migrates the CorePluginSourceDownloader. Also implements a help class `SourceBuilder`
 
### Issues Resolved
#134 
 
### Check List


osbenchmark/builder/downloaders/builders/source_builder.py
```diff
@@ -0,0 +1,40 @@
...
+        self.executor = executor
+        self.path_manager = PathManager(executor)
+        self.jdk_resolver = JdkResolver(executor)
+
+        self.source_directory = source_directory
+        self.build_jdk = build_jdk
+        self.log_directory = log_directory
+
+    def build(self, host, build_commands, override_source_directory=None):
+        if isinstance(build_commands, str):
```
travisbenedict, line 21: Is there only going to be this Builder? I'm wondering if the type check makes sense to include if this is using an interface that multiple builders would implement.

Minor: But if this is the only Builder then line 24 in `CorePluginSourceDownloader` could just pass the string instead of a list. I would also consider changing the name of the field in that class to be `source_builder instead` of just `builder`

osbenchmark/builder/downloaders/builders/source_builder.py
```diff
@@ -0,0 +1,40 @@
...
+from osbenchmark.builder.utils.path_manager import PathManager
+from osbenchmark.exceptions import ExecutorError, BuildError
+
+
+class SourceBuilder:
+    def __init__(self, executor, source_directory, build_jdk, log_directory):
+        self.logger = logging.getLogger(__name__)
+        self.executor = executor
+        self.path_manager = PathManager(executor)
+        self.jdk_resolver = JdkResolver(executor)
```
achitojha, line 14: General comment : How about we use factory pattern here to avoid hard bounding concrete dependencies?

engechas: That's a good idea, will refactor

osbenchmark/builder/downloaders/builders/source_builder.py
engechas: This is the only one as of now. There could very well be a need for other ones in the future so I'll refactor it into an interface. I'll get rid of the typecheck then too, was added as I made the mistake of passing a string in the unit tests and figured that could be an easy mistake for others to make as well :)

osbenchmark/builder/downloaders/builders/source_binary_builder.py
```diff
@@ -0,0 +1,36 @@
+import logging
+import os
+
+from osbenchmark.builder.downloaders.builders.binary_builder import BinaryBuilder
+from osbenchmark.exceptions import ExecutorError, BuildError
+
+
+class SourceBinaryBuilder(BinaryBuilder):
+    def __init__(self, executor, path_manager, jdk_resolver, source_directory, build_jdk, log_directory):
```
weifenh, line 9: I'd rename it as ``build_jdk_version``, but up to you. The naming ``build_jdk`` mostly makes sense, it's about the jdk version, but there's a slight chance that people may think it's a boolean first then figure out "build a JDK" wound not make sense here.

weifenh (approved):
Generally LGTM.

osbenchmark/builder/downloaders/builders/source_binary_builder.py
engechas: That's a good callout, will rename

osbenchmark/builder/downloaders/builders/binary_builder.py
```diff
@@ -0,0 +1,19 @@
...
+
+    @abstractmethod
+    def build(self, host, build_commands, override_source_directory):
+        """
+        Runs the provided commands on the given host to build binaries
+
+        ;param host: A host object representing the machine on which to run the commands
+        ;param build_commands: A list of strings representing sequential bash commands used to build the binaries
+        ;param override_source_directory: A string representing the source directory where the pre-binary code is located
+        ;return None
```
travisbenedict, line 17: Nit: Should these semicolons be colons?

engechas: Yes. Not sure where I picked up the semi-colon thing but I've made that mistake elsewhere too. Will fix this instance then blanket refactor the rest in a followup PR

travisbenedict (approved):

Issue #150, opened 2022-03-02
Remove mapping types


dreamer-89:
As part of v2.0 release, `mapping types` are getting removed from OpenSearch engine. Below are the changes in the opensearch-engine.
- Removal of type from rest end-points
- Removal of include_type_name parameter from API requests.

As part of this issue, please verify if type removal change on OpenSearch engine impacts this repository. If yes, then please remove the type references/usage from this repository

Related: [OpenSearch-engine meta issue](https://github.com/opensearch-project/opensearch/issues/1940)

engechas:
This change will affect OSB, we will need to prioritize a fix.

### Areas of Change
https://github.com/opensearch-project/opensearch-benchmark/blob/bd88b91c5a38aa3fd328c087189ca944ad0d7714/osbenchmark/metrics.py#L80
https://github.com/opensearch-project/opensearch-benchmark/blob/f640f98cca7df30c1399262d8f26dcdd28ac6547/osbenchmark/resources/test-executions-template.json#L12
https://github.com/opensearch-project/opensearch-benchmark/blob/f640f98cca7df30c1399262d8f26dcdd28ac6547/osbenchmark/resources/metrics-template.json#L12
https://github.com/opensearch-project/opensearch-benchmark/blob/f640f98cca7df30c1399262d8f26dcdd28ac6547/osbenchmark/resources/results-template.json#L12

### Potential other areas that will need to change (needs further investigation)
https://github.com/opensearch-project/opensearch-benchmark/blob/bd88b91c5a38aa3fd328c087189ca944ad0d7714/tests/metrics_test.py#L376
https://github.com/opensearch-project/opensearch-benchmark/blob/bd88b91c5a38aa3fd328c087189ca944ad0d7714/tests/worker_coordinator/runner_test.py#L429
https://github.com/opensearch-project/opensearch-benchmark/blob/bd88b91c5a38aa3fd328c087189ca944ad0d7714/tests/worker_coordinator/runner_test.py#L447
https://github.com/opensearch-project/opensearch-benchmark/blob/bd88b91c5a38aa3fd328c087189ca944ad0d7714/osbenchmark/metrics.py#L777
https://github.com/opensearch-project/opensearch-benchmark/blob/bd88b91c5a38aa3fd328c087189ca944ad0d7714/osbenchmark/metrics.py#L1464
[...]

PR #186, opened 2022-04-18
Change bulk operation for metrics and index templates to support Open…

travisbenedict:
…Search 2.0 as a data store

Signed-off-by: travisbenedict <travisrb10@gmail.com>

### Description
Add support for OpenSearch 2.0 as a metrics data store by updating the MDS index templates, removing the `doc_type` argument from bulk requests and removing `include_type_name` from put template requests. 

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-benchmark/issues/150
 
### Check List
- [X] Commits are signed per the DCO using --signoff 

### DIFF
--- a/osbenchmark/metrics.py
+++ b/osbenchmark/metrics.py
@@ -27,7 +27,6 @@ import glob
 import json
 import logging
 import math
-import opensearchpy.helpers
 import os
 import pickle
 import random
@@ -37,7 +36,7 @@ import time
 import zlib
 from enum import Enum, IntEnum
 from http.client import responses
-
+import opensearchpy.helpers
 import tabulate
 
 from osbenchmark import client, time, exceptions, config, version, paths
--- a/osbenchmark/metrics.py
+++ b/osbenchmark/metrics.py
@@ -27,6 +27,7 @@ import glob
 import json
 import logging
 import math
+import opensearchpy.helpers
 import os
 import pickle
 import random
@@ -54,7 +55,6 @@ class OsClient:
         self._cluster_version = cluster_version
         self._cluster_distribution = None
 
-    # TODO #653: Remove version-specific support for metrics stores before 7.0.0.
     def probe_version(self):
         info = self.guarded(self._client.info)
         try:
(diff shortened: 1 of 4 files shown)


engechas:
This change was discussed a bit in #108. By merging this PR we are dropping support for metrics stores <ES7. This seems fine to me but there are sure to be other cases of ES/OS conflict in the future. It would be good to align on a decision about how to handle any and all of these conflicts going forwards.

travisbenedict:
> This change was discussed a bit in #108. By merging this PR we are dropping support for metrics stores <ES7. This seems fine to me but there are sure to be other cases of ES/OS conflict in the future. It would be good to align on a decision about how to handle any and all of these conflicts going forwards.

I completely agree with you. A version support policy is something I'm planning to publish soon.

In this particular case I would argue that we implicitly dropped support for Elasticsearch as a data store back in https://github.com/opensearch-project/opensearch-benchmark/pull/74 prior to the alpha release

osbenchmark/metrics.py
```diff
@@ -102,13 +93,7 @@ def refresh(self, index):
         return self.guarded(self._client.indices.refresh, index=index)
 
     def bulk_index(self, index, doc_type, items):
-        # TODO #653: Remove version-specific support for metrics stores before 7.0.0.
-        # pylint: disable=import-outside-toplevel
-        import opensearchpy.helpers
-        if self._cluster_version[0] > 6:
-            self.guarded(opensearchpy.helpers.bulk, self._client, items, index=index, chunk_size=5000)
-        else:
-            self.guarded(opensearchpy.helpers.bulk, self._client, items, index=index, doc_type=doc_type, chunk_size=5000)
```
achitojha, line 111: Can you add some context on why removing metric store support for ES < 6  is needed here?

travisbenedict: Removing support for ES 6 and below is not explicitly needed but would require an additional maintenance burden for us. ES 7 deprecated certain parameters related to put template operations (more info [here](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/removal-of-types.html)). If we want to continue supporting ES 6 as a data store we would need to create duplicates of the index templates and add some additional branching logic.

engechas (approved):

IanHoang (approved):

PR #248, opened 2023-03-30
Add condition to check for distribution version when publishing warn statement

### LINKED ISSUE #243: [BUG] OSB does not recognize OpenSearch versions accurately
**Describe the bug**
When OSB is run with `--telemetry=node-stats`, OSB outputs a warning like the following despite having an OS version that is greater than 1.1. 
```
[WARNING] You have enabled the node-stats telemetry device with OpenSearch < 1.1.0. Requests to the
          _nodes/stats OpenSearch endpoint trigger additional refreshes and WILL SKEW results.
```
For example, a user who had an OS 2.3 ran `--telemetry=node-stats` and still received the warning. This is the output provided by running `curl localhost:9200` from within the cluster. 
```
  "version" : {
    "distribution" : "opensearch",
    "number" : "2.3.0",
```

### Two questions that need to be addressed:
- Is OSB incorrectly fetching the version number and if not, why is warning still being presented for OpenSearch versions greater than 1.1?
- Why does running OSB with node-stats affect OpenSearch versions less than 1.1? After viewing the git blame of the `telemetry.py` file, I have suspicions that the following line was arbitrarily altered for the sole purpose of the fork.  

https://github.com/opensearch-project/opensearch-benchmark/blob/a1f45502b5b9e69bad1c8d10e7a6c30bd0ed8469/osbenchmark/telemetry.py#L614

**To Reproduce**
Run OSB with `--telemetry=node-stats` on a version greater than OS 1.1. 

**Expected behavior**
The warning should not appear for versions less than OS 1.1. 

**Logs**
If applicable, add logs to help explain your problem.

**More Context (please complete the following information):**
- Seen with NYC_Taxis but could be all other workloads too
- OS Versions > 1.1
- OSB 0.2.0

**Additional context**

IanHoang:
### Description
Users who are using OpenSearch versions greater than or less than 1.1.0 are still seeing the warn statement, which is misleading. This statement was originally intended for users running telemetry on ES versions 7.2.0 or less. This quick PR adds a condition to check if the distribution name is `opensearch` or `elasticsearch` in case OSB still wants to test ES versions <=7.10 (or prior to OpenSearch 1.0). 

Also, revised a logging statement from `ES` to `OpenSearch` when referring to the client. 

### Issues Resolved
#243 

### Testing

Tested OSB with and without changes in a fresh Ubuntu environment. Confirmed that the warn statement appears on the console and logs only when distribution name is `elasticsearch` and version is 7.2.0 or less.

---

### DIFF
--- a/osbenchmark/telemetry.py
+++ b/osbenchmark/telemetry.py
@@ -614,6 +614,7 @@ class NodeStats(TelemetryDevice):
     warning = """You have enabled the node-stats telemetry device with OpenSearch < 1.1.0. Requests to the
           _nodes/stats OpenSearch endpoint trigger additional refreshes and WILL SKEW results.
     """
+    opensearch_distribution = "opensearch"
 
     def __init__(self, telemetry_params, clients, metrics_store):
         super().__init__()
@@ -625,10 +626,11 @@ class NodeStats(TelemetryDevice):
 
     def on_benchmark_start(self):
         default_client = self.clients["default"]
+        distribution_name = default_client.info()["version"]["distribution"]
         distribution_version = default_client.info()["version"]["number"]
         major, minor = components(distribution_version)[:2]
 
-        if major < 7 or (major == 7 and minor < 2):
+        if distribution_name != NodeStats.opensearch_distribution and (major < 7 or (major == 7 and minor < 2)):
             console.warn(NodeStats.warning, logger=self.logger)
 
         for cluster_name in self.specified_cluster_names:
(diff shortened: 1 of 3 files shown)


osbenchmark/telemetry.py
```diff
@@ -614,6 +614,7 @@ class NodeStats(TelemetryDevice):
     warning = """You have enabled the node-stats telemetry device with OpenSearch < 1.1.0. Requests to the
           _nodes/stats OpenSearch endpoint trigger additional refreshes and WILL SKEW results.
     """
+    opensearch_distribution = "opensearch"
```
gkamat, line 617: Might rename this as `opensearch_distribution_name` to be consistent with the variable names below.

gkamat (approved):

PR #291, opened 2023-05-01
Allow users to specify number of docs per index when creating workloads

### LINKED ISSUE #289: [FEATURE] Add support for number of documents when creating workloads
`create-workload` subcommand currently generates a workload using all the documents from appointed indices. It would be nice if users could specify a number of documents from the indices they've appointed instead of using all the documents.

Currently, this is what `create-workload` supports:
```
usage: opensearch-benchmark create-workload [-h] --workload WORKLOAD --indices INDICES --target-hosts TARGET_HOSTS [--client-options CLIENT_OPTIONS]
                                            [--output-path OUTPUT_PATH] [--quiet] [--offline]

options:
  -h, --help            show this help message and exit
  --workload WORKLOAD   Name of the generated workload
  --indices INDICES     Comma-separated list of indices to include in the workload
  --target-hosts TARGET_HOSTS
                        Comma-separated list of host:port pairs which should be targeted
  --client-options CLIENT_OPTIONS
                        Comma-separated list of client options to use. (default: timeout:60)
  --output-path OUTPUT_PATH
                        Workload output directory (default: workloads/)
  --quiet               Suppress as much as output as possible (default: false).
  --offline             Assume that Benchmark has no connection to the Internet (default: false).
```

There should be an additional parameter that takes in the number of documents of appointed indices to generate workload from.

**Describe alternatives you've considered**

**Additional context**

IanHoang:
### Description
`create-workload` currently fetches all documents from specified indices in `--indices`. Users should have the option to specify a subset of documents if they do not want to use all of the documents. In this PR, OSB now supports running `--total-docs` in conjunction with `--indices`. It takes in a comma-separated list of document counts that correspond to the respective index in the list of indices in `--indices`.
```
# Example: if movies and actors has a total document count of 3000 and 2000 respectively, users can specify that they want only 1500 documents from movies index and 1200 documents from actors index with these parameters.
--indices=movies,actors --number-of-docs movies:1500 actors:1200
```

See #289 for more details.

### Issues Resolved
#289 

### Testing

Tested it with a few indices in private cluster. Tested the following cases:
- with single index and single doc count 
- with multiple indices and multiple doc counts
- with multiple indices and single doc count (should result in error as both doc counts need to be provided)
- with single index (should get all documents)
- with multiple indices (should get all documents)

See #289 for outputs and more details

---

### DIFF
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -182,8 +182,10 @@ def create_arg_parser():
         help="Input JSON file to use containing custom workload queries that override the default match_all query")
     create_workload_parser.add_argument(
         "--number-of-docs",
-        type=opts.csv_to_list,
-        help="Comma-separated list of doc counts for each index specified in indices parameter. Ensure that order is preserved.")
+        action=opts.StoreKeyPairAsDict,
+        nargs='+',
+        metavar="KEY:VAL",
+        help="Map of index name and doc count to extract. Ensure that index name in key field exists in --indices parameter.")
 
     generate_parser = subparsers.add_parser("generate", help="Generate artifacts")
     generate_parser.add_argument(
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -55,12 +55,6 @@ def create_arg_parser():
             raise argparse.ArgumentError(argument=None, message="At least one argument required!")
         return lst
 
-    def list_or_none(arg):
-        lst = opts.csv_to_list(arg)
-        if len(lst) < 1:
-            return None
-        return lst
-
     def runtime_jdk(v):
         if v == "bundled":
             return v
@@ -185,11 +179,10 @@ def create_arg_parser():
     create_workload_parser.add_argument(
         "--custom-queries",
         type=argparse.FileType('r'),
-        help="Input JSON file to use containing custom workload queries that override the default match_all query"
-    )
+        help="Input JSON file to use containing custom workload queries that override the default match_all query")
     create_workload_parser.add_argument(
-        "--total-docs",
-        type=list_or_none,
+        "--number-of-docs",
+        type=opts.csv_to_list,
         help="Comma-separated list of doc counts for each index specified in indices parameter. Ensure that order is preserved.")
 
     generate_parser = subparsers.add_parser("generate", help="Generate artifacts")
@@ -917,7 +910,7 @@ def dispatch_sub_command(arg_parser, args, cfg):
             generate(cfg)
         elif sub_command == "create-workload":
             cfg.add(config.Scope.applicationOverride, "generator", "indices", args.indices)
-            cfg.add(config.Scope.applicationOverride, "generator", "total_docs", args.total_docs)
+            cfg.add(config.Scope.applicationOverride, "generator", "number_of_docs", args.number_of_docs)
             cfg.add(config.Scope.applicationOverride, "generator", "output.path", args.output_path)
             cfg.add(config.Scope.applicationOverride, "workload", "workload.name", args.workload)
             cfg.add(config.Scope.applicationOverride, "workload", "custom_queries", args.custom_queries)
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -55,6 +55,12 @@ def create_arg_parser():
             raise argparse.ArgumentError(argument=None, message="At least one argument required!")
         return lst
 
+    def list_or_none(arg):
+        lst = opts.csv_to_list(arg)
+        if len(lst) < 1:
+            return None
+        return lst
+
     def runtime_jdk(v):
         if v == "bundled":
             return v
@@ -181,6 +187,10 @@ def create_arg_parser():
         type=argparse.FileType('r'),
         help="Input JSON file to use containing custom workload queries that override the default match_all query"
     )
+    create_workload_parser.add_argument(
+        "--total-docs",
+        type=list_or_none,
+        help="Comma-separated list of doc counts for each index specified in indices parameter. Ensure that order is preserved.")
 
     generate_parser = subparsers.add_parser("generate", help="Generate artifacts")
     generate_parser.add_argument(
@@ -907,6 +917,7 @@ def dispatch_sub_command(arg_parser, args, cfg):
             generate(cfg)
         elif sub_command == "create-workload":
             cfg.add(config.Scope.applicationOverride, "generator", "indices", args.indices)
+            cfg.add(config.Scope.applicationOverride, "generator", "total_docs", args.total_docs)
             cfg.add(config.Scope.applicationOverride, "generator", "output.path", args.output_path)
             cfg.add(config.Scope.applicationOverride, "workload", "workload.name", args.workload)
             cfg.add(config.Scope.applicationOverride, "workload", "custom_queries", args.custom_queries)
--- a/osbenchmark/workload_generator/corpus.py
+++ b/osbenchmark/workload_generator/corpus.py
@@ -50,7 +50,7 @@ def get_doc_outpath(outdir, name, suffix=""):
     return os.path.join(outdir, f"{name}-documents{suffix}.json")
 
 
-def extract(client, output_path, index):
+def extract(client, output_path, index, total_docs_requested=None):
     """
     Scroll an index with a match-all query, dumping document source to ``outdir/documents.json``.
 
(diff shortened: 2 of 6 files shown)


osbenchmark/benchmark.py
```diff
@@ -176,6 +182,10 @@ def add_workload_source(subparser):
         "--output-path",
         default=os.path.join(os.getcwd(), "workloads"),
         help="Workload output directory (default: workloads/)")
+    create_workload_parser.add_argument(
+        "--total-docs",
+        type=list_or_none,
```
gkamat, line 187: May be better to just specify `opts.csv_to_list` here, rather than creating another function.

osbenchmark/benchmark.py
```diff
@@ -176,6 +182,10 @@ def add_workload_source(subparser):
         "--output-path",
         default=os.path.join(os.getcwd(), "workloads"),
         help="Workload output directory (default: workloads/)")
+    create_workload_parser.add_argument(
+        "--total-docs",
```
gkamat, line 186: Would rename this to `number-of-docs`, since it is not a "total".

osbenchmark/benchmark.py
```diff
@@ -902,6 +912,7 @@ def dispatch_sub_command(arg_parser, args, cfg):
             generate(cfg)
         elif sub_command == "create-workload":
             cfg.add(config.Scope.applicationOverride, "generator", "indices", args.indices)
+            cfg.add(config.Scope.applicationOverride, "generator", "total_docs", args.total_docs)
```
gkamat, line 915: This could be changed to:
```
args.number_of_docs if len(args.number_of_docs) > 0 else None
```
but even that may not be necessary since an empty list matches a `False` predicate.

osbenchmark/workload_generator/corpus.py
```diff
@@ -50,7 +50,7 @@ def get_doc_outpath(outdir, name, suffix=""):
     return os.path.join(outdir, f"{name}-documents{suffix}.json")
 
 
-def extract(client, output_path, index):
+def extract(client, output_path, index, total_docs_requested=None):
```
gkamat, line 53: Again, please change to `number_of_docs_requested` or something equivalent.

osbenchmark/workload_generator/corpus.py
```diff
@@ -62,7 +62,14 @@ def extract(client, output_path, index):
 
     logger = logging.getLogger(__name__)
 
-    total_docs = client.count(index=index)["count"]
+    total_docs_found = client.count(index=index)["count"]
```
gkamat, line 65: This variable could be left as is: `total_docs`.

osbenchmark/workload_generator/corpus.py
```diff
@@ -62,7 +62,14 @@ def extract(client, output_path, index):
...
+    if total_docs_requested is None:
+        # If none, just use all documents found
+        total_docs_requested = total_docs_found
+
+    # If total_docs_requested is >= than total_docs_found, it will just use total_docs_found. Other wise, use total_docs_requested
+    total_docs = min(total_docs_requested, total_docs_found)
+
```
gkamat, line 72:
```
ndocs_to_extract = total_docs if not number_of_docs_requested else min(total_docs, number_of_docs_requested)
```

osbenchmark/workload_generator/workload_generator.py
```diff
@@ -41,10 +41,24 @@ def process_template(templates_path, template_filename, template_vars, output_pa
...
+    if docs_were_requested and len(indices) != len(total_docs):
+        raise exceptions.SystemSetupError("Both --indices and --total-docs were provided but missing corresponding indices or total docs.")
+
+def extract_mappings_and_corpora(client, output_path, indices_to_extract, total_docs_requested):
     indices = []
     corpora = []
+    docs_were_requested = total_docs_requested is not None and len(total_docs_requested) > 0
+
+    check_if_indices_and_total_docs_match(indices_to_extract, total_docs_requested, docs_were_requested)
+
```
gkamat, line 61: Should likely work just with:
```
     check_if_indices_and_total_docs_match(indices_to_extract, total_docs_requested, number_of_docs_requested)
```

gkamat (changes requested):

gkamat:
Also consider modifying the user interface to options of this sort: `--number-of-docs idx1:count1 --number-of-docs idx2:count2`  which might be more intuitive.

osbenchmark/workload_generator/corpus.py
IanHoang: This is cleaner. Will use this.

osbenchmark/workload_generator/workload_generator.py
IanHoang: The function `check_if_indices_and_total_docs_match()` takes in a list of indices to extract docs from, a list of doc count to take from each index, and a boolean of whether docs were requested. I included the last parameter `docs_were_requested` since it's used in different places in the method and would reduce the amount of checks we have to rewrite.

osbenchmark/workload_generator/workload_generator.py
```diff
@@ -56,7 +68,9 @@ def extract_mappings_and_corpora(client, output_path, indices_to_extract):
 
     # That list only contains valid indices (with index patterns already resolved)
     for i in indices:
-        c = corpus.extract(client, output_path, i["name"])
+        docs_to_extract = number_of_docs_requested.get(i["name"], None) if docs_were_requested else None
+        docs_to_extract = int(docs_to_extract) if docs_to_extract is not None else docs_to_extract
```
gkamat, line 72: It would perhaps be stylistically better to do this as an `if`:
```
if docs_were_requested:
   [...]
```

osbenchmark/workload_generator/corpus.py
```diff
@@ -62,24 +62,27 @@ def extract(client, output_path, index):
 
     logger = logging.getLogger(__name__)
 
-    total_docs = client.count(index=index)["count"]
-    if total_docs > 0:
-        logger.info("[%d] total docs in index [%s].", total_docs, index)
+    number_of_docs = client.count(index=index)["count"]
```
gkamat, line 65: I could perhaps have been clearer in my previous review, but leaving this variable as `total_docs` might be better since it is the total doc count.  On the other hand, the `number_of_docs_requested` and `ndocs_to_extract` variables are self-explanatory and appropriate.

osbenchmark/workload_generator/corpus.py
```diff
@@ -62,24 +62,27 @@ def extract(client, output_path, index):
...
+
+    ndocs_to_extract = number_of_docs if not number_of_docs_requested else min(number_of_docs, number_of_docs_requested)
+
+    if ndocs_to_extract > 0:
+        logger.info("[%d] total docs in index [%s].", number_of_docs, index)
         docs_path = get_doc_outpath(output_path, index)
-        dump_documents(client, index, get_doc_outpath(output_path, index, "-1k"), min(total_docs, 1000), " for test mode")
-        dump_documents(client, index, docs_path, total_docs)
-        return template_vars(index, docs_path, total_docs)
+        dump_documents(client, index, get_doc_outpath(output_path, index, "-1k"), min(ndocs_to_extract, 1000), " for test mode")
```
gkamat, line 72: IMO, the doc count specified should not be respected for test mode -- it should always be min(1000, total_docs).  Users generally don't know the internals of test mode and it is unlikely they intend to specify that value.

osbenchmark/workload_generator/workload_generator.py
```diff
@@ -42,10 +42,22 @@ def process_template(templates_path, template_filename, template_vars, output_pa
...
-
-def extract_mappings_and_corpora(client, output_path, indices_to_extract):
+def check_if_indices_and_number_of_docs_match(indices, number_of_docs, docs_were_requested):
+    '''
+    if key(s) and value(s) were provided to --number-of-docs,
+    the number of pairs need to be less than or equal to number of indices
+    '''
+    if docs_were_requested and len(indices) < len(number_of_docs):
+        raise exceptions.SystemSetupError("Number of key:value pairs exceeds number of indices mentioned in --indices. " +
+                                          "Ensure it is less than or equal to.")
```
gkamat, line 52: It would probably suffice to check if the index name in each K:V pair matches one of the indices in the `--indices` parameter and that would be more stringent.

gkamat (approved):

IanHoang (commented):
Reviewed the changes and it looks good to me. Thanks for moving this into a separate file!

PR #292, opened 2023-05-05
Create workloads with default query or custom queries

### LINKED ISSUE #290: [FEATURE] Add support for adding basic queries to create-workload
Currently, when users create workloads using the OSB subcommand `create-workload`, the resulting workload only has a few operations, which are delete index, create index, check cluster health, and bulk ingestion. It requires users to manually add the queries themselves. Example provided below:
```
{% import "benchmark.helpers" as benchmark with context %}
{
  "version": 2,
  "description": "Tracker-generated workload for movies-2000",
  "indices": [
    {
      "name": "movies-2000",
      "body": "movies-2000.json"
    }
  ],
  "corpora": [
    {
      "name": "movies-2000",
      "documents": [
        {
          "target-index": "movies-2000",
          "source-file": "movies-2000-documents.json.bz2",
          "document-count": 2000,
          "compressed-bytes": 15061,
          "uncompressed-bytes": 287917
        }
      ]
    }
  ],
  "schedule": [
    {
      "operation": "delete-index"
    },
    {
      "operation": {
        "operation-type": "create-index",
        "settings": {{index_settings | default({}) | tojson}}
      }
    },
    {
      "operation": {
        "operation-type": "cluster-health",
        "index": "movies-2000",
        "request-params": {
          "wait_for_status": "{{cluster_health | default('green')}}",
          "wait_for_no_relocating_shards": "true"
        },
        "retry-until-success": true
      }
    },
    {
      "operation": {
        "operation-type": "bulk",
        "bulk-size": {{bulk_size | default(5000)}},
        "ingest-percentage": {{ingest_percentage | default(100)}}
      },
      "clients": {{bulk_indexing_clients | default(8)}}
    }
  ]
}
```

It would be nice if OSB `create-workload` came with a default `match_all` query and users could override this default by specifying a JSON file with queries they would like to use. OSB would use this JSON file to and incorporate it into the workload.

**Describe alternatives you've considered**

**Additional context**

IanHoang:
### Current situation
OSB's `create-workload` subcommand creates a custom workload based off of a specific index in a cluster. However, it only generates operations: 
- delete index
- create index
- cluster health check
- bulk ingestion into index

Users will benefit if the command is extended and allows users to provide queries to use. If no queries are specified, then the workload should come with a default query, such as `match_all` query, which is the simplest query that matches all documents in the index.

### This PR adds the following changes:
- For Jinja2 best practices, have renamed the current `workload.json.j2` to be `base-workload.json.j2`. This acts as a parent template that child templates -- `default-query-workloads.json.j2` and `custom-query-workloads.json.j2` -- inherit from. 
-if `create-workloads` subcommand invoked without `--custom-queries`, the `default-query-workloads.json.j2` template is used. This just appends a `match_all` query to the end of `base-workload.json.j2`
- if `create-workloads` subcommand invoked with `--custom-queries`, the `custom-query-workloads.json.j2` template is used and appends all the queries the user has provided to the end of `base-workload.json.j2`. This parameter requires a JSON file that contains an array of queries like the following. Each query entry in the array needs to have a `name`, `operation-type`, and a `body`.
```
# Example of queries to provide to --custom-queries parameter
[
  {
    "name": "<NAME>",
    "operation-type": "<OPERATION-TYPE>",
    "body": {
             <BODY OF QUERY>
    }
  },
  {
    "name": "<NAME>",
    "operation-type": "<OPERATION-TYPE>",
    "body": {
            <BODY OF QUERY>
    }
  }
]
```
### Issues Resolved
#290 

### Testing

Generated various workloads with both default query and custom queries tailored to corpus data. Tested out the workloads after creating.
---

### DIFF
--- a/osbenchmark/workload_generator/workload_generator.py
+++ b/osbenchmark/workload_generator/workload_generator.py
@@ -24,6 +24,7 @@
 
 import logging
 import os
+import json
 
 from opensearchpy import OpenSearchException
 from jinja2 import Environment, FileSystemLoader, select_autoescape
@@ -61,6 +62,13 @@ def extract_mappings_and_corpora(client, output_path, indices_to_extract):
 
     return indices, corpora
 
+def process_custom_queries(custom_queries):
+    with custom_queries as queries:
+        logger = logging.getLogger(__name__)
+        data = json.load(queries)
+        logger.info("DATA: %s", data)
+
+    return data
 
 def create_workload(cfg):
     logger = logging.getLogger(__name__)
@@ -70,6 +78,10 @@ def create_workload(cfg):
     root_path = cfg.opts("generator", "output.path")
     target_hosts = cfg.opts("client", "hosts")
     client_options = cfg.opts("client", "options")
+    unprocessed_custom_queries = cfg.opts("workload", "custom_queries")
+
+    custom_queries = process_custom_queries(unprocessed_custom_queries)
+    logger.info("Custom Queries: %s", type(custom_queries))
 
     logger.info("Creating workload [%s] matching indices [%s]", workload_name, indices)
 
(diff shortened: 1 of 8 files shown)


osbenchmark/workload_generator/workload_generator.py
```diff
@@ -61,6 +62,14 @@ def extract_mappings_and_corpora(client, output_path, indices_to_extract):
 
     return indices, corpora
 
+def process_custom_queries(custom_queries):
+    if not custom_queries:
+        return []
+
+    with custom_queries as queries:
+        data = json.load(queries)
```
gkamat, line 70: It would be good to have some error reporting here, indicating the user's queries is not valid JSON, rather than relying upon the `json` module's error messages.

gkamat (approved):
Please add some documentation on this feature in a README file, with some examples.  Also, include the type of error messages to expect if a user provides incorrectly formatted input.  And, eventually when there is documentation, a pointer to the relevant section should then be added, so that they know how to add more queries, once the workload has been generated.

PR #298, opened 2023-05-17
Simplify Dockerfile and build OSB from Pypi

### LINKED ISSUE #293: Simplify OSB Docker Image

OSB Docker Image currently copies over the github repository directories. It's also unwieldy. 

 Since PyPi and Dockerhub publishing workflows on Jenkins happen in chronological order (PyPi first then Dockerhub second), we can just install the PyPi package instead of git cloning and running `python3 -m pip install -e .`.

IanHoang:
### Description
This PR updates the Dockerfile to build OSB from Pypi and not from scratch. The previous Dockerfile content has been moved into a separate Dockerfile called Dockerfile-development. I've updated the READMe as wel.

### Issues Resolved
#293 

### Testing

- Tested several OSB commands after building Docker image from new Dockerfile
- Ran a geonames test in test mode in Docker image

---


rishabh6788:
LGTM. Coulf you confirm if the `expand-corpus` feature is also working as expected? @IanHoang

rishabh6788:
@peterzhuamazon Could you please take a look as well?

IanHoang:
> LGTM. Coulf you confirm if the `expand-corpus` feature is also working as expected? @IanHoang

I did a `expand-data-corpus.py --help` and it showed up correctly.

docker/Dockerfile
```diff
@@ -1,52 +1,18 @@
...
-
 FROM python:3.11.2-slim
+ARG VERSION
+ARG BUILD_DATE
 
 ENV BENCHMARK_RUNNING_IN_DOCKER True
 
 RUN apt-get -y update && \
-    apt-get install -y curl git pbzip2 pigz && \
+    apt-get install -y curl git gcc pbzip2 pigz && \
```
gkamat, line 8: Where is `gcc` used -- why didn' t it appear in the previous Dockerfile?

docker/Dockerfile
```diff
@@ -1,52 +1,18 @@
-################################################################################
-# Build stage 0 `builder`:
-# Install OpenSearch Benchmark from source inside a virtualenv
-################################################################################
```
gkamat, line 4: Some comments would be helpful -- "Install OpenSearch Benchmark from PyPI to build a Docker image", or something similar.

docker/Dockerfile
```diff
@@ -57,9 +23,6 @@ USER 1000
 
 ENV PATH=/opensearch-benchmark/venv/bin:$PATH
```
gkamat, line 24: Is this correct?  I would assume `pip` would install OSB into the user (or system library) and there would be no `venv` in this case?

docker/README.md
```diff
@@ -24,7 +24,7 @@ docker run -it --entrypoint expand-data-corpus.py opensearchstaging/opensearch-b
 # Building a Copy of OpenSearch Benchmark Image
 1. Git clone OpenSearch Benchmark Github repository
 2. Make changes locally
-3. Run `docker build -t <docker image name>:<optional tag> -f ./docker/Dockerfile .`
+3. Run `docker build --build-arg VERSION=<PYPI VERSION> -t <docker image name>:<optional tag> -f ./docker/Dockerfile .`. Ensure that build arg VERSION takes in an existing pypi version of OSB.
```
gkamat, line 27: A reference to the scenario where `Dockerfile-development` would be useful will help the reader understand why it is provided and which option to choose.

docker/Dockerfile
IanHoang: I included it since the base layer image had it installed in the `Docker-development` file. 
https://github.com/opensearch-project/opensearch-benchmark/blob/61d47687f38f2bda85842eeab2af735e63e48cee/docker/Dockerfile#L9

docker/Dockerfile
IanHoang: This is correcet. When I remove the line, it doesn't change anything else except for when I check `which opensearch-benchmark`, it preappends `opensearch-benchmark/venv/bin:/` to the `$PATH`. 
```
$ echo $PATH
/opensearch-benchmark/venv/bin:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

gkamat (approved):

docker/Dockerfile
```diff
@@ -1,52 +1,22 @@
...
-# Copy OpenSearch Benchmark from stage 0 and fix permissions to support randomized UIDs
-# Define VOLUME for ~/.benchmark
-################################################################################
+###############################################################################
+# Install OpenSearch Benchmark from PyPI to build a Docker image
+###############################################################################
 
 FROM python:3.11.2-slim
+ARG VERSION
+ARG BUILD_DATE
```
peterzhuamazon, line 7: Move BUILD_DATE to the bottom before label can improve docker cache during builds.

docker/Dockerfile
```diff
@@ -1,52 +1,22 @@
...
 
 ENV BENCHMARK_RUNNING_IN_DOCKER True
 
 RUN apt-get -y update && \
-    apt-get install -y curl git pbzip2 pigz && \
+    apt-get install -y curl git gcc pbzip2 pigz && \
```
peterzhuamazon, line 12: Why need gcc?

docker/Dockerfile
```diff
@@ -55,11 +25,6 @@ RUN mkdir -p /opensearch-benchmark/.benchmark && \
 
 USER 1000
 
-ENV PATH=/opensearch-benchmark/venv/bin:$PATH
-
-ARG VERSION
-ARG BUILD_DATE
-
```
peterzhuamazon, line 62: We recently improved our build repo docker file by adding the dockerfile link to label, can you do the same as here: https://github.com/opensearch-project/opensearch-build/blob/main/docker/release/dockerfiles/opensearch.al2.dockerfile#L109?

Thanks.

docker/README.md
```diff
@@ -24,7 +24,13 @@ docker run -it --entrypoint expand-data-corpus.py opensearchstaging/opensearch-b
 # Building a Copy of OpenSearch Benchmark Image
 1. Git clone OpenSearch Benchmark Github repository
 2. Make changes locally
-3. Run `docker build -t <docker image name>:<optional tag> -f ./docker/Dockerfile .`
+3. Run `docker build --build-arg VERSION=<PYPI VERSION> -t <docker image name>:<optional tag> -f ./docker/Dockerfile .`
```
peterzhuamazon, line 27: Build time is missed here due to the value is coming from the build script in build repo.

peterzhuamazon (commented):
Need to verify if the `VERSION` is present before proceed.
Also need to compare the image size between current and old ones.
The old ones use stages probably reduce the size, but since new one is using pypi, probably better.

docker/Dockerfile
IanHoang: Will remove. It's probably needed for building code from source

IanHoang:
> Need to verify if the `VERSION` is present before proceed. Also need to compare the image size between current and old ones. The old ones use stages probably reduce the size, but since new one is using pypi, probably better.

Here is the breakdown:
```
```

PR #305, opened 2023-05-23
Updating integ tests to use current versions of OpenSearch

gkamat:
### Description
The integration tests are currently using obsolete versions of OpenSearch, 1.0.0 and 1.0.1.  This updates the versions to the newer ones in the v1 and v2 branches.  There is a separate batch of improvements to these tests, which will be checked in post the next OSB release.

### Testing
Ran the integ tests on Ubuntu and Amazon Linux 2.  All passed.

On Mac, most of the tests passed, but there were occasional timeouts with one of the tests when runs were repeated.  This seemed to be the case even prior to these changes, so that will be made more robust separately.

---

### DIFF
--- a/.github/workflows/manual-integ.yml
+++ b/.github/workflows/manual-integ.yml
@@ -44,6 +44,12 @@ jobs:
           distribution: 'adopt'
           java-version: '16'
       - run: echo "JAVA16_HOME=$JAVA_HOME" >> $GITHUB_ENV
+      - name: Install JDK 17
+        uses: actions/setup-java@v3
+        with:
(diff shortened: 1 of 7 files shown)


IanHoang:
Since you ran tests on Amazon Linux 2, Ubuntu, and MacOS, do you have any findings (tips regarding setup, running tests, or instructions specific to an operating system) that will be useful? If so, it'd be good to update the DEVELOPER_GUIDE.md. We could also have a separate document for Testing alone and just link it in the DEVELOPER_GUIDE.md.

IanHoang (approved):

Issue #188, opened 2022-04-28
Increase benchmark-metrics-* index shard count


engechas:
**Is your feature request related to a problem? Please describe.**

If the same OpenSearch metrics store is used for many tests, the document count can reach the 2B limit/shard. 

**Describe the solution you'd like**

The [number of shards used for the metrics index](https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/resources/metrics-template.json#L8) should be increased.

**Describe alternatives you've considered**

Another alternative is to create the index [more frequently than monthly](https://github.com/opensearch-project/opensearch-benchmark/blob/36742c8fe730dda233b670d58564fd5ce7bc105e/osbenchmark/metrics.py#L810-L812)

IanHoang:
Posting comment from PR to this issue too:

On second thought, we should revert this change. This change potentially impacts other users since this enforces every user's datastore to use a 10 primary and 1 replicas index regardless of cluster configuration. 

I encountered an issue with my personal datastore after this PR was merged in. Prior to the change, my OSB runs were writing metrics to an external OpenSearch datastore. After this change was incorporated, it encountered an issue where  the number of shards (10) was not a multiple of the number of awareness attributes (3). For more details, see issue https://github.com/opensearch-project/opensearch-benchmark/issues/330. I could change the cluster settings to get around this issue but users should not have to change their clusters to meet the requirements made by OSB.

Instead of changing the default values as done in this PR, users can override the default number of primaries and replicas with `datastore.number_of_replicas` and `datastore.number_of_shards` in the datastore config (~/.benchmark/benchmark.ini). This is easy to change and prevents us from messing with the default `benchmark-metrics` settings that can potentially interfere with other users' setups.

PR #321, opened 2023-06-02
Increase the shard and replica count for the benchmark-metrics-* indices in the data store

ayushav12:
### Description
This change will increase the number of primary shards in the monthly index "benchmark-metrics-yyyy-mm" to accomodate more documents and avoid reaching the shard limit of 2B on the metrics data store.

We reached the max. number of 2 billion documents per shard in the monthly "benchmark-metrics-2023-05" index, a limit imposed by lucene, on our metrics data store which we use to rigorously benchmark domain(s). This doc limit was reached within first 3-4 days of the month. 
Considering this amount of data to be created consistently throughout the month , we made the decision to have 10 primary shards. We created a new index with 10 primary shards and re-indexed with new same-index index having 10 primary shards and 1 replica shard to make sure no data loss happens in case of node failures or write block during re-indexing.

### Issues Resolved
https://github.com/opensearch-project/opensearch-benchmark/issues/188

---

### DIFF
--- a/osbenchmark/resources/metrics-template.json
+++ b/osbenchmark/resources/metrics-template.json
@@ -5,7 +5,8 @@
   "settings": {
     "index": {
       "refresh_interval": "5s",
-      "number_of_shards": 1
+      "number_of_shards": 10,
+      "number_of_replicas": 1
     }
   },
   "mappings": {


gkamat (approved):
Please update the comment to something like this:

Increase the shard and replica count for the benchmark-metrics-* indices in the data store.

IanHoang:
Please also add in the description why you have specifically chosen 10 shards for the primaries and 1 for the replicas. This will help future developers understand why this was changed in the past. Also, It might be good to add a brief explanation that clarifies that the 2 billion documents per shard limit is imposed by lucene.

IanHoang (approved):

IanHoang:
On second thought, we should revert this change. This change potentially impacts other users since this enforces every user's datastore to use a 10 primary and 1 replicas index regardless of cluster configuration. 

I encountered an issue with my personal datastore after this PR was merged in. Prior to the change, my OSB runs were writing metrics to an external OpenSearch datastore. After this change was incorporated, it encountered an issue where  the number of shards (10) was not a multiple of the number of awareness attributes (3). For more details, see issue https://github.com/opensearch-project/opensearch-benchmark/issues/330. I could change the cluster settings to get around this issue but users should not have to change their clusters to meet the requirements made by OSB.

Instead of changing the default values as done in this PR, users can override the default number of primaries and replicas with `datastore.number_of_replicas` and `datastore.number_of_shards` in the datastore config (~/.benchmark/benchmark.ini). This is easy to change and prevents us from messing with the default `benchmark-metrics` settings that can potentially interfere with other users' setups.

PR #338, opened 2023-06-16
Fix issues with the git module

gkamat:
### Description
There were several errors with the `git` module.  The version check did not actually work.  Also, the error message generated when OSB atarted up if the git package was not installed, was not user-friendly:
```
ERROR] Cannot execute-test. Error in test execution orchestrator ([Errno 2] No such file or directory: 'git')
```

### Testing
The git module tests were updated to incorporate the code changes.  There were some errors as well in the earlier tests, which resulted in some of then not actually working.

---

### DIFF
--- a/osbenchmark/utils/git.py
+++ b/osbenchmark/utils/git.py
@@ -22,25 +22,29 @@
 # specific language governing permissions and limitations
 # under the License.
 
-import logging
 import os
+import re
 
 from osbenchmark import exceptions
 from osbenchmark.utils import io, process
 
+MIN_REQUIRED_MAJOR_VERSION = 2
+VERSION_REGEX = r'.* ([0-9]+)\.([0-9]+)\..*'
 
 def probed(f):
     def probe(src, *args, **kwargs):
         # Probe for -C
-        if not process.exit_status_as_bool(lambda: process.run_subprocess_with_logging(
-                "git -C {} --version".format(io.escape_path(src)), level=logging.DEBUG), quiet=True):
-            version = process.run_subprocess_with_output("git --version")
-            if version:
-                version = str(version).strip()
-            else:
-                version = "Unknown"
-            raise exceptions.SystemSetupError("Your git version is [%s] but Benchmark requires at least git 1.9. "
-            "Please update git." % version)
+        try:
+            out, _, status = process.run_subprocess_with_out_and_err(
+                "git -C {} --version".format(io.escape_path(src)))
+        except FileNotFoundError:
+            status = 1
+        if status != 0:
+            raise exceptions.SystemSetupError("Error invoking 'git', please install (or re-install).")
+        match = re.search(VERSION_REGEX, out)
+        if not match or int(match.group(1)) < MIN_REQUIRED_MAJOR_VERSION:
+            raise exceptions.SystemSetupError("OpenSearch Benchmark requires at least version 2 of git.  "
+                                              f"You have {out}.  Please update git.")
         return f(src, *args, **kwargs)
     return probe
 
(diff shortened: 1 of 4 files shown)


osbenchmark/utils/git.py
```diff
@@ -32,15 +33,16 @@
...
-            "Please update git." % version)
+        try:
+            out, err, status = process.run_subprocess_with_out_and_err(
+                "git -C {} --version".format(io.escape_path(src)))
+        except FileNotFoundError:
+            status = 1
+        if status != 0:
+            raise exceptions.SystemSetupError("Error invoking 'git', please install (or re-install).")
+        match = re.search(r'.* ([0-9]+)\.([0-9]+)\..*', out)
+        if not match or int(match.group(1)) < 2:
```
IanHoang, line 44: Instead of having `2` hardcoded here and in the exception, it'd be better to include it as a global variable, such as `REQUIRED_GIT_MAJOR_VER` at the top of this file and referenced in the if statement and exception. 
```
if not match or int(match.group(1)) < REQUIRED_GIT_MAJOR_VER:
    raise exceptions.SystemSetupError(f"OpenSearch Benchmark requires at least version {REQUIRED_GIT_MAJOR_VER} of git.  You have {out}.  Please update git.")
```

osbenchmark/utils/process.py
```diff
@@ -52,15 +52,26 @@ def run_subprocess_with_output(command_line):
...
+def run_subprocess_with_stderr(command_line):
+    logger = logging.getLogger(__name__)
+    logger.debug("Running subprocess [%s] with stderr but no stdout.", command_line)
     command_line_args = shlex.split(command_line)
 
     sp = subprocess.Popen(command_line_args, stdout=subprocess.DEVNULL, stderr=subprocess.PIPE)
     sp.wait()
     _, err = sp.communicate()
-    return err.decode("UTF-8"), sp.returncode
+    return err.decode('UTF-8'), sp.returncode
```
IanHoang, line 74: Just curious, why was this changed to single quotes? Aren't they equivalent in Python?

IanHoang (changes requested):
Provided some comments above. Also, CI Unit Tests are failing due to pylint. Should be a quick fix and a quick repush. You can run `make pylint` locally after making the changes to ensure all have been addressed before pushing up to this branch again. After that, I can approve!

osbenchmark/utils/process.py
gkamat: Yes, they are.  Generally, single quotes are used for short strings, double quotes for longer ones.

osbenchmark/utils/git.py
gkamat: Good point.  The earlier version was hardcoded at 1.9, but this is cleaner.

osbenchmark/utils/git.py
```diff
@@ -22,25 +22,29 @@
...
+        try:
+            out, _, status = process.run_subprocess_with_out_and_err(
+                "git -C {} --version".format(io.escape_path(src)))
+        except FileNotFoundError:
+            status = 1
+        if status != 0:
+            raise exceptions.SystemSetupError("Error invoking 'git', please install (or re-install).")
+        match = re.search(VERSION_REGEX, out)
+        if not match or int(match.group(1)) < MIN_REQUIRED_MAJOR_VERSION:
+            raise exceptions.SystemSetupError("OpenSearch Benchmark requires at least version 2 of git.  "
```
IanHoang, line 46: Super nit but now that we have a global variable, we can incorporate it into the message too
```
raise exceptions.SystemSetupError(f"OpenSearch Benchmark requires at least version {MIN_REQUIRED_MAJOR_VERSION } of git."
```

IanHoang (approved):

Issue #354, opened 2023-08-10
Add support to OSB for serverless OpenSearch implementations


gkamat:
**Is your feature request related to a problem? Please describe.**

Serverless OpenSearch implementations don't support the full complement of OpenSearch APIs, since details of the internals such as node counts, allocation, shard locations, etc. are abstracted away.  For instance, the cluster-health API may not be available.  OSB will not start a test if the health check fails.

**Describe the solution you'd like**

Enhance OSB to add support for serverless OpenSearch implementations.

Changes will need to bypass the cluster health check in favor of some other test.  In addition, features like telemetry and node stats will likely not be available and will need to be handled appropriately.

IanHoang:
This has been implemented into OSB 1.1.0

saikatsarkar056:
@gkamat, @IanHoang: I am trying to run [OpenSearch Benchmark with AWS Signature Version 4](https://opensearch.org/docs/latest/benchmark/tutorials/sigv4/) for serverless opensearch. I am trying to use the following command:

```
opensearch-benchmark execute-test \
--target-hosts=<serverless_opensearch_endpoint> \
--pipeline=benchmark-only \
--workload=geonames \
--client-options=timeout:120,amazon_aws_log_in:environment
```

I am getting 403 forbidden. 

According to [benchmark using sigv4](https://opensearch.org/docs/latest/benchmark/tutorials/sigv4/), target-hosts is <CLUSTER ENDPOINT>. But, there is no concept of cluster endpoint in serverless opensearch collection. Instead, it provides an opensearch endpoint.

Can you please help me understand what I am missing here?

PR #355, opened 2023-08-10
Changes for serverless implementations which may not support the full complement of OpenSearch APIs

gkamat:
### Description
Serverless OpenSearch implementations don't support the full complement of OpenSearch APIs, since details of the internals such as node counts, allocation, shard locations, etc. are abstracted away.  For instance, the cluster-health API may not be available.  The changes here ensure that OSB will work for such systems.  There may be more changes required to ensure that flags like telemetry continue to function, but those issues will be handled in a subsequent change.

### Issues Resolved
https://github.com/opensearch-project/opensearch-benchmark/issues/354

### Testing
- Added a test to exercise the bypassing of the cluster health API call
- Ran unit and integ tests

---


IanHoang (approved):
Changes look good to me. We should consider creating documentation in the future with some basic guidance on how users can run this with available serverless offerings.

Issue #339, opened 2023-06-20
Capture and report replication lag metric


mch2:
**Is your feature request related to a problem? Please describe.**

OpenSearch has recently launched Segment Replication.  Segrep includes a new API exposing metrics specific to replication performance.  The most critical of these metrics is replication lag, which measures the time between a primary refreshing on a new set of segments and a replica refreshing on the same set of segments.  It would be valuable to capture and report on this metric when benchmarking using the segment replication strategy.

The service currently returns the last completed lag and any ongoing lag (shard is currently syncing to a new set of segments) through its /_cat/segment_replication API.

**Describe the solution you'd like**

Capture and report the min/max/avg lags for a benchmark run.

**Describe alternatives you've considered**
Aggregate the metrics service side and invoke the API  to fetch after a benchmark is completed.

**Additional context**

https://opensearch.org/docs/latest/api-reference/cat/cat-segment-replication/

tlfeng:
I will work on this feature request to add new telemeter device of segment replication statistic by adding codes to the file https://github.com/opensearch-project/opensearch-benchmark/blob/1.0.0/osbenchmark/telemetry.py#L43 to capture response of `/_cat/segment_replication` API.

gkamat:
Please raise a PR when your work is complete.  Thanks.

tlfeng:
Temporarily put the code I wrote here https://github.com/tlfeng/opensearch-benchmark/commit/f2eaf20994d924fca9103fbda4cea8382dc457fb
For unknown reason, it's having the below error. I remembered there was no error on last Friday, but I forgot what changed which caused the error keeps showing.
```
2023-07-10 17:35:59,56 ActorAddr-(T|:36707)/PID:156715 osbenchmark.telemetry ERROR Could not determine segment replication stats
Traceback (most recent call last):

  File "/home/ftianli/github/opensearch-benchmark/osbenchmark/telemetry.py", line 161, in run
    self.recorder.record()

  File "/home/ftianli/github/opensearch-benchmark/osbenchmark/telemetry.py", line 1767, in record
    stats_api_endpoint = "/_cat/segment_replication"

NameError: name 'index' is not defined
```

tlfeng:
The above mysterious error got solved after restarting the system, it might be an issue of the python environment. 😅
[...]

PR #356, opened 2023-08-10
Improve the telemetry device "Segment Replication Stats" by changing to parse JSON format and using numeric bytes value

tlfeng:
### Description

Although PR https://github.com/opensearch-project/opensearch-benchmark/issues/339 added new telemetry device "Segment Replication Stats", there are some problems. The PR will solve the problem.
- Fix the problem of using single whitespace `str.split(" ")` as the delimiter to parse the REST API response, where maybe more than 1 space between adjacent values.
- Change to parse the REST API response in JSON format to improve the robustness.
- Change the byte value field to be purely numeric to be directly comparable.

For detail, see https://github.com/opensearch-project/opensearch-benchmark/issues/339#issuecomment-1672297062.

### Issues Resolved
Address a problem related to issue https://github.com/opensearch-project/opensearch-benchmark/issues/339

The screenshot illustrates the current problem. Sometime when there are more than 1 space between adjacent values, it leads mismatch in the fields and values of the result.
Problem with parsing the API response

### Testing

End to end test:
Verified that every field has correct value.

Command used: `opensearch-benchmark execute-test --workload=http_logs --pipeline=benchmark-only \
--target-hosts=opens-clust-xxx.elb.us-west-2.amazonaws.com:80 \
--include-tasks=delete-index,create-index,check-cluster-health,index-append \
--workload-params='{"index_settings":{"number_of_shards": 1, "number_of_replicas": 1 }}' \
--telemetry=segment-replication-stats --test-mode `

image

image

image

---


tlfeng:
@mch2 @IanHoang @gkamat Looking forward to your opinions. Thank you! I'd like to make this change into the next release of version 1.1.0 .

tlfeng:
End to end test:
Verified that every field has correct value.
image

image

image

tests/telemetry_test.py
```diff
@@ -3431,27 +3449,27 @@ def test_stores_default_stats(self, metrics_store_put_doc):
 
         metrics_store_put_doc.assert_has_calls([call({
             "name": "segment-replication-stats",
-            "shard_id": "[so][0]",
-            "target_node": "node-1",
-            "target_host": "127.0.0.1",
+            "shard_id": "[logs-211998][6]",
```
dreamer-89, line 3452: nit: Is `shard_id` change needed here ?

dreamer-89 (approved):

tests/telemetry_test.py
tlfeng: @dreamer-89 Thanks a lot for your review! 👏 The change is indeed not necessary, I just updated the test with a real API response from my recent performance test. I think it's fine to keep the changes.

tlfeng:
Manually tested again after the last commit
image
image

IanHoang (approved):
LGTM. Thanks for discovering this bug and providing the fix @tlfeng!

PR #374, opened 2023-09-20
Update README and Developer Guide

IanHoang:
### Description
Shorten readme and improve developer guide

### Issues Resolved
N/A

### Testing
N/A

---

### DIFF
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -42,7 +42,7 @@ To develop OSB properly, it is recommended that you fork the official OpenSearch
 
 After you git cloned the forked copy of OpenSearch Benchmark, use the following command-line instructions to set up OpenSearch Benchmark for development:
 ```
-cd OpenSearch-Benchmark
+cd opensearch-benchmark
 make prereq
 make install
 ```
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -98,8 +98,9 @@ Invoke integration tests by running the following command within the root direct
 
 ```
 make it
-
 ```
+
+Integration tests are expected to run for approximately 20-30 mins.
 
 ## Submitting your changes for a pull request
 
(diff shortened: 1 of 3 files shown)


DEVELOPER_GUIDE.md
```diff
@@ -2,30 +2,46 @@
...
 
 ### Setup
 
-Use the following command-line instructions to set up OpenSearch Benchmark for development:
+To develop OSB properly, it is recommended that you fork the official OpenSearch Benchmark repository.
+
+After you git cloned the forked copy of OpenSearch Benchmark, use the following command-line instructions to set up OpenSearch Benchmark for development:
 ```
-git clone https://github.com/opensearch-project/OpenSearch-Benchmark
 cd OpenSearch-Benchmark
```
gkamat, line 45: It may be better to use lowercase for consistency, since the GitHub project name is in lowercase.  Else, there can be issues on case-sensitive systems if there is already a directory in the wrong case.

DEVELOPER_GUIDE.md
```diff
@@ -80,6 +94,7 @@ Integration tests can be run on the following operating systems:
   * Amazon Linux 2
   * MacOS
 
+Invoke integration tests by running the following command within the root directory of the repository:
```
gkamat, line 97: Would be good to note the amount of time the user should expect to run the integ tests.

osbenchmark/results_publisher.py
```diff
@@ -130,6 +130,8 @@ def publish(self):
 
         for record in stats.op_metrics:
             task = record["task"]
+            self.logger = logging.getLogger(__name__)
+            self.logger.info("Record: [%s], Task: [%s]", record, task)
```
gkamat, line 134: Might be better to have this in a separate commit.

gkamat (approved):
Minor changes suggested -- but they can be addressed in a future change.

osbenchmark/results_publisher.py
IanHoang: Good catch, let me remove this!

DEVELOPER_GUIDE.md
IanHoang: Good catch!

Issue #366, opened 2023-08-22
[Refactor] Rename put-pipeline to create-ingest-pipeline


noCharger:
We enabled [the creation of search pipelines](https://github.com/opensearch-project/opensearch-benchmark/pull/364.) in opensearch-benchmark with the introduction of search pipelines in OS 2.9.0. To separate from the search pipeline, the ingest pipeline API put-pipeline should be refactored.

Dantochi:
Hey, I would like to work on the first task which is: Refactor put-pipeline https://github.com/opensearch-project/opensearch-benchmark/issues/366...

IanHoang:
@Dantochi if you'd like to contribute to this issue, please setup an OpenSearch Benchmark in Developer Mode. You can [see the Developer Guide] (https://github.com/opensearch-project/OpenSearch-Benchmark/blob/main/DEVELOPER_GUIDE.md) for reference.

maddox05:
Hi! Saw the slack message and am ready to work on it.

PR #399, opened 2023-10-25
renamed put pipeline to create ingest pipeline

maddox05:
### Description
I have renamed the fields put-pipeline and all according names to create ingest pipeline.

### Issues Resolved
Solves issue  #366 

### Testing
No testing should be needed for renamed variables/classes.

Looked over by me and others to make sure 100% I got the name correct ( will be looked over again in PR )

---

### DIFF
--- a/osbenchmark/worker_coordinator/runner.py
+++ b/osbenchmark/worker_coordinator/runner.py
@@ -1101,7 +1101,7 @@ class CreateIngestPipeline(Runner):
                                      )
 
     def __repr__(self, *args, **kwargs):
-        return "create-ingest-pipeline"
+        return opensearch.ingest.put_pipeline.return_value = as_future()
 
 # TODO: refactor it after python client support search pipeline https://github.com/opensearch-project/opensearch-py/issues/474
 class CreateSearchPipeline(Runner):
--- a/osbenchmark/worker_coordinator/runner.py
+++ b/osbenchmark/worker_coordinator/runner.py
@@ -72,7 +72,7 @@ def register_default_runners():
     register_runner(workload.OperationType.RestoreSnapshot, RestoreSnapshot(), async_runner=True)
     # We treat the following as administrative commands and thus already start to wrap them in a retry.
     register_runner(workload.OperationType.ClusterHealth, Retry(ClusterHealth()), async_runner=True)
-    register_runner(workload.OperationType.PutPipeline, Retry(PutPipeline()), async_runner=True)
+    register_runner(workload.OperationType.CreateIngestPipeline, Retry(CreateIngestPipeline()), async_runner=True)
     register_runner(workload.OperationType.Refresh, Retry(Refresh()), async_runner=True)
     register_runner(workload.OperationType.CreateIndex, Retry(CreateIndex()), async_runner=True)
     register_runner(workload.OperationType.DeleteIndex, Retry(DeleteIndex()), async_runner=True)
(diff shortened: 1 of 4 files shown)


maddox05:
Ok I would like to talk more about this in our meeting, but I think I understand where I went wrong with how your test cases work.

IanHoang:
@maddox05 Nice! To fix the unittests, you just need to update lines 2248's and 2265's error messages in `runner_test.py` by replacing `put-pipeline` with `create-ingest-pipeline`. Currently, those unit tests are expecting the term `put-pipeline` in the raised error messages but are actually receiving `create-ingest-pipeline`.

IanHoang (changes requested):
Thanks for the quick changes! Just left a comment on how to quickly fix the unittests

maddox05:
could not find how -- opensearch.ingest.put_pipeline.return_value = as_future() so I changed it to -- opensearch.ingest.create_ingest_pipeline.return_value = as_future()

IanHoang:
> opensearch.ingest.put_pipeline.return_value = as_future()

@maddox05 This line is included in the unittests but it's referring to the opensearchpy library's `put-pipeline` and doesn't need to be changed. See the following snippet of code from opensearchpy repository:
https://github.com/opensearch-project/opensearch-py/blob/0d8a23dd78803d2712aa7d8e92663b0953158ba1/opensearchpy/client/ingest.py#L62

The following line occurs twice in runner_test.py:
```
        with self.assertRaisesRegex(exceptions.DataError,
                                    "Parameter source for operation 'put-pipeline' did not provide the mandatory parameter 'body'. "
                                    "Add it to your parameter source and try again."):
```
This is what's causing the unittests to fail. The test is expecting an error message that includes "put-pipeline" but it's getting an error message that includes "create-ingest-pipeline". In order to fix the unittests, we just need to replace "put-pipeline" in the error message above to "create-ingest-pipeline".

tests/worker_coordinator/runner_test.py
```diff
@@ -2232,29 +2237,30 @@ async def test_create_pipeline(self, opensearch):
...
             await r(opensearch, params)
 
-        self.assertEqual(0, opensearch.ingest.put_pipeline.call_count)
+        self.assertEqual(0, opensearch.ingest.create_ingest_pipeline.call_count)
 
     @mock.patch("opensearchpy.OpenSearch")
     @run_async
     async def test_param_id_mandatory(self, opensearch):
-        opensearch.ingest.put_pipeline.return_value = as_future()
+        opensearch.ingest.create_ingest_pipeline.return_value = as_future()
```
IanHoang, line 2263: We won't want to change these because this is an opensearchpy function call. We're only changing occurrences of "put-pipeline" in OSB. Let me know if you have more questions about this and we can discuss this on a call or during office hours.

osbenchmark/worker_coordinator/runner.py
```diff
@@ -1101,7 +1101,7 @@ async def __call__(self, opensearch, params):
                                      )
 
     def __repr__(self, *args, **kwargs):
-        return "put-pipeline"
+        return opensearch.ingest.put_pipeline.return_value = as_future()
```
IanHoang, line 1104: We should change this line back as it might be impacting the CI runs

maddox05: not sure why thats left, thought I changed all of them. will change today, sorry about that.

IanHoang: No worries!

IanHoang (approved):

Issue #385, opened 2023-10-05
Add ZSTD Compression and Decompression Support for Corpora


IanHoang:
**Is your feature request related to a problem? Please describe.**

As of now, OSB compresses files with bzip2 (when creating workloads) and can decompress corpora with the following file extensions before running workloads:
- zip
- bzip2
- gz
- tar
- tar.gz
- tgz
- tar.bz2

We should also add support zstd compression and decompression for corpora. When compared to bzip2, ZSTD's compressed files are notably smaller in size. Compared to other decompression algorithms, ZSTD is very fast and takes less time to decompress. Giving users the option to decompress and compress large corpora quickly would be beneficial.

For more insight into compression algorithms, see: https://linuxreviews.org/Comparison_of_Compression_Algorithms#Notable_Takeaways

beaioun:
Hi guys, I'll start to work on this issue and I am open to collaboration. Comment below or DM if you want to work together!

beaioun:
Hey @IanHoang, As I was working with testing, I found that especially for nyc_taxis which has a size of over 70g, it takes a very long time to download and decompress before it can be worked on. I think we could add a zstd compressed version of workloads to the osb-workload repository, for a faster decompression rate.

beaioun:
Hey @IanHoang as I was performing tests, I found that large workloads such as the nyc_taxis, which has a size of over 70g, take a very long time to download and decompress. Since we are now capable of decompressing zstd files, we could create zstd compressed version of the workloads in the workload repo for a faster decompression rate.

IanHoang:
> Hey @IanHoang as I was performing tests, I found that large workloads such as the nyc_taxis, which has a size of over 70g, take a very long time to download and decompress. Since we are now capable of decompressing zstd files, we could create zstd compressed version of the workloads in the workload repo for a faster decompression rate.

@beaioun Thanks for the contribution here. This will be certainly useful. I'll add a good first issue in the workloads repository related to your comment.

PR #410, opened 2023-11-08
Adding zstd compression and decompression support for workload corpora

beaioun:
### Description
This PR adds support for zstd compression and decompression of workload corpora

### Issues Resolved
This PR is aiming to solve issue #385 

### Testing

REQUEST FOR HELP:
I need help understanding compression in the io module. I have not found where the io.compress function is referenced thus not really sure how the zstd compression will work in this case. But I included compress_zstd as a separate function to be called in actual use cases.

---


osbenchmark/utils/io.py
```diff
@@ -344,6 +370,18 @@ def _do_decompress_manually_with_lib(target_directory, filename, compressed_file
         compressed_file.close()
 
 
+def _do_decompress_zstd(target_directory, filename):
```
IanHoang, line 373: Instead of creating a separate function for this, is it possible to implement this through the `_do_decompress_manually` as is done with other decompressors like `bz2`? This would remove boilerplate code

IanHoang:
@beaioun You are correct, there is no io.compress(). As of now, the only time compressing is used is in the metrics.py and corpus.py, but these call the libraries zlib and bz2 directly. Adding a separate function to io.py to compress like you did doesn't hurt. The issue I wrote should've clarified this but I can create a separate issue on this: it'd be good to include different compression options in corpus.py. I'll explain this further in a separate issue since this is out of scope for this PR!

IanHoang (changes requested):
Just need to fix the pylint error in the CI Unittests and we should be good

osbenchmark/utils/io.py
beaioun: Yes I'll be working on that, thanks

beaioun:
Thanks Ian, will do☺️!

beaioun:
> @beaioun You are correct, there is no io.compress(). As of now, the only time compressing is used is in the metrics.py and corpus.py, but these call the libraries zlib and bz2 directly. Adding a separate function to io.py to compress like you did doesn't hurt. The issue I wrote should've clarified this but I can create a separate issue on this: it'd be good to include different compression options in corpus.py. I'll explain this further in a separate issue since this is out of scope for this PR!

Thank you Ian, for the clarification. I'll take a look at corpus.py and see if we can implement that.

beaioun:
> Just need to fix the pylint error in the CI Unittests and we should be good 

Thanks @IanHoang, will do

IanHoang:
> > Just need to fix the pylint error in the CI Unittests and we should be good
> 
> Thanks @IanHoang, will do

@beaioun I think it's fine to go ahead and fix the CI unittests here and we can merge this in. We can open a separate issue for corpus.py and create a separate PR for it. This PR has a good amount of changes

beaioun:
> > > Just need to fix the pylint error in the CI Unittests and we should be good
> > 
> > 
> > Thanks @IanHoang, will do
> 
> @beaioun I think it's fine to go ahead and fix the CI unittests here and we can merge this in. We can open a separate issue for corpus.py and create a separate PR for it. This PR has a good amount of changes

Hey Ian, it's been a few days. I'll go ahead and commit the recent changes. Even though I am still left with this throughput throttled error unfixed, I'd like to run through the test what it has to say. And I will fix it from there.
image

IanHoang (approved):

IanHoang (approved):
Thanks for this!

PR #411, opened 2023-11-08
Switching to the default connection class to address slower connection with RequestsHttpConnection

### LINKED ISSUE #408: [FEATURE]  Replace RequestsHttpConnection by Urllib3HttpConnection
**Describe the bug**

Looking at https://github.com/opensearch-project/opensearch-benchmark/blob/3adacc0a9383cf7f81dbe0cba4c4f6c0b206cab2/osbenchmark/client.py#L296 the code is using `RequestsHttpConnection`, which is much slower than the default `Urllib3HttpConnection `.

The reason is probably Sigv4 support, which was added to `Urllib3HttpConnection` in https://github.com/opensearch-project/opensearch-py/pull/547. Replace `RequestsHttpConnection` with `Urllib3HttpConnection`.

beaioun:
### Description
This PR implements the default Urllib3HttpConnection to replace RequestsHttpConnection to address the slower connection issue, as recommended by the opensearch-py guide for connection classes. Upon checking the code, I discovered this was the only place RequestsHttpConnection was used. Please do point out if I was wrong here.

### Issues Resolved
#408 

### Testing

---

### DIFF
--- a/osbenchmark/client.py
+++ b/osbenchmark/client.py
@@ -294,7 +294,7 @@ class OsClientFactory:
         aws_auth = opensearchpy.AWSV4SignerAuth(credentials, self.aws_log_in_dict["region"],
                                                 self.aws_log_in_dict["service"])
         return opensearchpy.OpenSearch(hosts=self.hosts, use_ssl=True, verify_certs=True, http_auth=aws_auth,
-                                       connection_class=opensearchpy.RequestsHttpConnection)
+                                       connection_class=opensearchpy.Urllib3HttpConnection)
 
     def create_async(self):
         # pylint: disable=import-outside-toplevel
--- a/osbenchmark/client.py
+++ b/osbenchmark/client.py
@@ -291,7 +291,7 @@ class OsClientFactory:
         credentials = Credentials(access_key=self.aws_log_in_dict["aws_access_key_id"],
                                   secret_key=self.aws_log_in_dict["aws_secret_access_key"],
                                   token=self.aws_log_in_dict["aws_session_token"])
-        aws_auth = opensearchpy.AWSV4SignerAuth(credentials, self.aws_log_in_dict["region"],
+        aws_auth = opensearchpy.Urllib3AWSV4SignerAuth(credentials, self.aws_log_in_dict["region"],
                                                 self.aws_log_in_dict["service"])
         return opensearchpy.OpenSearch(hosts=self.hosts, use_ssl=True, verify_certs=True, http_auth=aws_auth,
                                        connection_class=opensearchpy.Urllib3HttpConnection)
--- a/osbenchmark/utils/opts.py
+++ b/osbenchmark/utils/opts.py
@@ -179,7 +179,7 @@ class TargetHosts(ConnectOptions):
             defined as a json string or file.
             """
             # pylint: disable=import-outside-toplevel
-            from opensearchpy.client import _normalize_hosts
+            from opensearchpy.client.utils import _normalize_hosts
             return {TargetHosts.DEFAULT: _normalize_hosts(arg)}
 
         self.parsed_options = to_dict(self.argvalue, default_parser=normalize_to_dict)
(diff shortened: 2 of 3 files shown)


IanHoang (changes requested):
With the new change, it's failing for a unittest that tests client creation with aws credentials. Seems like it's failing in the underlying packages. We'll need to see why this only happens when Urllib3HttpConnection is used and not when RequestsHttpConnection is used.
```
s = <opensearchpy.helpers.signer.AWSV4SignerAuth object at 0x10b73c940>

    def b(s):
>       return s.encode("latin-1")
E       AttributeError: 'AWSV4SignerAuth' object has no attribute 'encode'

.venv/lib/python3.8/site-packages/urllib3/packages/six.py:687: AttributeError
```

beaioun:
Hey @IanHoang , I figured out where the issue was. The Urllib3HttpConnection has to pair with Urllib3AWSV4SignerAuth for http_auth because that is callable, whereas RequestsAWSV4SignerAuth is not and will try to call make_headers in the urllib3 and throws an error for the subsequent actions. Could you try to run the workflow again? Thanks.

rishabh6788:
Also, I don't think `opensearch-py-2.3.2` has support for the new methods you added. 
Can you please update the `opensearch-py` version to 2.4.1 , test it locally and update the PR?
@beaioun

beaioun:
> Also, I don't think `opensearch-py-2.3.2` has support for the new methods you added.
> Can you please update the `opensearch-py` version to 2.4.1 , test it locally and update the PR?

@rishabh6788 I see, I'll update to 2.4.1 see if that works.

rishabh6788:
FYI, there is bug in `opensearch-py-2.4.1` where it is not working with OpenSearch Serverless. 
https://github.com/opensearch-project/opensearch-py/issues/600

osbenchmark/client.py
```diff
@@ -291,10 +291,10 @@ def create(self):
         credentials = Credentials(access_key=self.aws_log_in_dict["aws_access_key_id"],
                                   secret_key=self.aws_log_in_dict["aws_secret_access_key"],
                                   token=self.aws_log_in_dict["aws_session_token"])
-        aws_auth = opensearchpy.AWSV4SignerAuth(credentials, self.aws_log_in_dict["region"],
+        aws_auth = opensearchpy.Urllib3AWSV4SignerAuth(credentials, self.aws_log_in_dict["region"],
                                                 self.aws_log_in_dict["service"])
         return opensearchpy.OpenSearch(hosts=self.hosts, use_ssl=True, verify_certs=True, http_auth=aws_auth,
-                                       connection_class=opensearchpy.RequestsHttpConnection)
+                                       connection_class=opensearchpy.Urllib3HttpConnection)
```
dblock, line 297: FYI this is now the default and you could remove this option if you want to. Doesn't hurt to keep it either.

dblock:
> FYI, there is bug in `opensearch-py-2.4.1` where it is not working with OpenSearch Serverless.
> [opensearch-project/opensearch-py#600](https://github.com/opensearch-project/opensearch-py/issues/600)

This is not the case, it works with AOSS well. The LangChain integration was failing because it [checks for `signer.service` explicitly and that field was removed](https://github.com/langchain-ai/langchain/blob/b4312aac5c0567088353178fb70fdb356b372e12/libs/langchain/langchain/vectorstores/opensearch_vector_search.py#L83).

osbenchmark/client.py
rishabh6788: Thank you @dblock , so if we just use `opensearch-py-2.4.1` then we don't need to explicitly specify `connection_class` parameter?

beaioun:
> Also, I don't think `opensearch-py-2.3.2` has support for the new methods you added. Can you please update the `opensearch-py` version to 2.4.1 , test it locally and update the PR? @beaioun

@rishabh6788 I updated the opensearch-py version to 2.4.1 and fixed some other importing issues with this version. All tests passed on my end. 769ff90bc024efbe725b4d11c14b0781a137d63c

IanHoang (approved):

osbenchmark/client.py
dblock: Yes

Issue #103, opened 2022-01-12
Support Benchmarking K-NN Plugin and Vectorsearch Workload


jmazanec15:
**Is your feature request related to a problem? Please describe.**

The k-NN plugin adds support for a new field type, `knn_vector`, which can be thought of as an array of floating point numbers. The plugin then also adds support for running approximate k nearest neighbor search on these fields.

For benchmarking the plugin, we are interested in several metrics including:
1. Query latency
2. Indexing throughput
3. Refresh time
3. Recall (the ratio of neighbors returned by an Approximate search that are actually in the ground truth nearest neighbors)
4. Training latency
5. Native memory footprint
6. Disk utilization

Currently, we have our own custom code to get these metrics: https://github.com/opensearch-project/k-NN/tree/main/benchmarks/perf-tool. The reason we decided to build our own tool was that we needed some functionality that was not easily available in Rally/OpenSearch Benchmark: ability to compute the recall from a set of queries, integration of our own custom APIs and metrics, using datasets in alternative forms, etc.

**Describe the solution you'd like**

We would prefer to use OpenSearch Benchmarks to collect the metrics above so that we don't have to maintain our own tool and that allows customers to not have to adopt another tool other than OpenSearch Benchmarks. I saw #98 was created and I imagine we may need that in order to reach our goal. I would be interested in helping contribute this feature.

**Describe alternatives you've considered**

The alternative is to continue to use our own benchmarking tool. However, this has several drawbacks mentioned above.

jmazanec15:
@travisbenedict @achitojha Im working on adding  training component from k-NN to a custom runner (i.e. train-model)

I would like a user to be able to specify the body for the train API in a file and parametrize it from the workload-params:

(train-body.json)
```
{
    "training_index": " {{ training_index }}",
    "training_field": " {{ training_field }}",
    "dimension":  {{ dimension] }},
    "max_training_vector_count":  {{ max_training_vector_count | default(8) }},
    "search_size":  {{ search_size | default(8) }},
    "description": "My model",
    "method": {
        "name":"ivf",
        "engine":"faiss",
        "space_type": "l2",
        "parameters":{
            "nlists": {{ nlists | default(8) }},
            "encoder":{
                "name":"pq",
                "parameters":{
                    "code_size": {{ code_size | default(8) }}
                }
            }
        }
    }
}
```

I know that this works when defining indices, but is there a way for this work for arbitrary custom json file parameters?

[...]

PR #424, opened 2023-12-18
Add dataset parser for vector search

VijayanB:
### Description
Vector Search Workloads usually uses dataset of type hdf5 and bigann. Hence, added support
to parse, partition those file formats. This will be later used by VectorSearch Param source in next pr.

### Issues Resolved
part of #103 

### Testing

```
make test
tests/utils/parse_test.py::ParseParamsFunctionalTests::test_parse_float_parameter_default PASSED [ 55%]
tests/utils/parse_test.py::ParseParamsFunctionalTests::test_parse_float_parameter_from_params PASSED [ 55%]
tests/utils/parse_test.py::ParseParamsFunctionalTests::test_parse_int_parameter_default PASSED [ 56%]
tests/utils/parse_test.py::ParseParamsFunctionalTests::test_parse_int_parameter_from_params PASSED [ 56%]
tests/utils/parse_test.py::ParseParamsFunctionalTests::test_parse_string_parameter_default PASSED [ 56%]
tests/utils/parse_test.py::ParseParamsFunctionalTests::test_parse_string_parameter_from_params PASSED [ 56%]

tests/utils/dataset_test.py::DataSetTestCase::testBigANNAsAcceptableDataSetFormatWithFloatExtension PASSED [ 47%]
tests/utils/dataset_test.py::DataSetTestCase::testHDF5AsAcceptableDataSetFormat PASSED [ 47%]
tests/utils/dataset_test.py::DataSetTestCase::testUnSupportedDataSetFormat PASSED [ 47%]

================= 1200 passed, 5 skipped, 3 warnings in 15.22s =================
```
---

### DIFF
--- a/osbenchmark/exceptions.py
+++ b/osbenchmark/exceptions.py
@@ -110,3 +110,17 @@ class WorkloadConfigError(BenchmarkError):
 
 class NotFound(BenchmarkError):
     pass
+
+
+class InvalidExtensionException(BenchmarkError):
+    """
+    Thrown when invalid or unsupported file extension is passed in config
+    """
+
+
+class ConfigurationError(BenchmarkError):
+    """Exception raised for errors configuration.
+
+    Attributes:
+        message -- explanation of the error
+    """
--- a/osbenchmark/utils/dataset.py
+++ b/osbenchmark/utils/dataset.py
@@ -1,0 +1,242 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+
+import os
+import struct
+from abc import ABC, ABCMeta, abstractmethod
+from enum import Enum
+from typing import cast
+
+import h5py
+import numpy as np
+
+from osbenchmark.exceptions import InvalidExtensionException
+from osbenchmark.utils.parse import ConfigurationError
+
+
+class Context(Enum):
+    """DataSet context enum. Can be used to add additional context for how a
+    data-set should be interpreted.
+    """
+    INDEX = 1
+    QUERY = 2
+    NEIGHBORS = 3
+
+
+class DataSet(ABC):
+    """DataSet interface. Used for reading data-sets from files.
+
+    Methods:
+        read: Read a chunk of data from the data-set
+        seek: Get to position in the data-set
+        size: Gets the number of items in the data-set
+        reset: Resets internal state of data-set to beginning
+    """
+    __metaclass__ = ABCMeta
+
+    BEGINNING = 0
+
+    @abstractmethod
+    def read(self, chunk_size: int):
+        """Read vector for given chunk size
+        @param chunk_size: limits vector size to read
+        """
+
+    @abstractmethod
+    def seek(self, offset: int):
+        """
+        Move reader to given offset
+        @param offset: value to move reader pointer to
+        """
+
+    @abstractmethod
+    def size(self):
+        """
+        Returns size of dataset
+        """
+
+    @abstractmethod
+    def reset(self):
+        """
+        Resets the dataset reader
+        """
+
+
+def get_data_set(data_set_format: str, path: str, context: Context):
+    """
+    Factory method to get instance of Dataset for given format.
+    Args:
+        data_set_format: File format like hdf5, bigann
+        path: Data set file path
+        context: Dataset Context Enum
+    Returns: DataSet instance
+    """
+    if data_set_format == HDF5DataSet.FORMAT_NAME:
+        return HDF5DataSet(path, context)
+    if data_set_format == BigANNVectorDataSet.FORMAT_NAME:
+        return BigANNVectorDataSet(path)
+    raise ConfigurationError("Invalid data set format")
+
+
+class HDF5DataSet(DataSet):
+    """ Data-set format corresponding to `ANN Benchmarks
+    <https://github.com/erikbern/ann-benchmarks#data-sets>`_
+    """
+
+    FORMAT_NAME = "hdf5"
+
+    def __init__(self, dataset_path: str, context: Context):
+        file = h5py.File(dataset_path)
+        self.data = cast(h5py.Dataset, file[self.parse_context(context)])
+        self.current = self.BEGINNING
+
+    def read(self, chunk_size: int):
+        if self.current >= self.size():
+            return None
+
+        end_offset = self.current + chunk_size
+        if end_offset > self.size():
+            end_offset = self.size()
+
+        vectors = cast(np.ndarray, self.data[self.current:end_offset])
+        self.current = end_offset
+        return vectors
+
+    def seek(self, offset: int):
+
+        if offset < self.BEGINNING:
+            raise Exception("Offset must be greater than or equal to 0")
+
+        if offset >= self.size():
+            raise Exception("Offset must be less than the data set size")
+
+        self.current = offset
+
+    def size(self):
+        return self.data.len()
+
+    def reset(self):
+        self.current = self.BEGINNING
+
+    @staticmethod
+    def parse_context(context: Context) -> str:
+        if context == Context.NEIGHBORS:
+            return "neighbors"
+
+        if context == Context.INDEX:
+            return "train"
+
+        if context == Context.QUERY:
+            return "test"
+
+        raise Exception("Unsupported context")
+
+
+class BigANNVectorDataSet(DataSet):
+    """ Data-set format for vector data-sets for `Big ANN Benchmarks
+    <https://big-ann-benchmarks.com/index.html#bench-datasets>`_
+    """
+
+    DATA_SET_HEADER_LENGTH = 8
+    U8BIN_EXTENSION = "u8bin"
+    FBIN_EXTENSION = "fbin"
+    FORMAT_NAME = "bigann"
+    SUPPORTED_EXTENSION = [
+        FBIN_EXTENSION, U8BIN_EXTENSION
+    ]
+
+    BYTES_PER_U8INT = 1
+    BYTES_PER_FLOAT = 4
+
+    def __init__(self, dataset_path: str):
+        self.file = open(dataset_path, 'rb')
+        self.file.seek(BigANNVectorDataSet.BEGINNING, os.SEEK_END)
+        num_bytes = self.file.tell()
+        self.file.seek(BigANNVectorDataSet.BEGINNING)
+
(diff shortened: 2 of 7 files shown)


VijayanB:
I tested this with my next change where i actually perform vector search using new param source that accepts hdf5 and u8bin dataset.

jmazanec15 (approved):

osbenchmark/utils/dataset.py
```diff
@@ -0,0 +1,230 @@
...
+    BYTES_PER_FLOAT = 4
+
+    def __init__(self, dataset_path: str):
+        self.file = open(dataset_path, 'rb')
+        self.file.seek(BigANNVectorDataSet.BEGINNING, os.SEEK_END)
+        num_bytes = self.file.tell()
+        self.file.seek(BigANNVectorDataSet.BEGINNING)
+
+        if num_bytes < BigANNVectorDataSet.DATA_SET_HEADER_LENGTH:
+            raise Exception("File is invalid")
```
navneet1v, line 157: fix this error message to provide details why this error happened.

osbenchmark/utils/dataset.py
```diff
@@ -0,0 +1,230 @@
...
+
+        self.num_points = int.from_bytes(self.file.read(4), "little")
+        self.dimension = int.from_bytes(self.file.read(4), "little")
+        self.bytes_per_num = self._get_data_size(dataset_path)
+
+        if (num_bytes - BigANNVectorDataSet.DATA_SET_HEADER_LENGTH) != self.num_points * \
+                self.dimension * self.bytes_per_num:
+            raise Exception("File is invalid")
```
navneet1v, line 165: same as above.

osbenchmark/utils/dataset.py
```diff
@@ -0,0 +1,230 @@
...
+    def seek(self, offset: int):
+
+        if offset < self.BEGINNING:
+            raise Exception("Offset must be greater than or equal to 0")
+
+        if offset >= self.size():
+            raise Exception("Offset must be less than the data set size")
+
+        bytes_offset = BigANNVectorDataSet.DATA_SET_HEADER_LENGTH + \
+                       self.dimension * self.bytes_per_num * offset
```
navneet1v, line 192: I can see 2 operators here `+` and `\` what does the mean?

VijayanB: '\' is line break operator. I will update it to be clear

osbenchmark/utils/dataset.py
VijayanB: Ack

osbenchmark/utils/dataset.py
VijayanB: Ack

osbenchmark/utils/dataset.py
navneet1v: why in code we need a line break? is this specific to python?

navneet1v (approved):

osbenchmark/utils/dataset.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def read(self, chunk_size: int):
+        if self.current >= self.size():
+            return None
+
+        end_offset = self.current + chunk_size
+        if end_offset > self.size():
+            end_offset = self.size()
+
+        v = cast(np.ndarray, self.data[self.current:end_offset])
```
IanHoang, line 103: Nit: what does `v` represent? Is there a more descriptive term that you can be used instead?

osbenchmark/utils/dataset.py
```diff
@@ -0,0 +1,232 @@
...
+    BYTES_PER_FLOAT = 4
+
+    def __init__(self, dataset_path: str):
+        self.file = open(dataset_path, 'rb')
+        self.file.seek(BigANNVectorDataSet.BEGINNING, os.SEEK_END)
+        num_bytes = self.file.tell()
+        self.file.seek(BigANNVectorDataSet.BEGINNING)
+
+        if num_bytes < BigANNVectorDataSet.DATA_SET_HEADER_LENGTH:
+            raise Exception("Invalid file: file size cannot be lesser than {} bytes".format(
```
IanHoang, line 157: Nit: `cannot be lesser than` -> `cannot be less than`

osbenchmark/utils/dataset.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def read(self, chunk_size: int):
+        if self.current >= self.size():
+            return None
+
+        end_offset = self.current + chunk_size
+        if end_offset > self.size():
+            end_offset = self.size()
+
+        v = np.asarray([self._read_vector() for _ in
```
IanHoang, line 180: I see, `v` represents `vector`. Nit: update `v` --> `vector` to be more descriptive and clear to readers

osbenchmark/utils/dataset.py
IanHoang: Update: `v` represents `vector`. Can we update the variable to be `vector`?

osbenchmark/utils/dataset.py
```diff
@@ -0,0 +1,232 @@
...
+    @staticmethod
+    def _get_data_size(file_name):
+        ext = file_name.split('.')[-1]
+        if ext == BigANNVectorDataSet.U8BIN_EXTENSION:
+            return BigANNVectorDataSet.BYTES_PER_U8INT
+
+        if ext == BigANNVectorDataSet.FBIN_EXTENSION:
+            return BigANNVectorDataSet.BYTES_PER_FLOAT
+
+        raise Exception("Unknown extension")
```
IanHoang, line 221: Ideally, we should avoid having generic exceptions like this and use have a clear exception for this in `exceptions.py`. 

Something like 
```
raise InvalidExtensionException("Invalid extension. File extensions should be u8bin or fbin")
```

Nit: Additionally, do we see any possibilities of adding more extensions in the future? If so, it'd be good to place these extensions in an enum and have it dynamically get called when creating the exception. But it's not necessary at this moment if we're just sticking with two extensions.

osbenchmark/utils/dataset.py
```diff
@@ -0,0 +1,232 @@
...
+    @staticmethod
+    def _value_reader(file_name):
+        ext = file_name.split('.')[-1]
+        if ext == BigANNVectorDataSet.U8BIN_EXTENSION:
+            return lambda file: float(int.from_bytes(file.read(BigANNVectorDataSet.BYTES_PER_U8INT), "little"))
+
+        if ext == BigANNVectorDataSet.FBIN_EXTENSION:
+            return lambda file: struct.unpack('<f', file.read(BigANNVectorDataSet.BYTES_PER_FLOAT))
+
+        raise Exception("Unknown extension")
```
IanHoang, line 232: See comment here https://github.com/opensearch-project/opensearch-benchmark/pull/424/files#r1432990229

osbenchmark/utils/parse.py
```diff
@@ -0,0 +1,59 @@
...
+            "Value cannot be None for param {}".format(key)
+        )
+
+    if isinstance(params[key], float):
+        return params[key]
+
+    raise ConfigurationError("Value must be a float for param {}".format(key))
+
+
+class ConfigurationError(Exception):
```
IanHoang, line 50: Let's move this to [osbenchmark/exceptions.py](https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/exceptions.py)

IanHoang, line 50: Let's move this class to [osbenchmark/exceptions.py](https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/exceptions.py)

osbenchmark/utils/dataset.py
VijayanB: Ack

osbenchmark/utils/dataset.py
VijayanB: Ack

osbenchmark/utils/dataset.py
VijayanB: Ack

IanHoang (changes requested):
Left some comments

osbenchmark/utils/dataset.py
VijayanB: Ack

osbenchmark/utils/dataset.py
VijayanB: Ack

osbenchmark/utils/parse.py
```diff
@@ -0,0 +1,59 @@
...
+            "Value cannot be None for param {}".format(key)
+        )
+
+    if isinstance(params[key], str):
+        return params[key]
+
+    raise ConfigurationError("Value must be a string for param {}".format(key))
+
+
+def parse_int_parameter(key: str, params: dict, default: int = None) -> int:
```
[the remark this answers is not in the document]
VijayanB: What happens if we pass integer value to string parameter?  The above looks complicated for simple function. I don't see any advantage. However, if you have strong opinion, i can refactor.

osbenchmark/utils/parse.py
VijayanB: Ack

osbenchmark/utils/parse.py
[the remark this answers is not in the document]
IanHoang: > What happens if we pass integer value to string parameter? 
Are you referring to passing an integer value to the string parameter `key`? The code above is based off of the three functions (`parse_str_parameter`, `parse_float_parameter`, and `parse_int_parameter`) and how they were used in the unittests in `parse_test.py`. In `parse_test.py`, it doesn't seem like there's a scenario where an integer value is passed to `key` since only values such as "int-value", "string-value", "default-value", "float-value" are provided. I was under the assumption that OSB must first identify if the parameter is one of the three valid datatypes and then call the function matching that data type. Let me know if I'm misunderstanding this. 

That being said, I don't have a strong preference and we can keep these functions as is 👍🏻

osbenchmark/exceptions.py
```diff
@@ -110,3 +110,17 @@ class WorkloadConfigError(BenchmarkError):
 
 class NotFound(BenchmarkError):
     pass
+
+
+class InvalidExtensionException(BenchmarkError):
```
IanHoang, line 115: Thanks for doing this!

IanHoang (approved):
LGTM!

osbenchmark/utils/dataset.py
VijayanB: python usually throws pylint error if we have more than x characters per line, i believe to avoid that we added "\" to tell python interpreter that following line is part of current line.

PR #430, opened 2023-12-28
Remove references to default admin creds

derek-ho:
### Description
With 2.12.0 release, the security plugin is making a change that requires an initial admin password to be set via an environment variable. This PR makes this change in CI/code/documentation. 

### Issues Resolved

### Testing

---

### DIFF
--- a/osbenchmark/resources/docker-compose.yml.j2
+++ b/osbenchmark/resources/docker-compose.yml.j2
@@ -39,7 +39,7 @@ services:
     networks:
       - opensearch-net
     healthcheck:
-          test: curl -f http://localhost:{{http_port}} -u admin:admin --insecure
+          test: curl -f http://localhost:{{http_port}} -u admin:myStrongPassword123! --insecure
           interval: 5s
           timeout: 2s
           retries: 10
--- a/osbenchmark/resources/docker-compose.yml.j2
+++ b/osbenchmark/resources/docker-compose.yml.j2
@@ -40,7 +40,12 @@ services:
     networks:
       - opensearch-net
     healthcheck:
-          test: curl -f http://localhost:{{http_port}} -u admin:myStrongPassword123! --insecure
+          {%- set version_numbers = os_version.split('.') | map('int') %}
+          {%- if version_numbers[0] > 2 or (version_numbers[0] == 2 and version_numbers[1] >= 12) %}
+            test: curl -f http://localhost:{{http_port}} -u admin:myStrongPassword123! --insecure
+          {%- else %}
+            test: curl -f http://localhost:{{http_port}} -u admin:admin --insecure
+          {%- endif %}
           interval: 5s
           timeout: 2s
           retries: 10
--- a/osbenchmark/resources/docker-compose.yml.j2
+++ b/osbenchmark/resources/docker-compose.yml.j2
@@ -18,6 +18,7 @@ services:
       - DISABLE_INSTALL_DEMO_CONFIG=true
       - bootstrap.memory_lock=true
       - "OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m"
+      - "OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123!" # starting with 2.12.0 a strong initial admin password must be set
     ulimits:
       memlock:
         soft: -1
--- a/samples/ccr/start.sh
+++ b/samples/ccr/start.sh
@@ -25,7 +25,7 @@ printf "Waiting for clusters to get ready "
 ALL_CLUSTERS_READY=false
 
 while ! $ALL_CLUSTERS_READY; do
-    (curl -ks -u admin:admin https://localhost:9200 -o /dev/null && curl -ks -u admin:admin https://localhost:9201 -o /dev/null && ALL_CLUSTERS_READY=true) || (printf "." && sleep 5)
+    (curl -ks -u admin:myStrongPassword123! https://localhost:9200 -o /dev/null && curl -ks -u admin:admin https://localhost:9201 -o /dev/null && ALL_CLUSTERS_READY=true) || (printf "." && sleep 5)
 done
 
 echo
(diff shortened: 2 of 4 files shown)


derek-ho:
@IanHoang @gkamat can you folks take a look/ I am expecting this to fail security integration tests, but would be needed to fix the CI after 2.12.0 docker image gets released. I also have some questions if I made the changes correctly, since I didn't fully understand how this repo is using some files. I am assuming some docker files with reference to 1.x line don't need changes/have no plans to be changed in the near future.

derek-ho:
https://github.com/opensearch-project/security/issues/3624

osbenchmark/resources/docker-compose.yml.j2
```diff
@@ -39,7 +39,7 @@ services:
     networks:
       - opensearch-net
     healthcheck:
-          test: curl -f http://localhost:{{http_port}} -u admin:admin --insecure
+          test: curl -f http://localhost:{{http_port}} -u admin:myStrongPassword123! --insecure
```
derek-ho, line 42: This command would fail if os_version < 2.12.0. Not sure how this file is used/how we can conditionally switch this password. Can maintainers help out here?

samples/ccr/start.sh
```diff
@@ -25,7 +25,7 @@ printf "Waiting for clusters to get ready "
 ALL_CLUSTERS_READY=false
 
 while ! $ALL_CLUSTERS_READY; do
-    (curl -ks -u admin:admin https://localhost:9200 -o /dev/null && curl -ks -u admin:admin https://localhost:9201 -o /dev/null && ALL_CLUSTERS_READY=true) || (printf "." && sleep 5)
+    (curl -ks -u admin:myStrongPassword123! https://localhost:9200 -o /dev/null && curl -ks -u admin:admin https://localhost:9201 -o /dev/null && ALL_CLUSTERS_READY=true) || (printf "." && sleep 5)
```
derek-ho, line 28: Not sure where the second cluster is coming from, I think the first cluster is coming from `docker-compose-metricstore`, which is on the latest release. This change will fail CI until 2.12.0 docker image is released, at which point it would fix the CI.

derek-ho: Can maintainers help out here in understanding if the second cluster needs similar changes/where those that cluster's password should be set?

IanHoang: I do not believe this script is being used anymore. @gkamat can you confirm?

IanHoang:
@derek-ho Can you update the following tests so that the assertions pass?
```
tests/builder/provisioner_test.py::DockerProvisionerTests::test_provisioning_with_defaults FAILED [ 20%]
tests/builder/provisioner_test.py::DockerProvisionerTests::test_provisioning_with_variables FAILED [ 20%]
tests/builder/installers/docker_installer_test.py::DockerProvisionerTests::test_provisioning_with_defaults FAILED [ 29%]
tests/builder/installers/docker_installer_test.py::DockerProvisionerTests::test_provisioning_with_variables FAILED [ 29%]
```

osbenchmark/resources/docker-compose.yml.j2
IanHoang: @derek-ho Off the top of my head, there's a couple ways that we could get around this. Others might have better ideas
1. You can add an additioonal method that checks the `os_version` variable that's being passed through and if it's < 2.12.0, use `admin`. You'll need to make that additional check in [this section of provisioner.py](https://github.com/opensearch-project/opensearch-benchmark/blob/b861175205a54725c3d7fbbeed02d9be834aa986/osbenchmark/builder/provisioner.py#L503-L505) and  [this section of docker_installer.py ](https://github.com/opensearch-project/opensearch-benchmark/blob/b861175205a54725c3d7fbbeed02d9be834aa986/osbenchmark/builder/installers/docker_installer.py#L109-L112)
2. I believe you could also leverage a library such as`yq` https://stackoverflow.com/questions/76526508/replace-a-block-of-yaml but that would require upkeep with this library and adding it in `setup.py`

IanHoang (changes requested):
Left some comments

samples/ccr/start.sh
derek-ho: @IanHoang I will remove this file then, if its not being used. @gkamat can you also confirm here?

derek-ho:
I tried to add some logic around versioning in the PR. I will put it into draft for now, since it will be easier to take care of once 2.12.0 is released and we can really test the difference between the logic spitting out different health checks based on the version. In any case, the current CI is in a good state and will be so until after the release. @IanHoang let me know if that makes sense for you.

derek-ho:
@IanHoang I just went through this PR once more. Are the dockerfiles that I touched only used to assert that they are read correctly? If that is the case then this PR may not be necessary. Can you point me to the versions of opensearch you are testing against? Does that matrix have 2.12 added to it after the release? I couldn't find it

IanHoang:
> @IanHoang I just went through this PR once more. Are the dockerfiles that I touched only used to assert that they are read correctly? If that is the case then this PR may not be necessary. Can you point me to the versions of opensearch you are testing against? Does that matrix have 2.12 added to it after the release? I couldn't find it

The dockerfiles that you touched are used in the integration tests and when users want to specifically run OSB against a Docker version of OpenSearch. 

These are the distributions that OpenSearch Benchmark tests against:
```
DISTRIBUTIONS = ["1.3.9", "2.5.0"]
```

To test if your changes in `docker-compose.yml.j2` really work, you can perform the procedure:
1. Alter the integration tests in `it/__init__.py` and change the DISTRIBUTIONS variable to point to a list containing `2.12.0`.
2. In your forked OSB repository, go to Github actions and run the integration tests

In the logs, you can see if any errors occur related to `distribution_test.py`, specifically `test_docker_distribution()`

IanHoang:
@derek-ho are you still working on this? If not, can we close this?

PR #445, opened 2024-01-23
Added support for opensearch-py client connections urllib3httpconnection, requestshttpconnection

### LINKED ISSUE #437: Support Multiple opensearch-py Client Connections
The objective is to extract opensearch-py client performance metrics in opensearch-benchmark with various connections (urllib3httpconnection, requestshttpconnection, aiohttpconnection). The current code exclusively uses aiohttpconnection. Historically, opensearch-benchmark supported urllib3httpconnection, and at a different time, aiohttpconnection. Aim to introduce the flexibility to choose between these connections, allowing users to gain insights into their respective performances.

Enable opensearch-benchmark to work with urllib3httpconnection and requestshttpconnection for opensearch-py client. Provide an option to switch the client connection using an argument in the execute-test command.

saimedhi:
### Description
Added support for opensearch-py client connections urllib3httpconnection, requestshttpconnection.
 

### Issues Resolved
#437 

### Follow up Issues 
Document how to use urllib3httpconnection, requestshttpconnection client connections.

### Usage examples:

- below connection_class in --client-options is case insensitive. 
opensearch-benchmark execute-test --distribution-version=2.11.1 --workload=geonames --test-mode --kill-running-processes --client-options "connection_class:urllib3"

- For urllib3httpconnection: both options (urllib3, urllib3httpconnection) will work. Case insensitive
- For requestshttpconnection: both options (requests, requestshttpconnection) will work. Case insensitive

---


osbenchmark/client.py
```diff
@@ -282,23 +281,35 @@ def parse_aws_log_in_params(self):
...
 
+        if "connection_class" in self.client_options:
+            connection = self.client_options.pop("connection_class", None)
+            if connection.lower() in {"requestshttpconnection", "requests"}:
+                connection_class = osbenchmark.sync_connection.RequestsHttpConnection
+            else:
+                connection_class = osbenchmark.sync_connection.Urllib3HttpConnection
+        else:
+            connection_class = osbenchmark.sync_connection.Urllib3HttpConnection
+
```
dblock, line 295: This seems unnecessarily complicated, and should raise an error if the user specified an invalid class.

```python
connection_class = self.client_options.get("connection_class", "urllib3")
if connection_class == "requests":
    connection_class = osbenchmark.sync_connection.RequestsHttpConnection
elif connection_class == "..."
   ...
else:
    throw ...
```


```

File: osbenchmark/worker_coordinator/worker_coordinator.py [hunk 2]
```diff
@@ -1449,7 +1449,17 @@ async def run(self):
         def os_clients(all_hosts, all_client_options):
             opensearch = {}
             for cluster_name, cluster_hosts in all_hosts.items():
-                opensearch[cluster_name] = client.OsClientFactory(cluster_hosts, all_client_options[cluster_name]).create_async()
+                if "connection_class" in all_client_options["default"]:
```
dblock, line 1452: Same as above, collapse the nested `if/else`.

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1493,7 +1503,10 @@ def os_clients(all_hosts, all_client_options):
             shutdown_asyncgens_end = time.perf_counter()
             self.logger.info("Total time to shutdown asyncgens: %f seconds.", (shutdown_asyncgens_end - run_end))
             for s in opensearch.values():
-                await s.transport.close()
+                if asyncio.iscoroutinefunction(s.transport.close):
+                    await s.transport.close()
+                else:
+                    s.transport.close()
```
dblock, line 1509: Ideally there should be a sync and an async worker coordinator that have a `close` method, or a `transport` class that wraps the async version so that you can change this code to just `s.transport.close()`.

osbenchmark/client.py
saimedhi: > This seems unnecessarily complicated, and should raise an error if the user specified an invalid class.
> 
> ```python
> connection_class = self.client_options.get("connection_class", "urllib3")
> if connection_class == "requests":
>     connection_class = osbenchmark.sync_connection.RequestsHttpConnection
> elif connection_class == "..."
>    ...
> else:
>     throw ...
> ```

- Error handling is done [here](https://github.com/opensearch-project/opensearch-benchmark/pull/445/files#diff-c89a0ed9511e7ad75cf883fcc7c10bc472902bcb8d3b9f1ba86f2eb50a560636R1460) 
- I will simplify the code

gkamat: Agree with @dblock   Since Python does not have a `switch` or `case` statement, at least pre-3.10, you can use a structured if-elif block or perhaps a map.  Also, a `default` arm with error handling is good practice, even if there is error handling elsewhere, since a reader of this section might not be aware.

osbenchmark/client.py
```diff
@@ -282,23 +281,35 @@ def parse_aws_log_in_params(self):
...
         if "amazon_aws_log_in" not in self.client_options:
-            return opensearchpy.OpenSearch(hosts=self.hosts, ssl_context=self.ssl_context, **self.client_options)
+            return self.BenchmarkOpenSearch(hosts=self.hosts, ssl_context=self.ssl_context,
```
gkamat, line 297: Would suggest renaming this to `BenchmarkSyncOpenSearch` to match the other class.  Likewise with `create()` vis-a-vis `create_async()`.  Please add comments prior to each indicating the presence of the other.

osbenchmark/worker_coordinator/worker_coordinator.py
gkamat: This could be refactored in a future check-in.

osbenchmark/sync_connection.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def perform_request(self, method, url, params=None, body=None, timeout=None, allow_redirects=True, ignore=(), headers=None):
+        try:
+            # pylint: disable=import-outside-toplevel
+            from osbenchmark.client import RequestContextHolder
+            request_context_holder = RequestContextHolder()
+            request_context_holder.on_request_start()
+            status, headers, raw_data = super().perform_request(method=method, url=url, params=params, body=body, timeout=timeout,
+                                                                allow_redirects=allow_redirects, ignore=ignore, headers=headers)
+            request_context_holder.on_request_end()
```
gkamat, line 45: There is no error handling here.  If the request fails, the `on_request_end` callback will not be executed, leading to incorrect metrics in the command dispatcher, or possibly an error there.

osbenchmark/sync_connection.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def perform_request(self, method, url, params=None, body=None, timeout=None, ignore=(), headers=None):
+        try:
+            # pylint: disable=import-outside-toplevel
+            from osbenchmark.client import RequestContextHolder
+            request_context_holder = RequestContextHolder()
+            request_context_holder.on_request_start()
+            status, headers, raw_data = super().perform_request(method=method, url=url, params=params, body=body,
+                                                                timeout=timeout, ignore=ignore, headers=headers)
+            request_context_holder.on_request_end()
```
gkamat, line 86: Same comment as above.

osbenchmark/worker_coordinator/runner.py
```diff
@@ -490,13 +490,20 @@ async def __call__(self, opensearch, params):
...
+                response = opensearch.bulk(doc_type=params.get("type"), params=bulk_params, **api_kwargs)
         else:
-            response = await opensearch.bulk(doc_type=params.get("type"), params=bulk_params, **api_kwargs)
-
+            if with_action_metadata:
+                api_kwargs.pop("index", None)
+                # only half of the lines are documents
+                response = await opensearch.bulk(params=bulk_params, **api_kwargs)
+            else:
+                response = await opensearch.bulk(doc_type=params.get("type"), params=bulk_params, **api_kwargs)
```
gkamat, line 506: There is a lot of code duplication in this file that is likely not necessary.  It should be possible to indirect the call to the client API via a routine that calls the desired sync or async function by switching on the predicate.  The rest of the code can stay the same.

gkamat (changes requested):

osbenchmark/worker_coordinator/worker_coordinator.py
gkamat, line 1452: Consider changing the user flag to `backend-client` or `client`, since this flag will continue to be used for future clients like a Rust client, C++ client, etc.

osbenchmark/worker_coordinator/runner.py
saimedhi: @gkamat, can we do it as below? 

- Adding await conditionally on each line where opensearch call is made. 

 
```
if with_action_metadata:
  api_kwargs.pop("index", None)
  response = opensearch.bulk(params=bulk_params, **api_kwargs) if "BenchmarkOpenSearch" in str(opensearch) else await opensearch.bulk(params=bulk_params, **api_kwargs)
else:
    response = opensearch.bulk(doc_type=params.get("type"), params=bulk_params, **api_kwargs) if "BenchmarkOpenSearch" in str(opensearch) else await opensearch.bulk(doc_type=params.get("type"), params=bulk_params, **api_kwargs)
```

osbenchmark/client.py
gkamat: Actually, you might. consider renaming both classes to be more descriptive, something like _PythonSyncClient_ and _PythonAsyncClient_, so future clients can be named accordingly.

osbenchmark/worker_coordinator/runner.py
gkamat: Yes, that is the ternary operator referred to above.  It will be more compact to encapsulate the predicate within a function.  Also, it is probably cleaner to define the classes at the top level in client.py, import them and test directly, rather than using `str()`:
```
opensearch.__class__ == PythonSyncClient
```
or something like that.

osbenchmark/sync_connection.py
gkamat: Note: this may be the wrong location to insert these calls, if they are intended to measure service time.  They should be at the lowest level possible just prior to the HTTP socket send call.  The aiohttp library has callbacks, but the others might not.  In that case, the calls will need to be somewhere within `opensearchpy`, not here.

saimedhi: Hey @dblock, any ideas on how we can get the exact timing? Should we make changes in opensearch-py?

saimedhi: @dblock

dblock: For the issue that @gkamat is bringing up, the calling code and the implementation should become a `with` block / `__enter__` and `__exit__` so that `on_request_end` is always called, even on error.

On the second issue of lowest level I think we _are_ trying to measure "client overhead" that begins "around" the client. I also do think it would be good to get lower level HTTPs library metrics for pure transport. Finally, the entire client will become swappable, so keeping the code here makes sense to me.

@gkamat WDYT?

gkamat: @dblock, yes, either a context manager or exception handling will be an appropriate mechanism to handle the issue.

Regarding the second point, I believe those calls were added by @saimedhi to measure _service time_, which reflects the time the request is placed on the wire and received from the network (as accurately as is feasible).  The `aiohttp` client provides trace event callbacks that make this possible.  

You are correct about the client processing time in that it measures latency around the client and that location above would be the correct entry point to measure this metric.  But I believe it is the other metric that is being captured here and the difference between the two would be pure client overhead.

osbenchmark/worker_coordinator/worker_coordinator.py
saimedhi: I will rename the user flag to backend-client

saimedhi:
Closing this PR. I will resubmit it once opensearch-py has [service time metrics](https://github.com/opensearch-project/opensearch-py/pull/716).

PR #452, opened 2024-01-30
Move data set loading outside constructor

VijayanB:
### Description
Param sources will be instatiated before corpora loading. Hence, set data set parameters to initial value during initialization and load data set during partition stage.

### Testing

---

### DIFF
--- a/osbenchmark/utils/dataset.py
+++ b/osbenchmark/utils/dataset.py
@@ -89,11 +89,18 @@ class HDF5DataSet(DataSet):
     FORMAT_NAME = "hdf5"
 
     def __init__(self, dataset_path: str, context: Context):
-        file = h5py.File(dataset_path)
-        self.data = cast(h5py.Dataset, file[self.parse_context(context)])
+        self.dataset_path = dataset_path
+        self.context = self.parse_context(context)
         self.current = self.BEGINNING
+        self.data = None
+
+    def _load(self):
+        if self.data is None:
+            file = h5py.File(self.dataset_path)
+            self.data = cast(h5py.Dataset, file[self.context])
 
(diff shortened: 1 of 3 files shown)


VijayanB:
@IanHoang @gkamat Any idea why is this failing? s this transient?
https://github.com/opensearch-project/opensearch-benchmark/actions/runs/7717063834/job/21035304056?pr=452
Thanks

osbenchmark/workload/params.py
```diff
@@ -872,6 +867,14 @@ def partition(self, partition_index, total_partitions):
         Returns:
             The parameter source for this particular partition
         """
+        if self.data_set is None:
```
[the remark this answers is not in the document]
VijayanB: As discussed offline, this will not have impact over multithreading since partition is called by original instance and by decoupling dataset loading from constructor, get_data_set is idempotent when passed same parameters. This is also consistent with bulk para source too.

jmazanec15 (approved):

osbenchmark/utils/dataset.py
```diff
@@ -106,7 +113,7 @@ def read(self, chunk_size: int):
         return vectors
 
     def seek(self, offset: int):
-
+        # load file first before seek
         if offset < self.BEGINNING:
```
IanHoang, line 117: Should there be a `self._load()` here ?

IanHoang (changes requested):
Left one comment

osbenchmark/utils/dataset.py
VijayanB: Updated. We don't have to load for seek. I removed that comment.

IanHoang (approved):

Issue #443, opened 2024-01-22
[RFC] Adding capabilities to introduce randomness in workload queries


sgup432:
**Is your feature request related to a problem? Please describe.**

I had added a proposal here in OSB workloads(https://github.com/opensearch-project/opensearch-benchmark-workloads/issues/152), shifting the conversation here instead as per suggestions.

As of now, opensearch workload mostly deals with static queries ie the values are fixed. For example in nyc_taxis(search recommended workload), we have a set of static queries which we use to perform benchmark. But this doesn't help in simulating real world scenarios where data distribution and access patterns are not predictable. Plus there is not way to use nyc_taxis to run test with caches(like request cache) turned on.

But considering we have features like Tiered caching coming up(https://github.com/opensearch-project/OpenSearch/issues/10024), we need capabilities to generate some workload with some dynamic queries(randomization in values) so that we can test with cache turned on as well. We wrote some custom logic on top of nyc_taxis to create some randomization.

**Describe the solution you'd like**

I recommend to have a parameter via which we can introduce some randomization in nyc_taxis(for example) query values which would be useful:

To perform benchmark with caches turned on. Via this parameter we can generate desired cache hits/misses.
To have some kind of randomization in workload so that it not very predictable.
We can have a parameter via which users can generate X number of repeated queries and rest unique queries. Where X>=0.

This parameter can be enabled on a demand basis if needed. I believe it might also help other workloads as well and not just nyc_taxis.

**Describe alternatives you've considered**

**Additional context**

peteralfonsi:
I have a working version of this for nyc_taxis only at https://github.com/opensearch-project/opensearch-benchmark-workloads/pull/154. We define new param-source functions in workload.py, which pull randomized values from a common function (`get_values`). Then we define operations using these param-sources in operations/default.json and add them to the schedule in test_procedures/default.json. We also define "value generator" functions to provide reasonable values depending on what workload/field is being searched. 

But to do this generically for all workloads, it would be messy, since we would have to move all the actual query definitions to workload.py and change operations/default.json and test_procedures/default.json for each workload. 

[...]

PR #455, opened 2024-02-01
Adds controllable randomization to range queries in workloads

peteralfonsi:
### Description
Allows the user to randomize the values in range queries in a controllable way. This allows reasonable tests of cache performance, where we can't use the same query repeatedly. 

To use this, in your workload's workload.py file, you must register a "standard value source" function in the register() function. This just means a function that takes no arguments and returns a dict of range values, with keys "gte", "lte", and optionally "format" (for dates). Usually you would want some randomness in this function, so that different values are produced every time you run it. During the workload, on each iteration, values from an output from this function get put into the range section of a query, and overwrite the default values. 

This registration is done on an operation and field level. For example, in nyc_taxis, say you wanted to randomize the field "trip_distance" in the operation "distance_amount_agg", to produce random values between 1 and 15. Your workload.py might look like: 

```
def trip_distance_std_value_source():
    gte = random.randrange(1, 7)
    lte = random.randrange(gte, 15)
    return {"gte":gte, "lte":lte}

def register(registry):
    registry.register_runner("delete-snapshot", delete_snapshot, async_runner=True)
    registry.register_standard_value_source("distance_amount_agg", "trip_distance", trip_distance_std_value_source)

```

Then, to enable randomization when you run the workload, you supply the flag `--randomization-enabled`. To control the fraction of queries which may be repeated (for example, to mimic a cache hit), you supply the flag `--randomization-rf 0.3`, which would mean the "repeat frequency" for this workload is 0.3 = 30%. If `--randomization-enabled` isn't passed, there is no change from the user's perspective and you don't have to register standard value sources. 

This works by adding a new WorkloadProcessor called QueryRandomizerWorkloadProcessor. If randomization is enabled, its on_after_load_workload() function will walk through the default workload schedule (the one specified in test_procedures/default.json). For all search operations, it will see if a standard value source was registered for that operation and field type. Then it gives the operation a new ParamSource. 

This ParamSource is a lambda supplied by the processor. When it's run, it finds the range section of the query, and replaces its gte/lte (and format, if present) values with the ones provided by the standard value source for that operation and field. 

To allow a controllable fraction of "cache hits", we use the rf parameter from above. 

If rf = 0.3, then 30% of the time, we want a repeated value. To achieve this, we generate and save some number N of standard values for each op/field pair at the beginning of the workload, using the same standard value source function. (Default = 5000, this is configurable via `--randomization-n`). We pick one of these saved values to return from the lambda. (We draw the index of this value from the Zipf distribution on N items, which empirically matches the distribution for online cache requests, and has some other advantages too). As the workload goes on, indices will be drawn multiple times, so the query values will repeat. 

The other 70% of the time, we generate a totally new value from the standard value source function, which will not have been seen before (a "cache miss"). 

### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-benchmark/issues/443

### Testing

Added UTs testing various components of the workflow. 

Also tested manually with several different workloads while watching the parameters with print statements, to make sure they were changing correctly and the fraction of ones being drawn randomly was correct. Example printout, where zipf-index = None means a "new" value, and other zipf-index values means that index was drawn from the saved values: 

```
Running range                                                                  [  7% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 2, 'lt': 5}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [  7% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 1, 'lt': 3}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [  9% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 2, 'lt': 5}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [  9% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 3, 'lt': 3}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [  9% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 3, 'lt': 6}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [  9% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 3, 'lt': 4}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 12% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 1, 'lt': 4}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 12% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 1, 'lt': 4}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 12% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 1, 'lt': 6}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 14% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 2, 'lt': 6}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': 2}
Running range                                                                  [ 14% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 1, 'lt': 2}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 14% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 3, 'lt': 4}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 17% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 2, 'lt': 2}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': 16}
Running range                                                                  [ 17% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 3, 'lt': 4}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 17% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 1, 'lt': 1}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': 92}
Running range                                                                  [ 17% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 1, 'lt': 1}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 19% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 3, 'lt': 4}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': 0}
Running range                                                                  [ 19% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 1, 'lt': 1}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 19% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 1, 'lt': 1}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 21% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 1, 'lt': 1}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 21% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 3, 'lt': 5}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 21% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 3, 'lt': 4}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': 1090}
Running range                                                                  [ 21% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 1, 'lt': 3}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 24% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 3, 'lt': 5}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 24% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 1, 'lt': 4}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 24% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 2, 'lt': 2}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': 9}
Running range                                                                  [ 26% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 2, 'lt': 2}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 26% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 2, 'lt': 6}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': 4}
Running range                                                                  [ 26% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 2, 'lt': 3}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 26% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 1, 'lt': 4}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 29% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 1, 'lt': 3}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': None}
Running range                                                                  [ 29% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 3, 'lt': 4}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': 0}
Running range                                                                  [ 29% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 2, 'lt': 6}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': 1084}
Running range                                                                  [ 31% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 2, 'lt': 2}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': 669}
Running range                                                                  [ 31% done]
Modified params = {'name': 'range', 'operation-type': 'search', 'body': {'query': {'range': {'total_amount': {'gte': 2, 'lt': 3}}}}, 'include-in-results_publishing': True, 'index': 'nyc_taxis', 'zipf-index': 204}
```

Finally, re-ran a cache-related workload that we had run before with a previous, non-generic way of randomizing queries, and made sure the number of hits, misses, evictions, etc matched what we observed then, to within some tolerance to allow for randomization. This was measured with the node stats API. I was confident about this previous non-generic way working, because we'd been using it for a while to benchmark the new tiered caching feature. (The test was on a tiered caching setup, which spills values from the heap tier to a disk tier when the heap tier fills).

Number of new hits/misses/etc for heap and disk tiers in the earlier, non-generic test (we had similar randomization logic living in the workload itself): 
* heap tier
    * evictions = 214754 
    * hits = 63374 
    * misses = 266226
    * entries = 5906 
* disk tier
   * evictions = 0
   * hits = 45566
   * misses = 220660

Number of new hits/misses/etc for the new test, with the setup in this PR: 
* heap tier
    * evictions = 215682
    * hits = 63432
    * misses = 266168
    * entries = 5926
* disk tier 
    * evictions = 0
    * hits = 44560
    * misses = 221608

which we can see match quite well. 
The workload here was [cacheable_nyc_taxis in this branch](https://github.com/peteralfonsi/opensearch-benchmark-workloads/tree/simple-randomization/cacheable_nyc_taxis) and the command to run it was 
```opensearch-benchmark execute-test --pipeline=benchmark-only --workload-path=/home/ec2-user/osb/opensearch-benchmark-workloads/cacheable_nyc_taxis --workload-params='{"requests_cache_enabled":"true"}' --target-host=http://localhost:9200 --randomization-enabled --randomization-rf 0.3 --randomization-n 5000```

---

### DIFF
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -893,7 +893,7 @@ def dispatch_sub_command(arg_parser, args, cfg):
             cfg.add(config.Scope.applicationOverride, "workload", "latency.percentiles", args.latency_percentiles)
             cfg.add(config.Scope.applicationOverride, "workload", "throughput.percentiles", args.throughput_percentiles)
             cfg.add(config.Scope.applicationOverride, "workload", "randomization.enabled", args.randomization_enabled)
-            cfg.add(config.Scope.applicationOverride, "workload", "randomization.rf", args.randomization_rf)
+            cfg.add(config.Scope.applicationOverride, "workload", "randomization.repeat_frequency", args.randomization_repeat_frequency)
             cfg.add(config.Scope.applicationOverride, "workload", "randomization.n", args.randomization_n)
             configure_workload_params(arg_parser, args, cfg)
             configure_connection_params(arg_parser, args, cfg)
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -573,7 +573,7 @@ def create_arg_parser():
         action="store_true")
     test_execution_parser.add_argument(
         "--randomization-repeat-frequency",
-        "--randomization-rf",
+        "-rf",
         help=f"The repeat_frequency for query randomization. Ignored if randomization is off"
              f"(default: {workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_RF}).",
         default=workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_RF)
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -573,11 +573,13 @@ def create_arg_parser():
         action="store_true")
     test_execution_parser.add_argument(
         "--randomization-rf",
-        help="The repeat_frequency for query randomization. Ignored if randomization is off (default: {workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_RF}).",
+        help=f"The repeat_frequency for query randomization. Ignored if randomization is off"
+             f"(default: {workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_RF}).",
         default=workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_RF)
     test_execution_parser.add_argument(
         "--randomization-n",
-        help="The number of standard values to generate for each field for query randomization. Ignored if randomization is off (default: {workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_N}).",
+        help=f"The number of standard values to generate for each field for query randomization."
+             f"Ignored if randomization is off (default: {workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_N}).",
         default=workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_N)
 
     ###############################################################################
@@ -887,14 +889,11 @@ def dispatch_sub_command(arg_parser, args, cfg):
                 "load_worker_coordinator_hosts",
                 opts.csv_to_list(args.load_worker_coordinator_hosts))
             cfg.add(config.Scope.applicationOverride, "workload", "test.mode.enabled", args.test_mode)
-<<<<<<< HEAD
             cfg.add(config.Scope.applicationOverride, "workload", "latency.percentiles", args.latency_percentiles)
             cfg.add(config.Scope.applicationOverride, "workload", "throughput.percentiles", args.throughput_percentiles)
-=======
             cfg.add(config.Scope.applicationOverride, "workload", "randomization.enabled", args.randomization_enabled)
             cfg.add(config.Scope.applicationOverride, "workload", "randomization.rf", args.randomization_rf)
             cfg.add(config.Scope.applicationOverride, "workload", "randomization.n", args.randomization_n)
->>>>>>> randomized
             configure_workload_params(arg_parser, args, cfg)
             configure_connection_params(arg_parser, args, cfg)
             configure_telemetry_params(args, cfg)
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -566,6 +566,19 @@ def create_arg_parser():
              f"(default: {metrics.GlobalStatsCalculator.DEFAULT_THROUGHPUT_PERCENTILES}).",
         default=metrics.GlobalStatsCalculator.DEFAULT_THROUGHPUT_PERCENTILES
     )
+    test_execution_parser.add_argument(
+        "--randomization-enabled",
+        help="Runs the given workload with query randomization enabled (default: false).",
+        default=False,
+        action="store_true")
+    test_execution_parser.add_argument(
+        "--randomization-rf",
+        help="The repeat_frequency for query randomization. Ignored if randomization is off (default: {workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_RF}).",
+        default=workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_RF)
+    test_execution_parser.add_argument(
+        "--randomization-n",
+        help="The number of standard values to generate for each field for query randomization. Ignored if randomization is off (default: {workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_N}).",
+        default=workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_N)
 
     ###############################################################################
     #
@@ -874,8 +887,14 @@ def dispatch_sub_command(arg_parser, args, cfg):
                 "load_worker_coordinator_hosts",
                 opts.csv_to_list(args.load_worker_coordinator_hosts))
             cfg.add(config.Scope.applicationOverride, "workload", "test.mode.enabled", args.test_mode)
+<<<<<<< HEAD
             cfg.add(config.Scope.applicationOverride, "workload", "latency.percentiles", args.latency_percentiles)
             cfg.add(config.Scope.applicationOverride, "workload", "throughput.percentiles", args.throughput_percentiles)
+=======
+            cfg.add(config.Scope.applicationOverride, "workload", "randomization.enabled", args.randomization_enabled)
+            cfg.add(config.Scope.applicationOverride, "workload", "randomization.rf", args.randomization_rf)
+            cfg.add(config.Scope.applicationOverride, "workload", "randomization.n", args.randomization_n)
+>>>>>>> randomized
             configure_workload_params(arg_parser, args, cfg)
             configure_connection_params(arg_parser, args, cfg)
             configure_telemetry_params(args, cfg)
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -572,6 +572,7 @@ def create_arg_parser():
         default=False,
         action="store_true")
     test_execution_parser.add_argument(
+        "--randomization-repeat-frequency",
         "--randomization-rf",
         help=f"The repeat_frequency for query randomization. Ignored if randomization is off"
              f"(default: {workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_RF}).",
--- a/osbenchmark/workload/loader.py
+++ b/osbenchmark/workload/loader.py
@@ -25,6 +25,7 @@ import glob
 import json
 import logging
 import os
+import random
 import re
 import sys
 import tempfile
@@ -74,7 +75,7 @@ class WorkloadProcessor:
 
 class WorkloadProcessorRegistry:
     def __init__(self, cfg):
-        self.required_processors = [TaskFilterWorkloadProcessor(cfg), TestModeWorkloadProcessor(cfg)]
+        self.required_processors = [TaskFilterWorkloadProcessor(cfg), TestModeWorkloadProcessor(cfg), QueryRandomizerWorkloadProcessor(cfg)]
         self.workload_processors = []
         self.offline = cfg.opts("system", "offline.mode")
         self.test_mode = cfg.opts("workload", "test.mode.enabled", mandatory=False, default_value=False)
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: osbenchmark/workload/loader.py [hunk 1]
```diff
@@ -918,8 +919,181 @@ def on_after_load_workload(self, workload):
...
+            if test_procedure.default:
+                default_test_procedure = test_procedure # TODO - not sure if this is correct
+                break
+
+        for task in default_test_procedure.schedule:
+            for leaf_task in task:
+                try:
+                    op_type = workload.OperationType.from_hyphenated_string(leaf_task.operation.type)
+                except KeyError:
+                    op_type = None
```
IanHoang, line 1083: Nit: For debugging purposes, it would be nice to have a logging statement in the exception here.

osbenchmark/workload/loader.py
```diff
@@ -918,8 +919,181 @@ def on_after_load_workload(self, workload):
...
+
+    def extract_fields_and_paths(self, params):
+        # Search for fields used in range queries, and the paths to those fields
+        # Return pairs of (field, path_to_field)
+        # TODO: Maybe only do this the first time, and assume for a given task, the same query structure is used.
+        # We could achieve this by passing in the task name to get_randomized_values as a kwarg?
+        try:
+            root = params["body"]["query"]
+        except KeyError:
+            raise exceptions.SystemSetupError("Cannot extract range query fields from these params, missing params[\"body\"][\"query\"]")
```
IanHoang, line 1002: Might be helpful to elaborate in which files users can commonly fix this error similar to what was added in line 73 of params.py.

osbenchmark/workload/loader.py
```diff
@@ -918,8 +919,181 @@ def on_after_load_workload(self, workload):
...
+            H_list.append(H_list[-1] + 1 / (j ** m))
+        return H_list
+
+    def zipf_cdf_inverse(self, u, H_list):
+        # To map a uniformly distributed u from [0, 1] to some probability distribution we plug it into its inverse CDF.
+        # as the zipf cdf is discontinuous there is no real inverse but we can use this solution:
+        # https://math.stackexchange.com/questions/53671/how-to-calculate-the-inverse-cdf-for-the-zipf-distribution
+        # Precompute all values H_i,alpha for a fixed alpha and pass in as H_list
+        if (u < 0 or u >= 1):
+            raise Exception("Input u must have 0 <= u < 1")
```
IanHoang, line 952: Could we use a more descriptive exception rather than `Exception`? If this error is encountered, what are some suggestions that we could provide to the user to help them resolve it?

osbenchmark/benchmark.py
```diff
@@ -566,6 +566,21 @@ def add_workload_source(subparser):
              f"(default: {metrics.GlobalStatsCalculator.DEFAULT_THROUGHPUT_PERCENTILES}).",
         default=metrics.GlobalStatsCalculator.DEFAULT_THROUGHPUT_PERCENTILES
     )
+    test_execution_parser.add_argument(
+        "--randomization-enabled",
+        help="Runs the given workload with query randomization enabled (default: false).",
+        default=False,
+        action="store_true")
+    test_execution_parser.add_argument(
+        "--randomization-rf",
```
IanHoang, line 575: We should aim to use full-length arguments as best practice, such as `--randomization-repeat-frequency`, in addition to the short-hand format, like `--randomization-rf` or `-rrf`. That way users don't have to look up what the meaning of the flags as much.

osbenchmark/workload/params.py
```diff
@@ -63,6 +65,14 @@ def param_source_for_name(name, workload, params):
     else:
         return param_source(workload, params)
 
+def get_standard_value_source(op_name, field_name):
+    try:
+        return __STANDARD_VALUE_SOURCES[op_name][field_name]
+    except KeyError:
+        raise exceptions.SystemSetupError(
+            "Could not find standard value source for operation {}, field {}! Make sure this is registered in workload.py"
```
IanHoang, line 73: Nice! This is a helpful suggestion. It'd be good for us to add documentation related to this in the official docs. Once we have that, we can add a link to that guide to this error message.

IanHoang (changes requested):
Left a few comments

IanHoang:
> Also tested manually with several different workloads while watching the parameters, to make sure they were changing correctly and the fraction of ones being drawn randomly was correct. Finally, re-ran a cache-related workload that we had run before with a previous, non-generic way of randomizing queries, and made sure the number of hits, misses, evictions, etc matched what we observed then.

Could you include some finer details on how you validated this? Any outputs pasted in this PR or the associated issue would be helpful!

IanHoang:
Also, could you elaborate on what you mean by "standard value source"? Is this essentially just a customized value?

osbenchmark/workload/loader.py
peteralfonsi: Oops - I had pasted this in from a test script and forgot to change this to an OSB exception. Will fix

peteralfonsi:
> Also, could you elaborate on what you mean by "standard value source"? Is this essentially a customized value for a specific operation?

Pretty much. It's a function you call to generate a random pair of values for a certain field. In the actual workload, we generate 5000 of these at the beginning and save them, to use for repeated values (mimicking cache hits). To mimic a totally new query that would cause a cache miss, we run the source function again and use that new output. Assuming your standard value source function has a lot of possible outputs, it probably won't have been seen before. 

I've updated the PR description to give more details.

peteralfonsi:
> > Also tested manually with several different workloads while watching the parameters, to make sure they were changing correctly and the fraction of ones being drawn randomly was correct. Finally, re-ran a cache-related workload that we had run before with a previous, non-generic way of randomizing queries, and made sure the number of hits, misses, evictions, etc matched what we observed then.
> 
> Could you include some finer details on how you validated this? Any outputs pasted in this PR or the associated issue would be helpful!

Just added some printouts and descriptions to the original PR comment. Let me know if you want more info!

osbenchmark/benchmark.py
IanHoang: @peteralfonsi Integration tests are failing due to extra parameter added has `--` and is not considered a "short-form" flag. Think you'll need to change this to "-rf" or some iteration with a single hyphen.

IanHoang:
> > > Also tested manually with several different workloads while watching the parameters, to make sure they were changing correctly and the fraction of ones being drawn randomly was correct. Finally, re-ran a cache-related workload that we had run before with a previous, non-generic way of randomizing queries, and made sure the number of hits, misses, evictions, etc matched what we observed then.
> > 
> > 
> > Could you include some finer details on how you validated this? Any outputs pasted in this PR or the associated issue would be helpful!
> 
> Just added some printouts and descriptions to the original PR comment. Let me know if you want more info!

This is great stuff! Thanks for the concise and supporting information added to the PR description

IanHoang:
Left a single comment on a fix regarding the integration tests.

osbenchmark/workload/loader.py
```diff
@@ -999,7 +1000,9 @@ def extract_fields_and_paths(self, params):
         try:
             root = params["body"]["query"]
         except KeyError:
-            raise exceptions.SystemSetupError("Cannot extract range query fields from these params, missing params[\"body\"][\"query\"]")
+            raise exceptions.SystemSetupError(
+                f"Cannot extract range query fields from these params: {params}\n, missing params[\"body\"][\"query\"]\n"
+                f"Make sure the operation in operations/default.json is well-formed")
```
IanHoang, line 1005: Nice!

osbenchmark/benchmark.py
peteralfonsi: Ahh - thanks, that probably would have taken me a while to figure out!

IanHoang: After second glance, `--randomization-repeat-frequency` is a bit lengthy and can be cumbersome for users to type out every time. Short-form flags should be used for more commonly used flags like `--target-hosts` or `--workload`. `--randomization-repeat-frequency` is a specific use-case and I think it will be fine to keep it as `--randomization-rf`.

For this option, we can either do what you did before and keep `--randomization-rf` or we can have the full-length version `--randomization-repeat-frequency` and a short-form version `-rf`. I'll leave it up to your preference!

IanHoang (approved):
LGTM. Thanks for the contribution!

PR #461, opened 2024-02-09
Add support for a data corpora source-url field to specify a corpus file directly

gkamat:
### Description
Files in data corpora may have paths that don't conform to the "corpus URL / filename" convention, for instance with object stores.  This change adds support for a `source-url` field that takes precedence over the corpus URL if it is specified.  Existing workloads will continue to work as is.

### Testing
- [X] New functionality includes additional unit tests.

Ran unit tests and integ tests, and a couple of workloads in test mode.
---


tests/workload/loader_test.py
```diff
@@ -289,6 +289,87 @@ def test_download_document_archive_if_no_file_available(self, is_file, get_size,
...
+
+        prepare_file_offset_table.return_value = 5
+
+        p = loader.DocumentSetPreparator(workload_name="unit-test",
+                                         downloader=loader.Downloader(offline=False, test_mode=False),
+                                         decompressor=loader.Decompressor())
+
+        p.prepare_document_set(document_set=workload.Documents(source_format=workload.Documents.SOURCE_FORMAT_BULK,
+                                                            base_url="http://benchmarks.elasticsearch.org/corpora",
+                                                    source_url="http://benchmarks.elasticsearch.org/corpora/unit-test/docs.json.bz2",
```
IanHoang, line 321: Nit: Although this is for a unittest, is there another link we could use to test it out with instead of one from elastic?

IanHoang (approved):

tests/workload/loader_test.py
gkamat: That's a good point.  These were cloned from other tests that have that link.  Will go through all and replace them in a separate commit.

PR #464, opened 2024-02-17
Make neighbor's dataset path as optional

VijayanB:
### Description
If neighbor's dataset path or corpus is not set,
use dataset path as neighbor's dataset path. This will avoid
users to set neighbors dataset path if same h5 file with
different dataset is used.

### Testing

---

### DIFF
--- a/osbenchmark/workload/params.py
+++ b/osbenchmark/workload/params.py
@@ -1052,7 +1052,7 @@ class VectorSearchPartitionParamSource(VectorDataSetPartitionParamSource):
             self.PARAMS_NAME_NEIGHBORS_DATA_SET_FORMAT, params, self.data_set_format)
         self.neighbors_data_set_path = params.get(self.PARAMS_NAME_NEIGHBORS_DATA_SET_PATH)
         self.neighbors_data_set_corpus = params.get(self.PARAMS_NAME_NEIGHBORS_DATA_SET_CORPUS)
-        self._validate_data_set(self.neighbors_data_set_path, self.neighbors_data_set_corpus)
+        self._validate_neighbors_data_set(self.neighbors_data_set_path, self.neighbors_data_set_corpus)
         self.neighbors_data_set = None
         operation_type = parse_string_parameter(self.PARAMS_NAME_OPERATION_TYPE, params,
                                                 self.PARAMS_VALUE_VECTOR_SEARCH)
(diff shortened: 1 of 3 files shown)


jmazanec15 (approved):

osbenchmark/workload/params.py
```diff
@@ -1072,11 +1072,8 @@ def __init__(self, workloads, params, query_params, **kwargs):
             neighbors_corpora = self.extract_corpora(self.neighbors_data_set_corpus, self.neighbors_data_set_format)
             self.corpora.extend(corpora for corpora in neighbors_corpora if corpora not in self.corpora)
 
-    def _validate_neighbors_data_set(self):
-        if not self.data_set_path and not self.data_set_corpus:
-            raise exceptions.ConfigurationError(
-                "Dataset is missing. Provide either dataset file path or valid corpus.")
-        if self.data_set_path and self.data_set_corpus:
+    def _validate_neighbors_data_set(self, file_path, corpus):
+        if file_path and corpus:
```
IanHoang, line 1076: Since `self.data_set_path` and `self.data_set_corpus` have been replaced with `file_path` and `corpus`, should the exception be updated to reflect those new arguments?

tests/workload/params_test.py
```diff
@@ -2862,20 +2862,20 @@ def test_params_default(self):
             Context.QUERY,
             self.data_set_dir
         )
-        neighbors_data_set_path = create_data_set(
+        create_data_set(
```
IanHoang, line 2865: When is this being used?

osbenchmark/workload/params.py
VijayanB: Ack

tests/workload/params_test.py
VijayanB: Good point, In this case we are creating neighbor's dataset as part of query's dataset. In first step, you see that query dataset is created and it returns the dataset path. In step two, we pass same file and in this case, we don't create new file, instead create new dataset with key as neighbors in same file.

IanHoang (approved):

tests/utils/dataset_helper.py
```diff
@@ -188,12 +188,16 @@ def create_data_set(
         dimension: int,
         extension: str,
         data_set_context: Context,
-        data_set_dir
+        data_set_dir,
```
gkamat, line 191: Please add a type hint for clarity.

PR #450, opened 2024-01-29
Added a new metric: Client Processing Time

saimedhi:
### Description
Introduced a new metric: **Client Processing Time.**

Client Processing Time: The delta between total request time and service time.

Total Request Time: Defined as the duration between the runner sending a request to the client OpenSearch-py and receiving the response.

Service Time: Represents the interval from the server receiving the request to the server sending the response. Note: There was a discrepancy in the documentation regarding Service Time, and I have clarified my observations in the associated [PR](https://github.com/opensearch-project/documentation-website/pull/5916#discussion_r1438397101).

### Issues Resolved
#432

PR #470, opened 2024-02-26
Fix client processing time calculation in ForceMerge Runner

saimedhi:
### Description
Enhance client processing time calculation in the ForceMerge Runner to address a critical issue. This improvement rectifies an error encountered when utilizing polling, where the force merge times out during the initial call **without raising an exception**, and subsequently completes successfully. Previously, the request_context_holder.on_client_request_end() wasn't appropriately calculated in such scenarios. For further context, please refer to the discussion [here](https://github.com/opensearch-project/opensearch-benchmark/pull/450#issuecomment-1962750962).
### Issues Resolved
https://github.com/opensearch-project/opensearch-benchmark/pull/450#issuecomment-1962750962

### Testing
- Tested by @VijayanB [here]( https://github.com/opensearch-project/opensearch-benchmark/pull/450#issuecomment-1964549671)

---

### DIFF
--- a/osbenchmark/worker_coordinator/runner.py
+++ b/osbenchmark/worker_coordinator/runner.py
@@ -698,6 +698,7 @@ class ForceMerge(Runner):
                 tasks = await opensearch.tasks.list(params={"actions": "indices:admin/forcemerge"})
                 if len(tasks["nodes"]) == 0:
                     # empty nodes response indicates no tasks
+                    request_context_holder.on_client_request_end()
                     complete = True
         else:
             request_context_holder.on_client_request_start()


VijayanB (approved):
LGTM

saimedhi:
@IanHoang, please take a look. Thank you.

IanHoang:
@VijayanB were you able to verify that an exception is raised when force merge encounters a timeout with this PR's changes?

VijayanB:
@IanHoang Sorry i missed your comment. Yes, it is reproducible in vector search for large segments.

osbenchmark/worker_coordinator/runner.py
```diff
@@ -698,6 +698,7 @@ async def __call__(self, opensearch, params):
                 tasks = await opensearch.tasks.list(params={"actions": "indices:admin/forcemerge"})
                 if len(tasks["nodes"]) == 0:
                     # empty nodes response indicates no tasks
+                    request_context_holder.on_client_request_end()
```
gkamat, line 701: This should be called just prior to the preceding `pass` statement, else there will be a mismatch between service time and the client processing time when an timeout exception is caught.

gkamat (changes requested):

osbenchmark/worker_coordinator/runner.py
saimedhi: Hey @gkamat, I've updated some code in my previous PR to ensure that request start is always calculated after the client request start, and that request end is always before the client request end. I'm sorry if I didn't address your concern properly. I believe that to get accurate metrics, request_context_holder.on_client_request_end() should be added after the task is completed. What do you think, @IanHoang and @VijayanB? Would love to hear your thoughts.

gkamat: @VijayanB, the time for the force merge reported is going to be inaccurate in any case, since the resolution with be the poll period, which is 10 sec.  It is just the degree of error, which should be weighed against the following.

The issue is how the client processing time is computed for multi-part operations.   In this case there is a truncated operation caused by the exception.  That should be handled on its own and the corresponding service time and associated client overhead computed.

Since there is a subsequent loop, both of these variables should be accumulators. Each time through the loop the poll period should be added to the service time.  Each call to the `list` operation will be a contributor as well, furthermore the client overhead of that call will be another metric that needs to be updated.

The root cause of this is how multi-part requests are handled, as was indicated to @saimedhi earlier.  Otherwise the concept of what the client overhead is for a particular call becomes ambiguous.

gkamat: Since this is holding the release back and the submitted fix will resolve the immediate problem, we can merge this in for now and address the larger issue subsequently.

gkamat (approved):

osbenchmark/worker_coordinator/runner.py
gkamat: @VijayanB yes, polling is not part of force merge and the issue being referred to is not so much the inaccuracy in the duration reported as the accuracy of client overhead computation.  This is somewhat similar to, say, how scroll queries are handled, where the same issues are at play.

PR #477, opened 2024-03-08
Jenkins file point to 1.3 branch temporarily instead of main

IanHoang:
### Description
Pushed patch release from new `1.3` branch but Jenkins is running into following error:
```
#14 [linux/amd64 4/6] RUN if [ -z "1.4.0" ] ; then python3 -m pip install opensearch-benchmark ; else python3 -m pip install opensearch-benchmark==1.4.0 ; fi
#0 1.734 ERROR: Ignored the following versions that require a different python version: 0.0.1 Requires-Python >=3.8,<3.10; 0.0.2 Requires-Python >=3.8,<3.10; 0.1.0 Requires-Python >=3.8,<3.10
#0 1.735 ERROR: Could not find a version that satisfies the requirement opensearch-benchmark==1.4.0 (from versions: 0.2.0, 0.3.0, 0.4.0, 0.4.1, 0.5.0, 1.0.0, 1.1.0, 1.2.0, 1.3.0, 1.3.1)
```

We need to update jenkins file to point to `1.3` instead of `main` as a quick workaround. In the future, we should have a way where the Github Actions can take a "branch" parameter so that `main` isn't the only branch we can push releases from.

### Issues Resolved

### Testing

---

### DIFF
--- a/jenkins/release.JenkinsFile
+++ b/jenkins/release.JenkinsFile
@@ -15,7 +15,7 @@ standardReleasePipelineWithGenericTrigger(
         wait: true,
         parameters: [
             string(name: 'DOCKER_BUILD_GIT_REPOSITORY', value: 'https://github.com/opensearch-project/opensearch-build'),
-            string(name: 'DOCKER_BUILD_GIT_REPOSITORY_REFERENCE', value: 'main'),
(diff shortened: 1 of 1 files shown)


gkamat (approved):

IanHoang:
We are going with a different approach to fix this. Closing now.

PR #552, opened 2024-06-14
Excluding opensearch-benchmarkd process while finding benchmark process

backslasht:
### Description
On Linux, /proc/PID/status truncates the command name to 15 characters which matches both opensearch-benchmark and opensearch-benchmarkd. Adding an additional check to exclude opensearch-benchmarkd process.

### Testing
Validated by running benchmark daemon on three nodes and then by running execute test with ```--load-worker-coordinator-hosts``` option. The test ran fine on the worker nodes.

---

### DIFF
--- a/osbenchmark/utils/process.py
+++ b/osbenchmark/utils/process.py
@@ -133,9 +133,12 @@ def run_subprocess_with_logging(command_line, header=None, level=logging.INFO, s
 
 def is_benchmark_process(p):
     cmdline = p.cmdline()
-    # On Linux, /proc/PID/status truncates the command name to 15 characters.
+    # On Linux, /proc/PID/status truncates the command name to 15 characters which matches both opensearch-benchmark
+    # and opensearch-benchmarkd. Hence, checking command line to exclude opensearch-benchmarkd process.
     return p.name() == "opensearch-benchmark" or \
-        p.name() == "opensearch-benc" or \
+        (p.name() == "opensearch-benc" and
+         len(cmdline) > 1 and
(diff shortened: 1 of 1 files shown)


rishabh6788 (approved):
LGTM.

IanHoang (approved):
Thanks for this!

osbenchmark/utils/process.py
```diff
@@ -133,9 +133,12 @@ def run_subprocess_with_logging(command_line, header=None, level=logging.INFO, s
...
 def is_benchmark_process(p):
     cmdline = p.cmdline()
-    # On Linux, /proc/PID/status truncates the command name to 15 characters.
+    # On Linux, /proc/PID/status truncates the command name to 15 characters which matches both opensearch-benchmark
+    # and opensearch-benchmarkd. Hence, checking command line to exclude opensearch-benchmarkd process.
     return p.name() == "opensearch-benchmark" or \
-        p.name() == "opensearch-benc" or \
+        (p.name() == "opensearch-benc" and
+         len(cmdline) > 1 and
+         os.path.basename(cmdline[1]) != "opensearch-benchmarkd") or \
```
gkamat, line 141: Good catch!  Can this section simply be eliminated, since the process name reported by `/proc/PID/status` will be different in various circumstances anyway, and is not reliable.  For instance:
```
#!/usr/bin/python3:
Name:   opensearch-benc

#!/usr/bin/env python3: 
Name:   python3
```

backslasht: Makes sense. Raised https://github.com/opensearch-project/opensearch-benchmark/pull/553 to remove the truncated check.

PR #553, opened 2024-06-15
Removing the duplicate process check based on truncated name and relying on cmdline check

backslasht:
### Description

On linux, the process name is truncated to 15 characters and hence it matches both opensearch-benchmark and opensearch-benchmarkd. This results in killing the daemon process and breaks the distributed run setup. Removing the name based check on the truncated name and relying on command line which provides the full command.

Keeping the name based check on full name as is.

### Testing
Tested in Linux system and verified daemon process is excluded in duplicate process check.

---


gkamat (approved):
Thanks for the fix.

PR #564, opened 2024-06-24
Adds --randomization-alpha flag controlling distribution of repeated queries

peteralfonsi:
### Description
https://github.com/opensearch-project/opensearch-benchmark/pull/455 introduced a way to controllably randomize range queries in workloads, which is useful in testing caching behavior. One relevant parameter is Zipf alpha, which changes the shape of the Zipf distribution used for repeated queries. If alpha is small, queries will be more evenly distributed, while if it's large the few most common queries will be preferred more. 

I intended to include this as a flag in my original PR but I apparently forgot to do this, and currently alpha is hardcoded to 1. This PR exposes `--randomization-alpha` so the user can set this value explicitly. 
### Issues Resolved
N/A

### Testing

Manually tested in a debug version of this branch, by checking the value of the flag made it to QueryRandomizerWorkloadProcessor.zipf_alpha, and checking the indices generated for different values of alpha match what we would expect.

---

### DIFF
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -590,7 +590,7 @@ def create_arg_parser():
     test_execution_parser.add_argument(
         "--randomization-alpha",
         help=f"The alpha parameter used for the Zipf distribution for query randomization. Low values spread the distribution out, "
-             f"high values favor the most common queries more. "
+             f"high values favor the most common queries. "
              f"Ignored if randomization is off (default: {workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_ALPHA}).",
         default=workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_ALPHA)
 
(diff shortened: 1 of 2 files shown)


osbenchmark/benchmark.py
```diff
@@ -587,6 +587,12 @@ def add_workload_source(subparser):
         help=f"The number of standard values to generate for each field for query randomization."
              f"Ignored if randomization is off (default: {workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_N}).",
         default=workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_N)
+    test_execution_parser.add_argument(
+        "--randomization-alpha",
+        help=f"The alpha parameter used for the Zipf distribution for query randomization. Low values spread the distribution out, "
+             f"high values favor the most common queries more. "
```
gkamat, line 593: nit: "more" is redundant.  "high values favor the most common queries."

gkamat (approved):

IanHoang (approved):

PR #604, opened 2024-07-31
Refactor build env logic to switch between testing and production

peterzhuamazon:
### Description
Add Docker Run UnitTest on x64 Runner

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/4723

### Testing

---


.github/workflows/docker-test.yml
```diff
@@ -30,9 +30,12 @@ jobs:
...
+      - name: Docker Build ${{ matrix.platform }}
         run: |
             docker buildx version
             cp -a opensearch-benchmark-git/* ./
             echo "Disable VERSION arg to enter docker build test mode"
-            docker buildx build --platform ${{ matrix.architecture }} --build-arg BUILD_DATE=`date -u +%Y-%m-%dT%H:%M:%SZ` -f "docker/Dockerfile" .
+            PLATFORM=${{ matrix.platform }}
+            PLATFORM=`echo $PLATFORM | tr '/' '-'`
+            docker buildx build --platform ${{ matrix.platform }} --build-arg BUILD_ENV=testing --build-arg BUILD_DATE=`date -u +%Y-%m-%dT%H:%M:%SZ` -f "docker/Dockerfile" -t "osb/osb-$PLATFORM" -o type=docker .
+            docker images | grep "osb/osb-$PLATFORM"
```
gkamat, line 41: Probably can be streamlined, but fine for testing.
What does the `docker images` line do?  Is it simply to print the output of the `grep`?

gkamat (approved):

.github/workflows/docker-test.yml
peterzhuamazon: Yes @gkamat 

It is to setup the next step of running docker container based on the image just created.

In multi-arch build most of the time image has to be pushed upon completion to registry.
I force the image to be saved locally so we can use the image to run containers.

Once the container is up we can use it to run it.
This is a pre-req before that happened.

And I agree the code can be streamlined a bit, which I will improve in the next version.

Thanks.

docker/docker-compose-tests.yml
IanHoang, line 1: We need this file for integration tests

PR #607, opened 2024-08-05
Add approximate filters

finnroblin:
### Description
This change adds support for ingesting non-vector fields from an HDF5 dataset into OpenSearch k-NN. It also adds support for writing post-filtering queries. These changes aid the transition from k-NN's perf tool to using OSB for release benchmarking. 

Please see [OSB Workloads PR 364](https://github.com/opensearch-project/opensearch-benchmark-workloads/pull/364) for usage examples. 

### Testing
- [X] New functionality includes testing

Unit tests were added that cover the new param and runner methods. Additionally all of the functionality has been tested via the parameter files in [OSB Workloads PR 364](https://github.com/opensearch-project/opensearch-benchmark-workloads/pull/364).

---

### DIFF
--- a/osbenchmark/utils/dataset.py
+++ b/osbenchmark/utils/dataset.py
@@ -27,6 +27,7 @@ class Context(Enum):
     MAX_DISTANCE_NEIGHBORS = 4
     MIN_SCORE_NEIGHBORS = 5
     PARENTS = 6
+    ATTRIBUTES = 7
 
 
 class DataSet(ABC):
@@ -133,6 +134,7 @@ class HDF5DataSet(DataSet):
     def reset(self):
         self.current = self.BEGINNING
 
+    # pylint: disable=R0911
     @staticmethod
     def parse_context(context: Context) -> str:
         if context == Context.NEIGHBORS:
@@ -151,6 +153,9 @@ class HDF5DataSet(DataSet):
 
         if context == Context.MIN_SCORE_NEIGHBORS:
             return "min_score_neighbors"
+
+        if context == Context.ATTRIBUTES:
+            return "attributes"
 
         raise Exception("Unsupported context")
 
--- a/osbenchmark/utils/parse.py
+++ b/osbenchmark/utils/parse.py
@@ -22,7 +22,7 @@ def parse_string_parameter(key: str, params: dict, default: str = None) -> str:
 
 def parse_int_parameter(key: str, params: dict, default: int = None) -> int:
     if key not in params:
-        if default:
+        if default is not None:
             return default
         raise ConfigurationError(
             "Value cannot be None for param {}".format(key)
@@ -46,3 +46,17 @@ def parse_float_parameter(key: str, params: dict, default: float = None) -> floa
         return params[key]
 
     raise ConfigurationError("Value must be a float for param {}".format(key))
+
+
+def parse_bool_parameter(key: str, params: dict, default: bool = None) -> bool:
+    if key not in params:
+        if default is not None:
+            return default
+        raise ConfigurationError(
+            "Value cannot be None for param {}".format(key)
+        )
+
+    if isinstance(params[key], bool):
+        return params[key]
+
+    raise ConfigurationError("Value must be a bool for param {}".format(key))
--- a/osbenchmark/worker_coordinator/runner.py
+++ b/osbenchmark/worker_coordinator/runner.py
@@ -1254,6 +1254,17 @@ class Query(Runner):
                     return _get_field_value(content["_source"], field_name)
                 return None
 
+            def binary_search_for_last_negative_1(neighbors):
+                low = 0
+                high = len(neighbors)
+                while low < high:
+                    mid = (low + high) // 2
+                    if neighbors[mid] == "-1":
+                        high = mid
+                    else:
+                        low = mid + 1
+                return low - 1
+
             def calculate_topk_search_recall(predictions, neighbors, top_k):
                 """
                 Calculates the recall by comparing top_k neighbors with predictions.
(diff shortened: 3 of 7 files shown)


osbenchmark/utils/dataset.py
```diff
@@ -24,8 +24,7 @@ class Context(Enum):
     INDEX = 1
     QUERY = 2
     NEIGHBORS = 3
-    MAX_DISTANCE_NEIGHBORS = 4
```
finnroblin, line 27: I'll rebase against main to pull in these changes on the next revision, didn't catch until I submitted the PR. Will also fixup the commits into a single one in the next revision.

osbenchmark/utils/parse.py
```diff
@@ -22,7 +22,7 @@ def parse_string_parameter(key: str, params: dict, default: str = None) -> str:
 
 def parse_int_parameter(key: str, params: dict, default: int = None) -> int:
     if key not in params:
-        if default:
+        if default is not None:
```
finnroblin, line 25: This change is technically not related to this PR, but the previous `if` statement was flaky if `default == 0`. All unit tests pass in both configurations.

osbenchmark/worker_coordinator/runner.py
```diff
@@ -1254,6 +1254,17 @@ def _get_field_value(content, field_name):
                     return _get_field_value(content["_source"], field_name)
                 return None
 
+            def binary_search_for_last_negative_1(neighbors):
```
finnroblin, line 1257: I'd rather use a Python utility here but I couldn't find one in the standard library. Any suggestions?

osbenchmark/worker_coordinator/runner.py
```diff
@@ -1270,7 +1281,20 @@ def calculate_topk_search_recall(predictions, neighbors, top_k):
                     self.logger.info("No neighbors are provided for recall calculation")
                     return 0.0
                 min_num_of_results = min(top_k, len(neighbors))
+                last_neighbor_is_negative_1 = int(neighbors[min_num_of_results-1]) == -1
```
finnroblin, line 1284: This logic is necessary when the dataset is first generated and then neighbors are filtered out. This occurs in some of the perf tools datasets like [here](https://github.com/opensearch-project/k-NN/blob/main/benchmarks/perf-tool/add-filters-to-dataset.py). These datasets first generate true neighbors for the test vector, such as `[5, 2, 3]`, and then post-filter the neighbors with attributes that do not pass the filter. The documents without the attributes are replaced by `-1`. However looking more closely at the dataset generation code I'm not sure that all of the `-1` entries are at the end of the array. @martin-gaievski I see you wrote the filtering code, could you please clarify? Thanks

martin-gaievski: they should be at the end of collection of nearest neighbors for each query, although I'm not clearly remember sorting them. Can you share example where -1 are in the row, but they are mixed with some other ids

finnroblin: I didn't see any in testing, but wanted to ask since I couldn't find a sort. I took another look at the [code for filtering neighbors](https://github.com/opensearch-project/k-NN/blob/a16b8aa965f66268799c001756c78174c7baba19/benchmarks/perf-tool/add-filters-to-dataset.py#L161) and it looks like the initial array with all `-1` entries is overwritten from left to right, so any remaining `-1` ids are at the end. I don't think we'll use these filter datasets anyways (since Heemin's nested fields script generates `k` true neighbors after filtering instead of generating `k` true neighbors and then post-filtering to have `<k` neighbors), but I'll call out the assumption that all `-1` are at the end in a comment.

osbenchmark/worker_coordinator/runner.py
jmazanec15: Im not aware of anything. Also, given that the list may not be sorted, it might be harder to do. That being said, I think this is good.

osbenchmark/workload/params.py
```diff
@@ -1186,45 +1179,68 @@ def params(self):
...
+
+        return knn_search_query
+
+    def _knn_query_with_filter(self, vector, knn_query, filter_type, filter_body) -> dict:
+        if filter_type == "script":
+            return {
+                "script_score": {
+                    "query": {"bool": {"filter": filter_body}},
+                    "script": {
+                        "source": "knn_score",
```
jmazanec15, line 1225: Why use knn script score here?

finnroblin: I added it in OSB since it's mentioned in the [filters documentation](https://opensearch.org/docs/latest/search-plugins/knn/filter-search-knn/#scoring-script-filter) but I haven't written a workload yet. Is Knn script score something that we would like to benchmark?

finnroblin: I spoke with @VijayanB and it seems like @navneet1v is adding a feature in 2.17 where exact search can be performed via a knn query (as opposed to the script score query). This change should be much easier to benchmark than exact search via a `script_score` query since it would just require a new workload without any OSB runner changes. I'll remove this script score code in the next revision.

navneet1v: @finnroblin but we will still need script score based benchmarks.

navneet1v: @jmazanec15, @finnroblin do we not want to add script score based benchmarks?

jmazanec15: We probably shoud have this as well. But I was thinking that exact search should use new method for implementing it

osbenchmark/worker_coordinator/runner.py
```diff
@@ -1270,7 +1281,20 @@ def calculate_topk_search_recall(predictions, neighbors, top_k):
...
                 truth_set = neighbors[:min_num_of_results]
+                if last_neighbor_is_negative_1:
+                    self.logger.info("Last neighbor is -1")
```
VijayanB, line 1287: remove this logging, if you want please make it debug

VijayanB:
@finnroblin Can you rebase this PR?

VijayanB (approved):

Issue #380, opened 2023-09-29
[BUG] Unable to run opensearch-benchmark [OSB] execute-test against OSB provisioned cluster due to failing health check


cgchinmay:
**Describe the bug**
opensearch-benchmark execute test cannot run as the health check with provisioned cluster doesn't succeed

Tried running opensearch-benchmark command execute-test as below 
```
opensearch-benchmark execute-test --pipeline=benchmark-only --workload=geonames --target-host=127.0.0.1:9200 --test-mode --kill-running-processes
```

But the above command gets stuck and checking the logs I see failing health checks
```
tail -f ~/.benchmark/logs/benchmark.log
2023-09-28 23:07:35,325 -not-actor-/PID:31934 opensearch WARNING GET http://127.0.0.1:9200/_cluster/health?wait_for_nodes=%3E%3D1 [status:503 request:30.013s]
2023-09-28 23:08:05,341 -not-actor-/PID:31934 opensearch WARNING GET http://127.0.0.1:9200/_cluster/health?wait_for_nodes=%3E%3D1 [status:503 request:30.015s]
2023-09-28 23:08:38,359 -not-actor-/PID:31934 opensearch WARNING GET http://127.0.0.1:9200/_cluster/health?wait_for_nodes=%3E%3D1 [status:503 request:30.017s]
2023-09-28 23:09:54,761 -not-actor-/PID:31934 opensearch WARNING GET http://127.0.0.1:9200/_cluster/health?wait_for_nodes=%3E%3D1 [status:503 request:76.402s]
```

**To Reproduce**
Provision the cluster using opensearch-benchmark pypi utility 
```
opensearch-benchmark install --provision-config-instance=defaults --distribution-version=2.10.0 --node-name="osb-node-1" --network-host="127.0.0.1" --http-port=9200 --master-nodes="opensearch-node-1" --seed-hosts="127.0.0.1:9300" --quiet
{
  "installation-id": "7bedc677-82d4-48e8-866c-bf62250eca9d"
}
```
Started Single Node Cluster using opensearch benchmark 
```
opensearch-benchmark start --installation-id=7bedc677-82d4-48e8-866c-bf62250eca9d --test-execution-id=benchmark
```

Validated the the cluster status
```
curl localhost:9200
{
  "name" : "osb-node-1",
  "cluster_name" : "benchmark-provisioned-cluster",
  "cluster_uuid" : "_na_",
  "version" : {
    "distribution" : "opensearch",
[...]

PR #621, opened 2024-08-19
add default value for master-node when provisioning cluster using OSB

OVI3D0:
### Description
As mentioned by @gkamat, when provisioning a cluster in OSB, the master node specification should match the node name. This change ensures that these two values always match, by performing a check before building the cluster and ensuring the two values match before moving on.

### Issues Resolved
[#380](https://github.com/opensearch-project/opensearch-benchmark/issues/380)

### Testing
- [✔️] New functionality includes testing

tested with `make test` + `make it` and by provisioning clusters with mismatched values, as well as missing values for either node-name or master-node value.

---

### DIFF
--- a/osbenchmark/builder/builder.py
+++ b/osbenchmark/builder/builder.py
@@ -62,6 +62,14 @@ def install(cfg):
     master_nodes = cfg.opts("builder", "master.nodes")
     seed_hosts = cfg.opts("builder", "seed.hosts")
 
+    # Ensure node_name and master_nodes match, using node_name as the default
+    if node_name not in master_nodes:
+        print(
+            f"The provided --node-name '{node_name}' and --master-nodes '{master_nodes}' are different. "
+            f"Using '{node_name}' for both node name and initial master node."
+        )
+        master_nodes = [node_name]
+
     if build_type == "tar":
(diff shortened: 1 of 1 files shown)


IanHoang (approved):
Looks good to me!

osbenchmark/builder/builder.py
```diff
@@ -62,6 +62,14 @@ def install(cfg):
     master_nodes = cfg.opts("builder", "master.nodes")
     seed_hosts = cfg.opts("builder", "seed.hosts")
 
+    # Ensure node_name and master_nodes match, using node_name as the default
+    if node_name not in master_nodes:
+        print(
+            f"The provided --node-name '{node_name}' and --master-nodes '{master_nodes}' are different. "
+            f"Using '{node_name}' for both node name and initial master node."
+        )
+        master_nodes = [node_name]
```
cgchinmay, line 71: I am wondering, If this is the desired behavior then do we want to expose master_nodes ? Here we are reporting error but override the master_nodes values and continue with provisioning. I think we should either fail so that user is notified or not expose the master_nodes field at all. @IanHoang would like to know your thoughts as well

PR #632, opened 2024-08-29
Refactored the OSB build system

gkamat:
### Description
Restructured the entire OSB build system, including standalone, Docker and via GitHub actions.

### Testing
Tested out most combinations of the workflows above.

---

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,4 +1,4 @@
-name: Docker Build and Test
+name: Docker Build
 on:
   pull_request:
   workflow_dispatch:
@@ -28,14 +28,12 @@ jobs:
         with:
           version: 'v0.9.1'
       - uses: actions/checkout@v4
-        with:
-          path: 'opensearch-benchmark-git'
+
       - name: Docker Build ${{ matrix.platform }}
         run: |
             docker buildx version
-            cp -a opensearch-benchmark-git/* ./
-            echo "Disable VERSION arg to enter docker build test mode"
-            PLATFORM=${{ matrix.platform }}
-            PLATFORM=`echo $PLATFORM | tr '/' '-'`
-            docker buildx build --platform ${{ matrix.platform }} --build-arg BUILD_ENV=testing --build-arg BUILD_DATE=`date -u +%Y-%m-%dT%H:%M:%SZ` -f "docker/Dockerfile" -t "osb/osb-$PLATFORM" -o type=docker .
-            docker images | grep "osb/osb-$PLATFORM"
+            tag=osb/osb-`echo ${{ matrix.platform }} | tr '/' '-'`
+            set -x
+            docker buildx build --platform ${{ matrix.platform }} --build-arg VERSION=`cat version.txt` --build-arg BUILD_DATE=`date -u +%Y-%m-%dT%H:%M:%SZ` -f docker/Dockerfile -t "$tag" -o type=docker .
+            set +x
+            docker images | grep "$tag"
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,36 +2,48 @@
 # Install OpenSearch Benchmark to build a Docker image #
 ########################################################
 
-ARG VERSION
-ARG BUILD_ENV=production
-
-FROM python:3.11.2-slim as build_env_testing
-ONBUILD COPY opensearch-benchmark-git/ ./
-
-FROM python:3.11.2-slim as build_env_production
-ONBUILD RUN echo Production Environment
-
-FROM build_env_${BUILD_ENV}
+#
+# Stage 1: build packages and compile where needed
+#
+FROM python:3.11.2-slim AS build-stage
 WORKDIR /opensearch-benchmark
-ENV BENCHMARK_RUNNING_IN_DOCKER=True
 
 RUN apt-get -y update && \
-    apt-get install -y curl git gcc pbzip2 pigz && \
-    apt-get -y upgrade && \
-    rm -rf /var/lib/apt/lists/*
+    apt-get install -y curl git gcc pbzip2 pigz make jq && \
+    apt-get -y upgrade
+
+COPY . opensearch-benchmark
+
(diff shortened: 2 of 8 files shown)


.github/workflows/docker-build.yml
```diff
@@ -28,14 +28,12 @@ jobs:
...
         run: |
             docker buildx version
-            cp -a opensearch-benchmark-git/* ./
-            echo "Disable VERSION arg to enter docker build test mode"
-            PLATFORM=${{ matrix.platform }}
-            PLATFORM=`echo $PLATFORM | tr '/' '-'`
-            docker buildx build --platform ${{ matrix.platform }} --build-arg BUILD_ENV=testing --build-arg BUILD_DATE=`date -u +%Y-%m-%dT%H:%M:%SZ` -f "docker/Dockerfile" -t "osb/osb-$PLATFORM" -o type=docker .
-            docker images | grep "osb/osb-$PLATFORM"
+            tag=osb/osb-`echo ${{ matrix.platform }} | tr '/' '-'`
+            set -x
```
peterzhuamazon, line 36: I understand your changes here but this change combined with the Dockerfile change basically cause the docker-build check to be inaccurate.

It will only check docker build based on merged code, instead of the commit sent with the PR. You can only detect issues once you merge the code, which defeat the purpose of PR check.

My original design was meant to have test mode testing the PR commit, while prod mode use the original pypi method. I think we should bring back the test mode, but only change prod mode to build from source instead of pulling from pypi, which should be the best of both approaches. Thanks.

docker/Dockerfile
```diff
@@ -2,36 +2,26 @@
...
     apt-get -y upgrade && \
     rm -rf /var/lib/apt/lists/*
 
 RUN groupadd --gid 1000 opensearch-benchmark && \
     useradd -d /opensearch-benchmark -m -k /dev/null -g 1000 -N -u 1000 -l -s /bin/bash benchmark
 
-ENV PIP_ONLY_BINARY=h5py
-RUN if [ "$BUILD_ENV" = "testing" ] ; then echo Testing; ls -l; python3 -m pip install -e . ; \
-    else echo Production; if [ -z "$VERSION" ] ; then python3 -m pip install opensearch-benchmark ; else python3 -m pip install opensearch-benchmark==$VERSION ; fi; fi
+RUN set -e; git clone https://github.com/opensearch-project/opensearch-benchmark; cd opensearch-benchmark; make install; cd ..; rm -rf opensearch-benchmark
```
peterzhuamazon, line 17: If we really want to build from source, then we should introduce two stages of docker build.

Stage 1 do the build with tools like make/jq installed, while Stage 2 copy only the binaries / installed folders cleanly. With this current change it will increase the size of the docker layer, which increase the size of the docker image as a whole.

peterzhuamazon: I do, still prefer that the prod mode is using release pypi build, to be much simpler and clean, but retain test mode with the build from source approach, if make sense.

peterzhuamazon (commented):
Hi @gkamat ,

Add some comments.

Thanks.

.github/workflows/docker-build.yml
gkamat: The updated changes should pull in the PR commit now.

docker/Dockerfile
gkamat: Changed this to multi-stage, so the image size has gone down:
```
<none>                <none>        d294c275b627   2 hours ago      694MB

osb/osb-linux-amd64   latest        b49285646311   23 seconds ago   475MB
```

I think there's some value to keeping this independent of the PyPI build, so left that alone for now.  If it turns out to be an issue, it can be changed fairly easily.

peterzhuamazon:
Seems like docker-lint does not pass.

docker/Dockerfile
```diff
@@ -2,36 +2,48 @@
...
-ENV PIP_ONLY_BINARY=h5py
-RUN if [ "$BUILD_ENV" = "testing" ] ; then echo Testing; ls -l; python3 -m pip install -e . ; \
-    else echo Production; if [ -z "$VERSION" ] ; then python3 -m pip install opensearch-benchmark ; else python3 -m pip install opensearch-benchmark==$VERSION ; fi; fi
-
 RUN mkdir -p /opensearch-benchmark/.benchmark && \
     chown -R 1000:0 /opensearch-benchmark/.benchmark
 
+COPY --from=build-stage /opensearch-benchmark/opensearch-benchmark/yappi-*.whl /opensearch-benchmark/opensearch-benchmark/opensearch_benchmark-*.whl ./
+
+# There is no binary package currently available for yappi on ARM.
```
peterzhuamazon, line 35: Will this cause issue on arm64 based on your comment?

Since we release multi-arch image to the public.

Thanks.

gkamat: No, this is handled by copying the locally-built `yappi` wheel from the first stage, so the C compiler does not need to be installed in the second stage, thereby keeping the image size small.

docker/Dockerfile
```diff
@@ -2,36 +2,48 @@
 # Install OpenSearch Benchmark to build a Docker image #
 ########################################################
 
-ARG VERSION
-ARG BUILD_ENV=production
+#
+# Stage 1: build packages and compile where needed
```
rishabh6788, line 6: did we get any significant size reduction in final image by splitting into multi-stage. It is fairly simple dockerfile.

docker/Dockerfile
peterzhuamazon: Got it 👍

docker/Dockerfile
peterzhuamazon: See this comment:
https://github.com/opensearch-project/opensearch-benchmark/pull/632#discussion_r1739523746

peterzhuamazon (approved):
Thanks @gkamat !

docker/Dockerfile
gkamat: Goes down from about 700 MB to 500 MB with two stages.

rishabh6788 (approved):

PR #651, opened 2024-09-13
Update release guide with more details

IanHoang:
### Description
After encountering a few issues in the release process, we updated the release guide.

---

### DIFF
--- a/RELEASE_GUIDE.md
+++ b/RELEASE_GUIDE.md
@@ -55,7 +55,7 @@ NOTE: The version number below is in semantic format, for instance, `1.2.0`.
 
 1. Create a tag: `git tag <VERSION> main`
     1. Ensure that this is done in the main official opensearch-benchmark repository
-    2. This should be the `<VERSION>` tag matches the version in version.txt.
+    2. This should be the `<VERSION>` tag that matches the entry in version.txt.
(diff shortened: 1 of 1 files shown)


RELEASE_GUIDE.md
```diff
@@ -55,7 +55,7 @@ NOTE: The version number below is in semantic format, for instance, `1.2.0`.
 
 1. Create a tag: `git tag <VERSION> main`
     1. Ensure that this is done in the main official opensearch-benchmark repository
-    2. This should be the new version that matches the version in version.txt.
+    2. This should be the `<VERSION>` tag matches the version in version.txt.
```
gkamat, line 58: This should be the `<VERSION>` tag that matches the entry in version.txt.

gkamat (commented):
Might be good to add some material about the steps in the release process -- the github action and the webhook that are distinct, the latter triggering jenkins, and how this can be aborted by not approving the issue, in case someone needs to deal with that. 

The editing of the release issue should also be added to "Error Handling" (or add a pointer back to the previous section.)

RELEASE_GUIDE.md
IanHoang: Good catch

gkamat (approved):

PR #655, opened 2024-09-24
add automatic testing and aggregation to OSB

### LINKED ISSUE #631: [Introduce Aggregate Command] Incorporate Automatic Aggregation
## Overview
This is an issue based off one of the proposed priorities in this RFC: #627

## Background
The proposed aggregate subcommand and enhanced compare subcommand will provide users with powerful tools for aggregating and comparing benchmark test results. However, there is an opportunity to leverage these features in other parts of OpenSearch Benchmark (OSB) to further enhance the user experience and provide more comprehensive performance testing capabilities.

## Proposed Design
We propose incorporating automatic aggregation into new and existing features of OSB, such as:

1. Automatically running a test multiple times and aggregating the results: OSB could include an option to automatically run a specified test multiple times and aggregate the results using the proposed aggregate subcommand. This would streamline the process of obtaining reliable and consistent performance measurements by reducing the need for manual intervention.

2. Aggregating results from distributed workload generation (DWG) tests across multiple load generation hosts: In DWG tests, where the workload is generated from multiple hosts, OSB could automatically aggregate the results from each host using the aggregate subcommand, providing a consolidated view of the overall performance under distributed load conditions.

The automatic aggregation feature could include options or configurations to handle failures or errors in individual test runs, such as excluding specific runs, using partial results up to the failure point, or retrying failed runs up to a certain limit.

## Proposed Priority
Incorporating automatic aggregation into new and existing features of OSB is a valuable enhancement that will improve the overall user experience and provide more comprehensive performance testing capabilities. It will streamline the process of obtaining reliable and consistent performance measurements, particularly in scenarios involving multiple test runs or distributed workload generation.

OVI3D0:
### Description
This change allows users to run benchmark tests with any configuration they choose, for `x` number of iterations, then generate an aggregated results at the end, all in one command!
Here's an example command:
`opensearch-benchmark auto-aggregate --test-iterations=2 --workload=geonames --target-hosts=127.0.0.1:9200`

The auto-aggregate argument parser inherits from the test-execution argument parser, so you can use any arguments/flags that you would use on a regular test execution:
`opensearch-benchmark auto-aggregate --test-iterations=2 --workload=geonames --target-hosts=127.0.0.1:9200 --test-mode --kill-running-processes`
And these will be applied to each test execution.

### Issues Resolved
#631 
---

### DIFF
--- a/osbenchmark/aggregator.py
+++ b/osbenchmark/aggregator.py
@@ -169,7 +169,9 @@ class Aggregator:
         test_procedure = loaded_workload.find_test_procedure_or_default(test_exe.test_procedure)
 
         test_execution = metrics.create_test_execution(self.config, loaded_workload, test_procedure, test_exe.workload_revision)
-        test_execution.user_tags = list(self.test_executions.keys())
+        test_execution.user_tags = {
+            "aggregation-of-runs": list(self.test_executions.keys())
+        }
         test_execution.add_results(AggregatedResults(aggregated_results))
         test_execution.distribution_version = test_exe.distribution_version
         test_execution.revision = test_exe.revision
--- a/osbenchmark/aggregator.py
+++ b/osbenchmark/aggregator.py
@@ -169,6 +169,7 @@ class Aggregator:
         test_procedure = loaded_workload.find_test_procedure_or_default(test_exe.test_procedure)
 
         test_execution = metrics.create_test_execution(self.config, loaded_workload, test_procedure, test_exe.workload_revision)
+        test_execution.user_tags = list(self.test_executions.keys())
         test_execution.add_results(AggregatedResults(aggregated_results))
         test_execution.distribution_version = test_exe.distribution_version
         test_execution.revision = test_exe.revision
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -631,7 +631,6 @@ def create_arg_parser():
         "--cancel-on-error",
         action="store_true",
         help="Stop executing tests if an error occurs in one of the test iterations (default: false).",
-        default=False
     )
 
     ###############################################################################
@@ -987,10 +986,10 @@ def dispatch_sub_command(arg_parser, args, cfg):
                 test_exes = []
                 for _ in range(iterations):
                     try:
-                        test_exes.append(args.test_execution_id)
                         configure_test(arg_parser, args, cfg)
                         execute_test(cfg, args.kill_running_processes)
                         time.sleep(int(args.sleep_timer))
+                        test_exes.append(args.test_execution_id)
                         args.test_execution_id = str(uuid.uuid4())
                     except Exception as e:
                         console.error(f"Error occurred during test execution {_+1}: {str(e)}")
@@ -1006,8 +1005,6 @@ def dispatch_sub_command(arg_parser, args, cfg):
             elif args.test_iterations == 1:
                 configure_test(arg_parser, args, cfg)
                 execute_test(cfg, args.kill_running_processes)
-            else:
-                console.info("Please enter a valid number of test iterations")
         elif sub_command == "create-workload":
             cfg.add(config.Scope.applicationOverride, "generator", "indices", args.indices)
             cfg.add(config.Scope.applicationOverride, "generator", "number_of_docs", args.number_of_docs)
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -28,7 +28,6 @@ import logging
 import os
 import platform
 import sys
-import threading
 import time
 import uuid
 
@@ -615,6 +614,25 @@ def create_arg_parser():
              f"high values favor the most common queries. "
              f"Ignored if randomization is off (default: {workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_ALPHA}).",
         default=workload.loader.QueryRandomizerWorkloadProcessor.DEFAULT_ALPHA)
+    test_execution_parser.add_argument(
+        "--test-iterations",
+        help="The number of times to run the workload (default: 1).",
+        default=1)
+    test_execution_parser.add_argument(
+        "--aggregate",
+        type=lambda x: (str(x).lower() in ['true', '1', 'yes', 'y']),
+        help="Aggregate the results of multiple test executions (default: true).",
+        default=True)
+    test_execution_parser.add_argument(
+        "--sleep-timer",
+        help="Sleep for the specified number of seconds before starting the next test execution (default: 5).",
+        default=5)
+    test_execution_parser.add_argument(
+        "--cancel-on-error",
+        action="store_true",
+        help="Stop executing tests if an error occurs in one of the test iterations (default: false).",
+        default=False
+    )
 
     ###############################################################################
     #
@@ -635,17 +653,7 @@ def create_arg_parser():
         action="store_true",
         default=False)
 
-    auto_aggregate_parser = subparsers.add_parser("auto-aggregate",
-                                                  add_help=False,
-                                                  help="Run multiple test executions with the same configuration and aggregate the results",
-                                                  parents=[test_execution_parser])
-    auto_aggregate_parser.add_argument(
-        "--test-iterations",
-        type=int,
-        required=True,
-        help="Number of test executions to run and aggregate")
-
-    for p in [list_parser, test_execution_parser, compare_parser, aggregate_parser, auto_aggregate_parser,
+    for p in [list_parser, test_execution_parser, compare_parser, aggregate_parser,
               download_parser, install_parser, start_parser, stop_parser, info_parser, create_workload_parser]:
         # This option is needed to support a separate configuration for the integration tests on the same machine
         p.add_argument(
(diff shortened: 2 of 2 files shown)


osbenchmark/benchmark.py
```diff
@@ -634,8 +635,18 @@ def add_workload_source(subparser):
...
-              start_parser, stop_parser, info_parser, create_workload_parser]:
+    auto_aggregate_parser = subparsers.add_parser("auto-aggregate",
+                                                  add_help=False,
+                                                  help="Run multiple test executions with the same configuration and aggregate the results",
+                                                  parents=[test_execution_parser])
+    auto_aggregate_parser.add_argument(
+        "--test-iterations",
+        type=int,
+        required=True,
+        help="Number of test executions to run and aggregate")
```
IanHoang, line 646: Could `auto-aggregate` be renamed to `execute-tests` and `auto_aggregate_parser` be renamed to `multiple_test_executions_parser`? Those names would be more suitable since the main intention behind this subcommand is to allow users to run multiple tests. 

But before renaming them, please consider the following:
If all the parameters between the test_execution_parser and auto_aggregate_parser are the same except `--test-iterations`, what are your thoughts on omitting a separate subcommand and just incorporating `--test-iterations` into test_execution_parser and have the multiple tests logic performed in `elif subcommand == "execute-test"`? 

There doesn't seem to be a large enough difference between running a single test and running multiple tests that justifies creating another subcommand. But let me know if there's another reason why a separate subcommand might be better.

osbenchmark/benchmark.py
```diff
@@ -863,6 +874,30 @@ def prepare_test_executions_dict(args, cfg):
...
+    test_exes = []
+    args.subcommand = "execute-test"
+    aggregate_id = args.test_execution_id
+
+    for _ in range(args.test_iterations):
+        console.info(f"Running test {_ + 1}...")
+        args.test_execution_id = str(uuid.uuid4()) # we reuse the same args for each test so need to refresh the id
+        test_exes.append(args.test_execution_id)
+        with semaphore:
+            dispatch_sub_command(arg_parser, args, cfg)
```
IanHoang, line 888: Is semaphore really needed? Wondering if there might be some conflicts with preexisting threading / Thespianpy within OSB. Might be overkilil. I think a simple for-loop should suffice?

IanHoang: We should avoid using `dispatch_sub_command` and manipulating the args in line 880.

osbenchmark/benchmark.py
```diff
@@ -863,6 +874,30 @@ def prepare_test_executions_dict(args, cfg):
...
+    console.info(f"Test executions: {', '.join(test_exes)}")
+    console.info("Aggregating results...")
+    aggregate_args = arg_parser.parse_args([
+    "aggregate",
+    f"--test-executions={','.join(test_exes)}",
+    f"--test-execution-id={aggregate_id}",
+    f"--results-file={args.results_file}",
+    f"--workload-repository={args.workload_repository}"
+    ])
+    dispatch_sub_command(arg_parser, aggregate_args, cfg)
```
IanHoang, line 899: I think we should avoid using `dispatch_sub_command` again and parsing args here. Let's sync offline about this.

IanHoang (commented):
### Some considerations for running multiple tests
- When users run multiple tests, we should ensure that they automatically have `--kill-runninng-processes` by default. If it isn't present in the parsed args, we need to add it by default, or else, we'll run into a wall of errors for every test after the first successful test because existing processes might be still alive. 
- I tested what happens when users cancel (ctrl + c) the test and they'll have to cancel the test for each iteration (e.g. if a user has 100 test iterations, they'll be spamming ctrl + c for some time 😆 ). Alternatively, they can open a separate window and kill all running processes. We should find a way to make it easier to cancel all tests with a single invocation of ctrl + c if possible
- When users run into errors across all tests, we should reduce the help text and errors. If we can't avoid that, we should include an additional parameter called "--cancel-upon-error" or something like that
- There are additional parameters that I think would be useful for running multiple tests:
    - `--time-between-tests` / `--sleep-time` (you can improve the name) = this would allow users to specify how much time there should be between tests. It's a nice-to-have as I've seen some users incorporate sleeps between OSB runs because they want to change something on the host's end between tests.
    - `--disable-aggregate` = by default we should have this set to False. But some users might not care about aggregating the final results. We should give them the option to disable it. 

### Some considerations for aggregating results after all tests
- Can we add the relative standard deviation (RSD) in aggregate feature? This would be useful so that users can see the spread of their data and how much each test deviates from one another 
- Maybe in user tag it would be include to runs and their ids in the user tags by default? That way users have a way to track down which test execution ids relate to the aggregated one once the console is wiped.

IanHoang:
@OVI3D0 Overall, great work. This feature has been long overdue in OSB. Let's sync offline if you have any questions.

osbenchmark/benchmark.py
IanHoang: Semaphore might be impacting the performance of each run. I ran a command that sets clients and target throughput to 10. This should ensure that final results have throughput of 10 ops/s. When I ran the command with these `auto-aggregate`, I see that target throughput is not being achieved across all runs and the aggregated results.

Command: `$ opensearch-benchmark auto-aggregate --test-iterations=3 --workload=geonames --target-hosts=XXXXXXX --client-options=basic_auth_user:'XXXXX',basic_auth_password:'XXXXXXX' --include-tasks="term" --test-mode --kill-running-processes --workload-params="search_clients:10,target_throughput:10"`
```
# Each test has the following results
|                                                  Segment count |        |         374 |        |
|                                                 Min Throughput |   term |       33.13 |  ops/s |
|                                                Mean Throughput |   term |       35.22 |  ops/s |
|                                              Median Throughput |   term |       35.69 |  ops/s |
|                                                 Max Throughput |   term |       36.39 |  ops/s |
```

However, when I run the command with `execute-test`, it reaches the target throughput.

Command: `$ opensearch-benchmark execute-test --workload=geonames --target-hosts=XXXXXXX --client-options=basic_auth_user:'XXXXXXX',basic_auth_password:'XXXXXXX--include-tasks="term" --kill-running-processes --workload-params="search_clients:10,target_throughput:10"`

```
|                                                  Segment count |        |         374 |        |
|                                                 Min Throughput |   term |        9.99 |  ops/s |
|                                                Mean Throughput |   term |          10 |  ops/s |
|                                              Median Throughput |   term |          10 |  ops/s |
|                                                 Max Throughput |   term |          10 |  ops/s |
```

Both commands were run on the same machine. We should reconsider using semaphore and go for a simpler approach. We can discuss this offline.

OVI3D0: > Is semaphore really needed? Wondering if there might be some conflicts with preexisting threading / Thespianpy within OSB. Might be overkilil. I think a simple for-loop should suffice?

I tried a simple for loop but i was having issues with already running processes 😕

OVI3D0:
> * I tested what happens when users cancel (ctrl + c) the test and they'll have to cancel the test for each iteration (e.g. if a user has 100 test iterations, they'll be spamming ctrl + c for some time 😆 ). Alternatively, they can open a separate window and kill all running processes. We should find a way to make it easier to cancel all tests with a single invocation of ctrl + c if possible

This is fixed!

> * When users run into errors across all tests, we should reduce the help text and errors. If we can't avoid that, we should include an additional parameter called "--cancel-upon-error" or something like that

Good idea! added this.

>   * `--time-between-tests` / `--sleep-time` (you can improve the name) = this would allow users to specify how much time there should be between tests. It's a nice-to-have as I've seen some users incorporate sleeps between OSB runs because they want to change something on the host's end between tests.

Added this as well!

>   * `--disable-aggregate` = by default we should have this set to False. But some users might not care about aggregating the final results. We should give them the option to disable it.

I instead added `--aggregate` and defaulted it to True. But if you think it makes more sense the other way around then I can fix this

> ### Some considerations for aggregating results after all tests
> * Can we add the relative standard deviation (RSD) in aggregate feature? This would be useful so that users can see the spread of their data and how much each test deviates from one another

I think this would be a great idea for a future PR

> * Maybe in user tag it would be include to runs and their ids in the user tags by default? That way users have a way to track down which test execution ids relate to the aggregated one once the console is wiped.

Yeah great idea, I added individual test ID's to the user tags of each aggregated test result.

osbenchmark/benchmark.py
OVI3D0: Discussed offline and added in the revision 👍

osbenchmark/benchmark.py
OVI3D0: Discussed more offline and used an alternate approach. I tried a few test runs on the same machine with and without the changes, and the results were very similar. Let me know if you find any differences

osbenchmark/aggregator.py
```diff
@@ -169,6 +169,7 @@ def build_aggregated_results(self):
         test_procedure = loaded_workload.find_test_procedure_or_default(test_exe.test_procedure)
 
         test_execution = metrics.create_test_execution(self.config, loaded_workload, test_procedure, test_exe.workload_revision)
+        test_execution.user_tags = list(self.test_executions.keys())
```
IanHoang, line 172: Alternatively, we could assign the user_tag to a k,v pair something like
```
 test_execution.user_tags = {
    "aggregation-of-runs": list(self.test_executions.keys())
}
```

IanHoang: Would help users who are trying to filtering for specific aggregated files in the MDS

osbenchmark/benchmark.py
```diff
@@ -614,6 +614,25 @@ def add_workload_source(subparser):
...
+        default=True)
+    test_execution_parser.add_argument(
+        "--sleep-timer",
+        help="Sleep for the specified number of seconds before starting the next test execution (default: 5).",
+        default=5)
+    test_execution_parser.add_argument(
+        "--cancel-on-error",
+        action="store_true",
+        help="Stop executing tests if an error occurs in one of the test iterations (default: false).",
+        default=False
```
IanHoang, line 634: No need to explicitly set `default=False` when using `store_true` or `store_false` actions in argparse. The default value is automatically set to False for `store_true` and True for `store_false`.

IanHoang (changes requested):

osbenchmark/benchmark.py
```diff
@@ -920,49 +982,32 @@ def dispatch_sub_command(arg_parser, args, cfg):
...
+                if args.aggregate:
+                    args.test_executions = test_exes
+                    test_executions_dict = prepare_test_executions_dict(args, cfg)
+                    aggregator_instance = aggregator.Aggregator(cfg, test_executions_dict, args)
+                    aggregator_instance.aggregate()
+            elif args.test_iterations == 1:
+                configure_test(arg_parser, args, cfg)
+                execute_test(cfg, args.kill_running_processes)
+            else:
+                console.info("Please enter a valid number of test iterations")
```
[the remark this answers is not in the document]
OVI3D0: True, I removed this. Although users can potentially enter invalid values

osbenchmark/benchmark.py
```diff
@@ -920,49 +982,32 @@ def dispatch_sub_command(arg_parser, args, cfg):
...
+                        time.sleep(int(args.sleep_timer))
+                        args.test_execution_id = str(uuid.uuid4())
+                    except Exception as e:
+                        console.error(f"Error occurred during test execution {_+1}: {str(e)}")
+                        if args.cancel_on_error:
+                            console.info("Cancelling remaining test executions.")
+                            break
+
+                if args.aggregate:
+                    args.test_executions = test_exes
```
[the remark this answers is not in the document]
OVI3D0: Nice catch, I moved like 990 so it only adds the test ID after the execute_test is called.

osbenchmark/aggregator.py
OVI3D0: I didn't consider this, I fixed it to look like your suggestion above!

osbenchmark/benchmark.py
OVI3D0: Haha oops. Fixed now :)

osbenchmark/benchmark.py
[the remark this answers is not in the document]
IanHoang: Oops, you're right. Let's add it back

IanHoang (approved):
LGTM

PR #677, opened 2024-10-04
Add support for multi-part data corpora downloads

### LINKED ISSUE #543: Permit large data corpora to be downloaded from the workloads repository
### Description
CloudFront limits restrict the size of data corpus files that can be downloaded to 50 GB.  Thus, in the case of the `big5` workload, for example, a corpus with an uncompressed size of 1 TB is currently not feasible, since the corresponding compressed size exceeds this limit.

It may be necessary to implement multipart downloads to work around this problem.  Other alternatives could also be investigated.

gkamat:
### Description
Permits data corpus files to be downloaded in parts.  This is not intended for performance, but rather, to work around the restriction on file size that services like CloudFront might have.

### Issues Resolved
#543 

### Testing
Lint, unit and integ tests.  Added a unit test to exercise the feature.

---

### DIFF
--- a/osbenchmark/resources/workload-schema.json
+++ b/osbenchmark/resources/workload-schema.json
@@ -406,9 +406,33 @@
                   "format": "uri",
                   "description": "The root URL for these documents."
                 },
+                "source-url": {
+                  "type": "string",
+                  "format": "uri",
+                  "description": "The full URL to the document file.  This is intended for cases like when a signed URL needs to be used to access the file directly."
+                },
                 "source-file": {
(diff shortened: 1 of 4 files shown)


IanHoang (approved):
This looks good. Should we open an issue in the workloads repository for splitting the 1 TB file in Big5 workload?

gkamat:
> This looks good. Should we open an issue in the workloads repository for splitting the 1 TB file in Big5 workload?

Will just make the change directly.

PR #683, opened 2024-10-16
store aggregated results in separate folder

### LINKED ISSUE #661: [FEATURE] Store Aggregated test results in a separate local folder

Currently, aggregated test execution results are stored in the same local folder as normal test execution results. It would be much cleaner if aggregated results were stored in their own folder, like `~/.benchmark/benchmarks/aggregations`

Store aggregations in a separate local folder, such as `~/.benchmark/benchmarks/aggregations`, and ensure the `compare` command knows to also search in this separate directory in the event that users want to compare aggregated results.

OVI3D0:
### Description
Changes OSB's file storage so that aggregated test results are stored in their own folder, `aggregated_results` alongside the `test_executions` folder.

### Issues Resolved
#661 

### Testing

`make it` + `make test`

---


osbenchmark/metrics.py
```diff
@@ -1550,6 +1550,14 @@ def store_test_execution(self, test_execution):
         with open(self._test_execution_file(), mode="wt", encoding="utf-8") as f:
             f.write(json.dumps(doc, indent=True, ensure_ascii=False))
 
+    def store_aggregated_execution(self, test_execution):
+        doc = test_execution.as_dict()
+        aggregated_execution_path = paths.aggregated_results_root(self.cfg, test_execution_id=test_execution.test_execution_id)
+        io.ensure_dir(aggregated_execution_path)
+        aggregated_file = os.path.join(aggregated_execution_path, "test_execution.json")
```
gkamat, line 1557: May be better to call this `aggregated_test_execution.json`.

gkamat (approved):

osbenchmark/metrics.py
IanHoang: +1

OVI3D0: @gkamat @IanHoang done + I added some changes to metrics.py so OSB can locate the new aggregate folder if you guys wouldn't mind taking a look👍

osbenchmark/metrics.py
```diff
@@ -1315,6 +1315,45 @@ def format_dict(d):
...
+                "Workload Parameters",
+                "TestProcedure",
+                "ProvisionConfigInstance",
+                "User Tags",
+                "workload Revision",
+                "Provision Config Revision"
+                ]))
+    else:
+        console.println("")
+        console.println("No recent aggregate tests found.")
```
IanHoang, line 1356: This works but a lot of this is duplicated from `list_test_executions()`. How do you feel about creating a separate helper method and just passing in the methods like the following pseudo code"

```
def list_test_executions(cfg):
    list_test_helper(test_execution_store(cfg).list_aggregations(), "test executions")

def list_aggregated_test_results(cfg):
    list_test_helper(test_execution_store(cfg).list_aggregations(), "aggregation tests")

def list_test_helper(test_store_method, test_execution_type):
    def format_dict(d):
        if d:
            items = sorted(d.items())
            return ", ".join(["%s=%s" % (k, v) for k, v in items])
        else:
            return None
    
    test_executions = []
    listed_test_executions = test_store_method
    for test_execution in listed_test_executions:
        ...
    console.println(f"No recent {test_execution_type} found.")
     ...
```
By putting boiler plate code in `list_test_helper()`, we will avoid repeating code and slim down both `list_test_executions(cfg)` and `list_aggregated_test_results(cfg)`.

osbenchmark/benchmark.py
```diff
@@ -697,6 +698,8 @@ def dispatch_list(cfg):
         test_execution_orchestrator.list_pipelines()
     elif what == "test_executions":
         metrics.list_test_executions(cfg)
+    elif what == "aggregated_results":
```
IanHoang, line 701: Since you've introduced a new method called `list_aggregations()`, how do you feel about using `aggregations` here as opposed to `aggregated_results`? Aggregations would be more simple for users to type out. But on second thought, aggregations might be confused with OpenSearch aggregations. I'm open to either

IanHoang (changes requested):
Left a couple suggestions

osbenchmark/metrics.py
OVI3D0: Good point! Did this

osbenchmark/benchmark.py
OVI3D0: Personally, I think `aggregated_results` is more explicit and clear about what it's referring to. But we could make it so either `aggregations` or `aggregated_results` triggers the same function maybe?

IanHoang: `aggregated_results` is more explicit and we can go forth with that

osbenchmark/metrics.py
```diff
@@ -1275,7 +1275,7 @@ def results_store(cfg):
         return NoopResultsStore()
 
 
-def list_test_executions(cfg):
+def list_test_helper(store, title):
```
IanHoang, line 1278: Nit: `store_item` might be more apt.

IanHoang (approved):

osbenchmark/metrics.py
OVI3D0: Done👍

PR #687, opened 2024-10-29
change min/max to overall_min/overall_max + update comparison results publisher

### LINKED ISSUE #684: [FEATURE] Add median values to aggregated test results and adjust min/max values
Currently, the new `aggregate` command in OSB combines multiple test executions into one averaged result. Rather than just combining all the minimum and maximum values from each test execution for an average 'min' and 'max' value, it might be more useful to provide the minimum and maximum across all test executions. In addition, another measurement like the median value across all test executions could also provide more insight to the user.

OVI3D0:
### Description
This change updates min/max values in aggregated results to reflect the min/max values across all test executions, rather than just an average of the min/max values. I also changed the names from `min` and `max` to `overall_min` and `overall_max` for clarity. Changes were also made to the comparison results publisher class, so if an aggregated result with these new metrics were to be used in a comparison, these new 'overall' min and max values can still be compared without issue.

### Issues Resolved
#684 

### Testing

`make test`

---

### DIFF
--- a/osbenchmark/aggregator.py
+++ b/osbenchmark/aggregator.py
@@ -202,15 +202,22 @@ class Aggregator:
                         weighted_metrics[metric][item_key] = values[0][item_key]
                     else:
                         item_values = [value.get(item_key, 0) for value in values]
-                        weighted_sum = sum(value * iterations for value in item_values)
-                        total_iterations = iterations * len(item_values)
-                        weighted_avg = weighted_sum / total_iterations
-                        weighted_metrics[metric][item_key] = weighted_avg
+                        if item_key == 'min':
+                            weighted_metrics[metric]['overall_min'] = min(item_values)
+                        elif item_key == 'max':
+                            weighted_metrics[metric]['overall_max'] = max(item_values)
+                        elif item_key == 'median':
+                            weighted_sum = sum(value * iterations for value in item_values)
+                            total_iterations = iterations * len(item_values)
+                            weighted_metrics[metric][item_key] = weighted_sum / total_iterations
+                        else:
+                            weighted_sum = sum(value * iterations for value in item_values)
+                            total_iterations = iterations * len(item_values)
+                            weighted_metrics[metric][item_key] = weighted_sum / total_iterations
             else:
                 weighted_sum = sum(value * iterations for value in values)
                 total_iterations = iterations * len(values)
-                weighted_avg = weighted_sum / total_iterations
-                weighted_metrics[metric] = weighted_avg
+                weighted_metrics[metric] = weighted_sum / total_iterations
 
         return weighted_metrics
 
--- a/osbenchmark/aggregator.py
+++ b/osbenchmark/aggregator.py
@@ -193,6 +193,8 @@ class Aggregator:
 
     def calculate_weighted_average(self, task_metrics: Dict[str, List[Any]], iterations: int) -> Dict[str, Any]:
         weighted_metrics = {}
+        num_executions = len(next(iter(task_metrics.values())))
+        total_iterations = iterations * num_executions
 
         for metric, values in task_metrics.items():
             if isinstance(values[0], dict):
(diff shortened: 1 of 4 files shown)


osbenchmark/aggregator.py
```diff
@@ -202,15 +202,22 @@ def calculate_weighted_average(self, task_metrics: Dict[str, List[Any]], iterati
...
-                        total_iterations = iterations * len(item_values)
-                        weighted_avg = weighted_sum / total_iterations
-                        weighted_metrics[metric][item_key] = weighted_avg
+                        if item_key == 'min':
+                            weighted_metrics[metric]['overall_min'] = min(item_values)
+                        elif item_key == 'max':
+                            weighted_metrics[metric]['overall_max'] = max(item_values)
+                        elif item_key == 'median':
+                            weighted_sum = sum(value * iterations for value in item_values)
+                            total_iterations = iterations * len(item_values)
```
gkamat, line 211: Isn't this already available from above and does not need to be re-computed?

osbenchmark/aggregator.py
```diff
@@ -202,15 +202,22 @@ def calculate_weighted_average(self, task_metrics: Dict[str, List[Any]], iterati
...
+                            weighted_metrics[metric][item_key] = weighted_sum / total_iterations
+                        else:
+                            weighted_sum = sum(value * iterations for value in item_values)
+                            total_iterations = iterations * len(item_values)
+                            weighted_metrics[metric][item_key] = weighted_sum / total_iterations
```
gkamat, line 216: The `elif` is probably not needed here, since the code in the two arms is the same?

osbenchmark/results_publisher.py
```diff
@@ -464,16 +464,16 @@ def _write_results(self, metrics_table, metrics_table_console):
                             data_plain=metrics_table, data_rich=metrics_table_console)
 
     def _publish_throughput(self, baseline_stats, contender_stats, task):
-        b_min = baseline_stats.metrics(task)["throughput"]["min"]
+        b_min = baseline_stats.metrics(task)["throughput"].get("overall_min") or baseline_stats.metrics(task)["throughput"]["min"]
```
gkamat, line 467: It would probably be better to set the `overall_min` key prior, so dealing with a special case won't be necessary.

gkamat (changes requested):
Please consider adding (or updating) a test as well.

osbenchmark/aggregator.py
IanHoang: +1 we can just have a comment in the else statement right after `else` stating what cases this for
```
else:
    # for items like median
```

osbenchmark/results_publisher.py
```diff
@@ -464,16 +464,16 @@ def _write_results(self, metrics_table, metrics_table_console):
...
         b_mean = baseline_stats.metrics(task)["throughput"]["mean"]
         b_median = baseline_stats.metrics(task)["throughput"]["median"]
-        b_max = baseline_stats.metrics(task)["throughput"]["max"]
+        b_max = baseline_stats.metrics(task)["throughput"].get("overall_max") or baseline_stats.metrics(task)["throughput"]["max"]
         b_unit = baseline_stats.metrics(task)["throughput"]["unit"]
 
-        c_min = contender_stats.metrics(task)["throughput"]["min"]
+        c_min = contender_stats.metrics(task)["throughput"].get("overall_min") or contender_stats.metrics(task)["throughput"]["min"]
```
IanHoang, line 473: This change was made in the ComparisonResultsPublisher but should they also be made in the SummaryResultsPublisher?

IanHoang (commented):
Could you add tests to results_publisher to confirm the changes work?

IanHoang (changes requested):
Left a few comments

osbenchmark/results_publisher.py
OVI3D0: I think this logic is necessary since the results publisher is also used for normal test executions which will not contain 'overall' min/max values

osbenchmark/results_publisher.py
OVI3D0: I don't think so, since the summary results publisher isn't used by the aggregator class. Although if we wanted to publish results after aggregation then this would be necessary

osbenchmark/aggregator.py
OVI3D0: True, I went ahead and moved this calculation outside of the loop.

osbenchmark/aggregator.py
OVI3D0: Got it, I added a comment to clarify what cases this is for

OVI3D0:
> Could you add tests to results_publisher to confirm the changes work?

I added a test for this, but let me know if I should add more!

I also updated the unit tests for the aggregator class.

PR #690, opened 2024-11-06
Fix workload parameter processing when a key is mapped to a JSON file

### LINKED ISSUE #502: [BUG] --workload-params="index_body:<filename>.json" is not working
**Describe the bug**
I'm running the opensearch benchmark test with customized mapping file, which is stored under the same workload folder: 

`opensearch-benchmark execute-test --pipeline=benchmark-only --workload-path=/home/ec2-user/opensearch-benchmark/xxx/xxx--target-hosts=https://xxxxxxx.us-east-1.es.amazonaws.com/ --client-options=timeout:300,use_ssl:true,verify_certs:false,basic_auth_user:'xxx',basic_auth_password:'xxxx' --kill-running-processes --workload-params="index_body:xxx.json" `

And then I got this error message:
`[ERROR] Cannot execute-test. [Errno 2] No such file or directory: './index_body:xxx.json'.`

**More Context (please complete the following information):**
- OpenSearch Version: 2.11

gkamat:
### Description
When the argument passed to the `workload-params` option includes a key-value pair, where the value is a JSON file, there parser may treat the entire option string incorrectly as a filename. 

### Issues Resolved
#502 

### Testing
Unit and integ tests.  Added unit tests.

---

### DIFF
--- a/osbenchmark/workload/loader.py
+++ b/osbenchmark/workload/loader.py
@@ -1278,7 +1278,11 @@ class WorkloadFileReader:
                         ve.instance, indent=4, sort_keys=True),
                         ve.absolute_path, ve.absolute_schema_path))
 
-        current_workload = self.read_workload(workload_name, workload_spec, mapping_dir)
+        try:
+            current_workload = self.read_workload(workload_name, workload_spec, mapping_dir)
+        except Exception as e:
+            console.error(e)
+            raise
 
         unused_user_defined_workload_params = self.complete_workload_params.unused_user_defined_workload_params()
         if len(unused_user_defined_workload_params) > 0:
--- a/osbenchmark/utils/opts.py
+++ b/osbenchmark/utils/opts.py
@@ -84,7 +84,7 @@ def kv_to_map(kvs):
 
 
 def to_dict(arg, default_parser=kv_to_map):
-    if io.has_extension(arg, ".json"):
+    if io.has_extension(arg, ".json") and ',' not in arg and ':' not in arg:
         with open(io.normalize_path(arg), mode="rt", encoding="utf-8") as f:
             return json.load(f)
     elif arg.startswith("{"):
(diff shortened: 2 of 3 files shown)


osbenchmark/workload/loader.py
```diff
@@ -1278,7 +1278,11 @@ def read(self, workload_name, workload_spec_file, mapping_dir):
                         ve.instance, indent=4, sort_keys=True),
                         ve.absolute_path, ve.absolute_schema_path))
 
-        current_workload = self.read_workload(workload_name, workload_spec, mapping_dir)
+        try:
+            current_workload = self.read_workload(workload_name, workload_spec, mapping_dir)
+        except Exception as e:
```
IanHoang, line 1283: Nit: We can use `WorkloadConfigError` from `exceptions.py` in place of the generic `Exception` class to be more descriptive. 
```
# from exceptions.py
class WorkloadConfigError(BenchmarkError):
    """
    Thrown when something is wrong with the workload config e.g. user supplied a workload-param
    that can't be set
    """

```

IanHoang (approved):
Left one comment but looks good to me

osbenchmark/workload/loader.py
gkamat: > Nit: We can use `WorkloadConfigError` from `exceptions.py` in place of the generic `Exception` class to be more descriptive.
> 
> ```
> # from exceptions.py
> class WorkloadConfigError(BenchmarkError):
>     """
>     Thrown when something is wrong with the workload config e.g. user supplied a workload-param
>     that can't be set
>     """
> ```

The reason `Exception` was used here is to catch all possible errors that might be raised by `read_workload`.  For, instance, it might raise a `WorkloadSyntaxError`.  However, it's correct that a wrapped `WorkloadConfigError` could be raised once the exception is caught.  Will consider doing that in a future revision.

PR #692, opened 2024-11-13
Change min/max to overall_min/overall_max + update comparison results publisher

### LINKED ISSUE #684: [FEATURE] Add median values to aggregated test results and adjust min/max values

OVI3D0:
### Description
This change updates min/max values in aggregated test results, to reflect the overall min/max values across all test executions, rather than just the average of each. The names of these values have also been prefixed with `overall_` in order to reflect these changes. Changes were also made to the `ComparisonResultsPublisher` class, so these values can still be read if used in a `compare` command. Tests were also updated to reflect the recent changes to the aggregator class, and also to test the `ComparisonResultsPublisher` class.

### Issues Resolved
#684 

### Testing

`make test`

---

### DIFF
--- a/osbenchmark/aggregator.py
+++ b/osbenchmark/aggregator.py
@@ -202,15 +202,22 @@ class Aggregator:
                         weighted_metrics[metric][item_key] = values[0][item_key]
                     else:
                         item_values = [value.get(item_key, 0) for value in values]
-                        weighted_sum = sum(value * iterations for value in item_values)
-                        total_iterations = iterations * len(item_values)
-                        weighted_avg = weighted_sum / total_iterations
-                        weighted_metrics[metric][item_key] = weighted_avg
+                        if item_key == 'min':
+                            weighted_metrics[metric]['overall_min'] = min(item_values)
+                        elif item_key == 'max':
+                            weighted_metrics[metric]['overall_max'] = max(item_values)
+                        elif item_key == 'median':
+                            weighted_sum = sum(value * iterations for value in item_values)
+                            total_iterations = iterations * len(item_values)
+                            weighted_metrics[metric][item_key] = weighted_sum / total_iterations
+                        else:
+                            weighted_sum = sum(value * iterations for value in item_values)
+                            total_iterations = iterations * len(item_values)
+                            weighted_metrics[metric][item_key] = weighted_sum / total_iterations
             else:
                 weighted_sum = sum(value * iterations for value in values)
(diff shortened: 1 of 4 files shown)


osbenchmark/aggregator.py
```diff
@@ -193,24 +193,27 @@ def build_aggregated_results(self):
 
     def calculate_weighted_average(self, task_metrics: Dict[str, List[Any]], iterations: int) -> Dict[str, Any]:
         weighted_metrics = {}
+        num_executions = len(next(iter(task_metrics.values())))
+        total_iterations = iterations * num_executions
```
IanHoang, line 197: To clarify, `iterations` is the number of iterations that the user inputted in workload params / default number of times a task is run in a workload? Not the number of times we executed the same test?

osbenchmark/aggregator.py
```diff
@@ -193,24 +193,27 @@ def build_aggregated_results(self):
...
 
         for metric, values in task_metrics.items():
             if isinstance(values[0], dict):
                 weighted_metrics[metric] = {}
                 for item_key in values[0].keys():
                     if item_key == 'unit':
                         weighted_metrics[metric][item_key] = values[0][item_key]
+                    elif item_key == 'min':
```
IanHoang, line 205: Nit: We can improve the readability of this section by renaming `item_key` to something like `metric_field`

osbenchmark/aggregator.py
```diff
@@ -193,24 +193,27 @@ def build_aggregated_results(self):
...
+                        weighted_metrics[metric]['overall_min'] = min(value.get(item_key, 0) for value in values)
+                    elif item_key == 'max':
+                        weighted_metrics[metric]['overall_max'] = max(value.get(item_key, 0) for value in values)
                     else:
+                        # for items like median or percentile values
```
IanHoang, line 210: For fields like `median` or containing percentile values

IanHoang (approved):
LGTM but suggested some quick fixes

osbenchmark/aggregator.py
OVI3D0: That's correct!

PR #696, opened 2024-11-19
Add support for custom iteration values

### LINKED ISSUE #670: [FEATURE] Support custom iteration values in new `aggregate` command

Currently, the new `aggregate` command calculates weighted averages for metrics using the number of iterations for each operation. However, some users may want to run their workloads with custom iteration values. We should update the `aggregate` command so it can support cases where users want to run operations with custom values.

First, the workloads should be updated to support [configurable iteration values](https://github.com/opensearch-project/opensearch-benchmark-workloads/issues/393#issue-2537111431). This way, there are named values that can be accessed to check if users ran a workload with custom iteration values. The aggregator class should then perform a check for these values and use them when calculating the weighted averages for each value.

_No response_

### Additional context

_No response_

OVI3D0:
### Description
Recently, OSB workloads were updated to include parameters for users to input custom iteration values in their test executions. This change gives the new aggregate command the ability to read these values and take them into account when calculating the average values for test executions.

Previously, iterations for each task were stored as a sum of iterations across all test execution. With this change, the iterations for each task are stored separately, so a test with x times the iterations of another will also hold that much more weight in the average calculation.

Also updates the unit tests to reflect the new changes.

### Issues Resolved
#670 

### Testing

`make test`

---

### DIFF
--- a/osbenchmark/aggregator.py
+++ b/osbenchmark/aggregator.py
@@ -127,7 +127,6 @@ class Aggregator:
         }
 
         for task, task_metrics in self.accumulated_results.items():
-            iterations = self.accumulated_iterations.get(task, 1)
             aggregated_task_metrics = self.calculate_weighted_average(task_metrics, task)
             op_metric = {
                 "task": task,
@@ -201,10 +200,9 @@ class Aggregator:
 
     def calculate_weighted_average(self, task_metrics: Dict[str, List[Any]], task_name: str) -> Dict[str, Any]:
         weighted_metrics = {}
-        num_executions = len(next(iter(task_metrics.values())))
-        
+
         # Get iterations for each test execution
-        iterations_per_execution = [self.accumulated_iterations[test_id][task_name] 
+        iterations_per_execution = [self.accumulated_iterations[test_id][task_name]
                                     for test_id in self.test_executions.keys()]
         total_iterations = sum(iterations_per_execution)
 
--- a/osbenchmark/aggregator.py
+++ b/osbenchmark/aggregator.py
@@ -17,32 +17,28 @@ class Aggregator:
         self.metrics = ["throughput", "latency", "service_time", "client_processing_time", "processing_time", "error_rate", "duration"]
         self.test_store = metrics.test_execution_store(self.config)
         self.cwd = cfg.opts("node", "benchmark.cwd")
+        self.test_execution = self.test_store.find_by_test_execution_id(list(self.test_executions.keys())[0])
+        self.test_procedure_name = self.test_execution.test_procedure
+        self.loaded_workload = None
 
     def count_iterations_for_each_op(self, test_execution) -> None:
-        loaded_workload = workload.load_workload(self.config)
-        test_procedure_name = self.config.opts("workload", "test_procedure.name")
-        test_procedure_found = False
-        workload_params = {} if getattr(test_execution, 'workload_params') is None else test_execution.workload_params
+        matching_test_procedure = next((tp for tp in self.loaded_workload.test_procedures if tp.name == self.test_procedure_name), None)
+        workload_params = getattr(test_execution, 'workload_params', {})
 
         test_execution_id = test_execution.test_execution_id
         self.accumulated_iterations[test_execution_id] = {}
 
-        for test_procedure in loaded_workload.test_procedures:
-            if test_procedure.name == test_procedure_name:
-                test_procedure_found = True
-                for task in test_procedure.schedule:
-                    task_name = task.name
-                    custom_key = f"{task_name}_iterations"
-                    if custom_key in workload_params:
-                        iterations = int(workload_params[custom_key])
-                    else:
-                        iterations = task.iterations or 1
-                    self.accumulated_iterations[test_execution_id][task_name] = iterations
-            else:
-                continue  # skip to the next test procedure if the name doesn't match
-
-        if not test_procedure_found:
-            raise ValueError(f"Test procedure '{test_procedure_name}' not found in the loaded workload.")
+        if matching_test_procedure:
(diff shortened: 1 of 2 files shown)


osbenchmark/aggregator.py
```diff
@@ -13,23 +13,31 @@ def __init__(self, cfg, test_executions_dict, args):
...
-        self.accumulated_iterations: Dict[str, int] = {}
+        self.accumulated_iterations: Dict[str, Dict[str, int]] = {}
         self.metrics = ["throughput", "latency", "service_time", "client_processing_time", "processing_time", "error_rate", "duration"]
         self.test_store = metrics.test_execution_store(self.config)
         self.cwd = cfg.opts("node", "benchmark.cwd")
 
-    def count_iterations_for_each_op(self) -> None:
+    def count_iterations_for_each_op(self, test_execution) -> None:
         loaded_workload = workload.load_workload(self.config)
         test_procedure_name = self.config.opts("workload", "test_procedure.name")
```
IanHoang, line 23: Should we initialize this as an instance variable so that it can be used throughout the aggregator class? Seeing areas where the aggregator has to fetch the test procedure name.

osbenchmark/aggregator.py
```diff
@@ -13,23 +13,31 @@ def __init__(self, cfg, test_executions_dict, args):
...
         test_procedure_found = False
+        workload_params = {} if getattr(test_execution, 'workload_params') is None else test_execution.workload_params
+
+        test_execution_id = test_execution.test_execution_id
+        self.accumulated_iterations[test_execution_id] = {}
 
         for test_procedure in loaded_workload.test_procedures:
             if test_procedure.name == test_procedure_name:
                 test_procedure_found = True
```
IanHoang, line 32: Super nit: Since `loaded_workload.test_procedures` is a list, we can reduce these three lines to:
```
if test_procedure_name in loaded_workload.test_procedures:
# Count iterations in each operation
else:
# raise value error
```

osbenchmark/aggregator.py
```diff
@@ -13,23 +13,31 @@ def __init__(self, cfg, test_executions_dict, args):
...
         self.accumulated_results: Dict[str, Dict[str, List[Any]]] = {}
-        self.accumulated_iterations: Dict[str, int] = {}
+        self.accumulated_iterations: Dict[str, Dict[str, int]] = {}
         self.metrics = ["throughput", "latency", "service_time", "client_processing_time", "processing_time", "error_rate", "duration"]
         self.test_store = metrics.test_execution_store(self.config)
         self.cwd = cfg.opts("node", "benchmark.cwd")
 
-    def count_iterations_for_each_op(self) -> None:
+    def count_iterations_for_each_op(self, test_execution) -> None:
         loaded_workload = workload.load_workload(self.config)
```
IanHoang, line 22: Seeing if we can improve this method: 
- Notice that we are loading the workload each time we pass in a test execution. Could we load the workload once and just reuse it?
- See that this method has some steps that checks if the test procedure is valid. Might be better to move that to the designated method `test_execution_compatibility_check`

IanHoang (changes requested):
Left some suggestions. Lmk your thoughts?

osbenchmark/aggregator.py
OVI3D0: Yeah, and for clarity I renamed the test_procedure in the build_aggregated_results function since it's not just the string name of the test procedure, but a test_procedure object with other attributes that are needed there

osbenchmark/aggregator.py
OVI3D0: hmm I don't think so because loaded_workload.test_procedures returns a list of test_procedure objects. So I'd have to match test_procedure_name to the name attribute in each of these objects

OVI3D0: Although I could use python's `next()` to search for the first object with a `name` attribute that matches the test_procedure_name. Something like:
```
matching_test_procedure = next((tp for tp in loaded_workload.test_procedures if tp.name == self.test_procedure_name), None)
if matching_test_procedure:
# count iterations
else:
# raise value error
```

osbenchmark/aggregator.py
OVI3D0: To your first point: Good call, this should be notably faster for larger aggregations now that I just load the workload once.

Also removed that extra logic👍

osbenchmark/aggregator.py
```diff
@@ -13,28 +13,32 @@ def __init__(self, cfg, test_executions_dict, args):
...
-        loaded_workload = workload.load_workload(self.config)
-        test_procedure_name = self.config.opts("workload", "test_procedure.name")
-        test_procedure_found = False
-
-        for test_procedure in loaded_workload.test_procedures:
-            if test_procedure.name == test_procedure_name:
-                test_procedure_found = True
-                for task in test_procedure.schedule:
-                    task_name = task.name
+        self.test_execution = None
```
IanHoang, line 20: Rename this to be more specific and add clairty

osbenchmark/aggregator.py
```diff
@@ -13,28 +13,32 @@ def __init__(self, cfg, test_executions_dict, args):
...
+        self.test_procedure_name = None
```
IanHoang, line 21: Something more descriptive like `self.shared_test_procedure_across_all_executions` would help

osbenchmark/aggregator.py
```diff
@@ -13,28 +13,32 @@ def __init__(self, cfg, test_executions_dict, args):
...
+        self.loaded_workload = None
```
IanHoang, line 22: Adding type hints to this and the ones above will help.

IanHoang (approved):
Let's address refactoring comments in subsequent PR

Issue #711, opened 2024-12-12
Make query randomization more flexible


peteralfonsi:
### Is your feature request related to a problem? Please describe

Query randomization (https://github.com/opensearch-project/opensearch-benchmark/pull/455) lets OSB automatically randomize the bounds of a range query in a controllable way, which is useful for testing cache performance. But it's currently hardcoded to only work on `range` queries with parameters `gte`/`gt`, `lte`/`lt`, and optionally `format`. This isn't ideal. For example, range queries like this don't usually enter the query cache, so testing the query cache requires more flexibility. 

### Describe the solution you'd like

For each operation, the user should be able to define the query name they are looking to randomize, and also specify the names of the parameters that should be randomized. For example, on a GeoPoint field named "pickup_location", they could randomize the following query: 

```
"query": {
          "geo_bounding_box": {
            "pickup_location": {
              "top_left": [-74.27, 40.92],
              "bottom_right": [-73.68, 40.49]
            }
          }
        }
```
by specifying the query name `geo_bounding_box` and the parameter names `top_left` and `bottom_right`. This can be done by registering these values on an operation level in workload.py, like how standard value sources for the randomness are registered.

Divyaasm:
Hey @rishabh6788, @IanHoang Can you please take a look! Thanks

IanHoang:
@peteralfonsi Thanks for contributing the PR. Will take a look when we free up

PR #712, opened 2024-12-12
Make query randomization more flexible

peteralfonsi:
### Description
Controllable query randomization was added in https://github.com/opensearch-project/opensearch-benchmark/pull/455, and was limited to randomizing `range` queries with fields `gte`/`gt`, `lte`/`lt`, and optionally `format`. This is limiting since we might need to randomize other queries for some use cases, like benchmarking the query cache. 

This PR lets the user register different query names and parameter names to be randomized for each operation. This is done by adding `registry.register_query_randomization_info(...)` in workload.py. See an [example](https://github.com/peteralfonsi/opensearch-benchmark-workloads/blob/geopoint-nyc-taxis/nyc_taxis/workload.py#L86). 

In this example we have the operation: 
```
{
  "name": "bbox", 
  "operation-type": "search", 
  "index": "nyc_taxis",
  "body": { 
    "size": 0,
    "query": {
      "geo_bounding_box": {
        "pickup_location": {
          "top_left": [-74.27, 40.92],
          "bottom_right": [-73.68, 40.49]
        }
      }
    }
  }
}
```
And in workload.py we run `registry.register_query_randomization_info("bbox", "geo_bounding_box", [["top_left"], ["bottom_right"]], [])`. 

The first argument, `"bbox"`, is the operation name. The second argument, `"geo_bounding_box"`, is the query type name. 

The third argument is a list of lists: `[["top_left"], ["bottom_right"]]`.  Each entry in the outer list represents one parameter name that will be randomized. It's a list because we may have multiple different versions of the same name that represent roughly the same thing. For example, `"gte"` or `"gt"`. In this case there's just one option for each parameter name. At least one version of each parameter name must be present in the original query for it to be randomized. 

The last argument is a list of optional parameters. If an optional parameter is present in the random standard value source, it will be put into the randomized version of the query. If it's not in the source, it's ignored. There are no optional parameters in this example, but the typical use case would be `"format"` in a range query. 

If nothing is registered, it falls back to the default query randomization info object, which randomizes range queries as was done before this PR. This is equivalent to registering `registry.register_query_randomization_info(<operation_name>, "range", [["gte", "gt"], ["lte", "lt"]], ["format"])`. 

The dict returned by the random standard value source should match the parameter names you are trying to randomize. For example the standard value source for the above example is: 

```
def bounding_box_source(): 
    top_longitude = random.uniform(-74.27, -73.68)
    top_latitude = random.uniform(40.49, 40.92)

    bottom_longitude = random.uniform(top_longitude, -73.68)
    bottom_latitude = random.uniform(40.49, top_latitude)

    return { 
        "top_left":[top_longitude, top_latitude],
        "bottom_right":[bottom_longitude, bottom_latitude]
    }
```

### Issues Resolved
https://github.com/opensearch-project/opensearch-benchmark/issues/711

### Testing

Adds UTs. Also tested manually with a [workload](https://github.com/peteralfonsi/opensearch-benchmark-workloads/tree/geopoint-nyc-taxis) designed for testing the query cache, which uses a non-default QueryRandomizationInfo. 

---

### DIFF
--- a/osbenchmark/workload/loader.py
+++ b/osbenchmark/workload/loader.py
@@ -1099,15 +1099,10 @@ class QueryRandomizerWorkloadProcessor(WorkloadProcessor):
         # minus 1 for mapping [1, N] to [0, N-1] of list indices
         return self.zipf_cdf_inverse(random.random(), self.H_list) - 1
 
-    def get_randomized_values(self, input_workload, input_params,
+    def get_randomized_values(self, input_workload, input_params, target_keys_info,
                               get_standard_value=params.get_standard_value,
                               get_standard_value_source=params.get_standard_value_source, # Made these configurable for simpler unit tests
-                              target_keys_info=None, 
                               **kwargs):
-        
-        if target_keys_info is None: 
-            target_keys_info = self.DEFAULT_TARGET_KEYS_INFO
-
         # The queries as listed in operations/default.json don't have the index param,
         # unlike the custom ones you would specify in workload.py, so we have to add them ourselves
         if not "index" in input_params:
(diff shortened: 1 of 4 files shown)


IanHoang:
@peteralfonsi Can you address the lint failures?

IanHoang (commented):
Side note: It'd be good for our users to know that this feature exists within OSB. @peteralfonsi can you work with @Naarcha-AWS on adding this to the documentation? It can go in the Optimizing Benchmarks section

IanHoang (approved):
Thanks for the contribution. This makes query randomization more flexible and less tied to range.

PR #716, opened 2024-12-19
add client ramp up feature. 

### LINKED ISSUE #713: Support gradual ramp up of clients

Currently, after a warmup period, all query tasks are executed together at the same time. After the first execution, they respect their configured scheduler for subsequent executions.
In cases with a high number of tasks (e.g. executing using a parallel block) this absence of a ramp up, can produce an initial extremely high load and result in excessive utilization.
Due to this it becomes very difficult to do red-line testing of OpenSearch cluster and reliably tell what is breaking point of a particular cluster. 

The objective of this feature is to support gradual ramp of clients to allow users to ramp-up load gradually by specifying the task property ramp-up-time-period. If a non-zero value is specified, OSB will gradually add clients during that time period until the target client count as specified by clients is reached. This reduces the potential for initial load spikes when running with many concurrent clients. 

_No response_

### Additional context

_No response_

rishabh6788:
### Description
As mentioned in #713 this PR introduces a new task property, `ramp-up-time-period`. When provided, opensearch-benchmark will gradually scale up clients based on the time-period and number of clients provided. 

### Issues Resolved
#713 

### Testing

---


osbenchmark/resources/workload-schema.json
```diff
@@ -28,6 +28,11 @@
                 "type": "integer",
                 "minimum": 1
               },
+              "ramp-up-time-period": {
+                "type": "integer",
+                "minimum": 0,
+                "description": "Defines the time period in seconds to gradually increase the number of clients."
+              },
```
OVI3D0, line 35: Will the OSB workloads also need to be updated to accommodate this change?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -220,22 +223,27 @@ def __init__(self):
         self.cluster_details = None
 
     def receiveMsg_PoisonMessage(self, poisonmsg, sender):
-        self.logger.error("Main worker_coordinator received a fatal indication from load generator (%s). Shutting down.", poisonmsg.details)
+        self.logger.error(
+            "Main worker_coordinator received a fatal indication from load generator (%s). Shutting down.", poisonmsg.details)
```
IanHoang, line 227: A lot of these changes in worker_coordinator.py are new lines in or reformatting existing lines. Are these intentional or was this done with an automated tool? We should minimize changes that aren't related to the proposed feature

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -2009,7 +2148,7 @@ def requires_time_period_schedule(task, task_runner, params):
 
 
 class ScheduleHandle:
-    def __init__(self, task_name, sched, task_progress_control, runner, params):
+    def __init__(self, task_allocation, sched, task_progress_control, runner, params):
         """
         Creates a generator that will yield individual task invocations for the provided schedule.
```
IanHoang, line 2154: The docstring should also be updated to reflect the change from task_name to task_allocation
```
:param task_name: The name of the task for which the schedule is generated.
```

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1941,7 +2075,7 @@ def clients(self):
 
 # Runs a concrete schedule on one worker client
 # Needs to determine the runners and concrete iterations per client.
-def schedule_for(task, client_index, parameter_source):
+def schedule_for(task_allocation, parameter_source):
     """
     Calculates a client's schedule for a given task.
```
IanHoang, line 2081: Same as in ScheduleHandler, docstrings should be updated to replace task with task_allocation and remove client_index

rishabh6788:
> @rishabh6788 Have left a few comments but could we revise the PR to not include the additional formatting changes? Seems like a lot of it was introduced by a lint checker from the code editor. It's hard to determine all that has changed.

Yes, the test workflow was failing with lint errors. Not sure why this wasn't caught in previous PRs.

PR #721, opened 2025-01-02
update rename ~/.benchmark to ~/.osb and add symlink

OVI3D0:
### Description
Updates OSB's config directory to `~/.osb` instead of `~/.benchmark`. Also adds a symlink if the old path exists but the new one does not. 'benchmark' is an overloaded term, while using `~/.osb` reduces ambiguity and is more consistent with the project's name.

### Issues Resolved

### Testing
- [] New functionality includes testing

---

### DIFF
--- a/osbenchmark/paths.py
+++ b/osbenchmark/paths.py
@@ -22,27 +22,29 @@
 # specific language governing permissions and limitations
 # under the License.
 import os
-
+import sys
+from osbenchmark.utils.io import ensure_dir
 
 def benchmark_confdir():
     default_home = os.path.expanduser("~")
     old_path = os.path.join(default_home, ".benchmark")
     new_path = os.path.join(default_home, ".osb")
 
-    # Create .benchmark directory if it doesn't exist
-    if not os.path.exists(old_path):
-        os.makedirs(old_path, exist_ok=True)
-
-    # Create .osb directory if it doesn't exist
-    if not os.path.exists(new_path):
-        os.makedirs(new_path, exist_ok=True)
+    # ensure both directories exist
+    ensure_dir(old_path)
+    ensure_dir(new_path)
 
     # Create symlink from .osb to .benchmark if it doesn't exist
     if not os.path.islink(new_path):
         try:
             os.symlink(old_path, new_path, target_is_directory=True)
-        except OSError:
-            print(f"Warning: Failed to create symlink from {new_path} to {old_path}")
+        except OSError as e:
+            error_message = (
+                f"OSError: Failed to create symlink from {new_path} to {old_path}\n"
+                f"Error type: {type(e).__name__}\n"
+                f"Error message: {str(e)}\n"
+            )
+            print(error_message, file=sys.stderr)
 
     return os.path.join(os.getenv("BENCHMARK_HOME", default_home), ".osb")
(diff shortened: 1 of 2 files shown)


gkamat (approved):

osbenchmark/paths.py
```diff
@@ -26,7 +26,11 @@
 
 def benchmark_confdir():
     default_home = os.path.expanduser("~")
-    return os.path.join(os.getenv("BENCHMARK_HOME", default_home), ".benchmark")
+    old_path = os.path.join(default_home, ".benchmark")
+    new_path = os.path.join(default_home, ".osb")
+    if os.path.exists(old_path) and not os.path.exists(new_path):
+        os.symlink(old_path, new_path)
+    return os.path.join(os.getenv("BENCHMARK_HOME", default_home), ".osb")
```
gkamat, line 33: If the `.benchmark` directory does not exist, you will need to create it, so both are present in all cases.

gkamat (changes requested):

OVI3D0:
@gkamat could there be a permissions issue when creating the symlink in the integ tests?
`it/distribution_test.py::test_tar_distributions[os-it] Warning: Failed to create symlink from /home/runner/.osb to /home/runner/.benchmark`

IanHoang:
We'll need to change the direction of this PR so that it merges into a separate branch as this is somewhat a breaking change (despite the symlink, documentation has not been updated)?

osbenchmark/paths.py
```diff
@@ -26,7 +26,25 @@
 
 def benchmark_confdir():
     default_home = os.path.expanduser("~")
-    return os.path.join(os.getenv("BENCHMARK_HOME", default_home), ".benchmark")
+    old_path = os.path.join(default_home, ".benchmark")
+    new_path = os.path.join(default_home, ".osb")
+
+    # Create .benchmark directory if it doesn't exist
+    if not os.path.exists(old_path):
+        os.makedirs(old_path, exist_ok=True)
```
IanHoang, line 34: We can reuse the `ensure_dir()` from utils. 
```
def ensure_dir(directory, mode=0o777):
    """
    Ensure that the provided directory and all of its parent directories exist.
    This function is safe to execute on existing directories (no op).

    :param directory: The directory to create (if it does not exist).
    :param mode: The permission flags to use (if it does not exist).
    """
    if directory:
        os.makedirs(directory, mode, exist_ok=True)
```

IanHoang:
> @gkamat could there be a permissions issue when creating the symlink in the integ tests? `it/distribution_test.py::test_tar_distributions[os-it] Warning: Failed to create symlink from /home/runner/.osb to /home/runner/.benchmark`

@OVI3D0 can we modify the exception clause to see what the exact error is from OSError?

IanHoang:
@OVI3D0 FileExistsError is showing up when `os.symlink()` is run. Can you add `os.path.islink()` to check if there is an existing symlink?

OVI3D0:
> @OVI3D0 FileExistsError is showing up when `os.symlink()` is run. Can you add `os.path.islink()` to check if there is an existing symlink?

I added a function that should hopefully stop these errors, called ensure_symlink in the utils file

PR #725, opened 2025-01-09
Changes to support ramp-up feature

### LINKED ISSUE #713: Support gradual ramp up of clients

rishabh6788:
### Description
As mentioned in https://github.com/opensearch-project/opensearch-benchmark/issues/713 this PR introduces a new task property, ramp-up-time-period. When provided, opensearch-benchmark will gradually scale up clients based on the time-period and number of clients provided.  

### Issues Resolved
#713 

### Testing

---


OVI3D0 (approved):
LGTM, apart from the linter errors of course. I ran this locally and all the unit tests pass otherwise.

Also, should the documentation be updated to include this new ramp up time period property?

rishabh6788:
> LGTM, apart from the linter errors of course. I ran this locally and all the unit tests pass otherwise.
> 
> Also, should the documentation be updated to include this new ramp up time period property?

Thank you. Yes, I'm working on a doc that explains how to use this feature.

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1806,18 +1815,28 @@ def __repr__(self, *args, **kwargs):
...
+        return hash(self.task) ^ hash(self.global_client_index)
 
     def __eq__(self, other):
-        return isinstance(other, type(self)) and self.task == other.task and self.client_index_in_task == other.client_index_in_task
+        return isinstance(other, type(self)) and self.task == other.task and self.global_client_index == other.global_client_index
 
     def __repr__(self, *args, **kwargs):
-        return "TaskAllocation [%d/%d] for %s" % (self.client_index_in_task, self.task.clients, self.task)
+        return f"TaskAllocation [{self.client_index_in_task}/{self.task.clients}] for {self.task} " \
+               f"and [{self.global_client_index}/{self.total_clients}] in total"
```
IanHoang, line 1839: Hoping to understand this better: Each tests has a total number of clients and each task is allocated group of those clients. So `self.client_index_in_task` tracks which client in the allocated group is performing the task while `self.global_client_index` tracks which client out of all clients is currently performing this task?

IanHoang (approved):
Left one question but functionality looks good to me

osbenchmark/worker_coordinator/worker_coordinator.py
OVI3D0: That's pretty much how I understood it. So now TaskAllocation knows both the clients position within a task (`self.client_index_in_task`) as well as the clients position in the overall benchmark (`self.global_client_index`). Maybe @rishabh6788 can correct me if I'm wrong

rishabh6788: yes. Consider a case there we have 3 parallel tasks a, b and c, where we have declared total 3 clients at parallel task level and then have 1 client each for task a and b and then 2 clients for task c. 
In the task allocation matrix the 3 clients declared at parallel block level will represent global client indices and clients declared at each task level will be used as client index at each task level. It does get a bit confusing, let me know if you have any questions.

PR #732, opened 2025-01-16
Change OSB Config directory from `~/.benchmark` to `~/.osb`

### LINKED ISSUE #728: Rename config directory from ~/.benchmark to ~/.osb
Similar to #705, we should move away from `benchmark` which is an overloaded term which could be referring to various benchmarking tools. We should instead instead use `osb`. OSB is less ambiguous to general users, and doesn't need as much context to know what it is referring to.

OVI3D0:
### Description
Updates OSB's config directory to ~/.osb instead of ~/.benchmark. Also adds a symlink if the old path exists but the new one does not. 'benchmark' is an overloaded term, while using ~/.osb reduces ambiguity and is more consistent with the project's name.

### Issues Resolved
#728 

### Testing

`make test` + `make it`

---

### DIFF
--- a/osbenchmark/paths.py
+++ b/osbenchmark/paths.py
@@ -22,8 +22,8 @@
 # specific language governing permissions and limitations
 # under the License.
 import os
-import sys
 from osbenchmark.utils.io import ensure_dir, ensure_symlink
+from osbenchmark.utils import console
 
 def benchmark_confdir():
     default_home = os.path.expanduser("~")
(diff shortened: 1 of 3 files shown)


osbenchmark/paths.py
```diff
@@ -22,12 +22,39 @@
...
+        ensure_dir(old_path)
+
+        # Ensure symlink from .osb to .benchmark
+        ensure_symlink(old_path, new_path)
+
+        final_path = os.path.join(os.getenv("BENCHMARK_HOME", default_home), ".osb")
+
+        return final_path
 
+    except Exception as e:
```
IanHoang, line 44: Do you recall what the exact exception was? Let's try to use that one instead of a general Exception.

osbenchmark/paths.py
```diff
@@ -22,12 +22,39 @@
...
+    new_path = os.path.join(default_home, ".osb")
+
+    try:
+        # Ensure .benchmark directory exists
+        ensure_dir(old_path)
+
+        # Ensure symlink from .osb to .benchmark
+        ensure_symlink(old_path, new_path)
+
+        final_path = os.path.join(os.getenv("BENCHMARK_HOME", default_home), ".osb")
```
IanHoang, line 40: Nit: could be more descriptive with the variable name, would rename `final_path` to `benchmark_confdir_path`.

osbenchmark/paths.py
```diff
@@ -22,12 +22,39 @@
...
+            f"Error type: {type(e).__name__}\n"
+            f"Error message: {str(e)}\n"
+            f"Current user: {os.getlogin()}\n"
+            f"Current working directory: {os.getcwd()}\n"
+            f"Python version: {sys.version}\n"
+            f"Operating system: {sys.platform}\n"
+            f"Permissions of {old_path}: {oct(os.stat(old_path).st_mode) if os.path.exists(old_path) else 'N/A'}\n"
+            f"Permissions of parent of {new_path}: {oct(os.stat(os.path.dirname(new_path)).st_mode)}"
+        )
+        print(error_message)
```
IanHoang, line 56: Let's use console.print from utils?

IanHoang (changes requested):
Left some comments

osbenchmark/paths.py
OVI3D0: It ended up being a `FileNotFoundError`, but not sure if it's worth changing to this exception since the latest commit includes a fix for that error. Lmk what you think though

IanHoang (approved):

PR #737, opened 2025-01-23
update to JDK 21 for integ tests

OVI3D0:
### Description
Integration tests were failing due to recent changes in OpenSearch that require Java 21 for gradle builds:
```
* What went wrong:
	A problem occurred evaluating project ':buildSrc'.
	> At least Java 21 is required to build opensearch gradle tools
```
([reference to the failing tests](https://github.com/opensearch-project/opensearch-benchmark/actions/runs/12936541840/job/36084369757?pr=720#step:9:410)). 

[Reference to the OpenSearch PR that updates the JDK version](https://github.com/opensearch-project/OpenSearch/commit/7c46f8f14e1beefdd24eb2fe61792c6737fe9023#diff-bc7a42af0a917334e95bb43c1ac3408e14a03a45ef844e044089d4dc233d5b74R77-R78).  

To fix this, I added a JDK 21 installation to the integ tests for the gradle builds. However, a JDK version between 11 and 17 is still needed for the integration tests. So, I also modified the `build.sh` file to just use JDK 21 if necessary for the gradle operations, otherwise using JDK 17.

I'm open to alternatives if there is a more elegant solution.

### Testing

`make it`

---

### DIFF
--- a/.github/workflows/integ-test.yml
+++ b/.github/workflows/integ-test.yml
@@ -40,12 +40,31 @@ jobs:
       - name: Install pyenv
         run: git clone https://github.com/pyenv/pyenv.git ~/.pyenv
 
+      - name: Install JDK 21
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'adopt'
+          java-version: '21'
+      - run: |
+          echo "JAVA21_HOME=$JAVA_HOME" >> $GITHUB_ENV
+          echo "BUILD_JAVA_HOME=$JAVA_HOME" >> $GITHUB_ENV
+
       - name: Install JDK 17
         uses: actions/setup-java@v3
         with:
           distribution: 'adopt'
           java-version: '17'
-      - run: echo "JAVA17_HOME=$JAVA_HOME" >> $GITHUB_ENV
+      - run: |
+          echo "JAVA17_HOME=$JAVA_HOME" >> $GITHUB_ENV
+          echo "JAVA_HOME=$JAVA_HOME" >> $GITHUB_ENV  # Set JDK 17 as default
 
       - name: Run the CI build script
-        run: bash .ci/build.sh run_it ${{ matrix.python-version }}
+        run: |
+          # For build operations that need JDK 21
+          if [ -n "$BUILD_JAVA_HOME" ]; then
+            export ORIG_JAVA_HOME=$JAVA_HOME
+            export JAVA_HOME=$BUILD_JAVA_HOME
+            # do build operations here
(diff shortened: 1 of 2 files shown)


.github/workflows/integ-test.yml
```diff
@@ -45,7 +45,20 @@ jobs:
         with:
           distribution: 'adopt'
           java-version: '17'
-      - run: echo "JAVA17_HOME=$JAVA_HOME" >> $GITHUB_ENV
+      - run: |
```
IanHoang, line 48: We should document in a comment in both files why there are two JDK versions being used.

OVI3D0: Added some small comments

IanHoang (approved):

IanHoang:
Approved but will wait for @gkamat as he wanted to check this out

gkamat (changes requested):
This change may fix the integration tests as run via GitHub actions.  However, integ tests can be run directly when OSB is installed in development mode, with `make it`, and this change will not be applicable in that case.  Also, switching the JDK back and forth makes this a bit harder to understand.

The better option would probably be to add JDK 21 to the `provision_config` resources:
```
# major version of the JDK that is used to build OpenSearch
build.jdk = 17
# list of JDK major versions that are used to run OpenSearch
runtime.jdk = 17,16,15,14,13,12,11,8
```

OVI3D0:
> This change may fix the integration tests as run via GitHub actions. However, integ tests can be run directly when OSB is installed in development mode, with `make it`, and this change will not be applicable in that case. Also, switching the JDK back and forth makes this a bit harder to understand.
> 
> The better option would probably be to add JDK 21 to the `provision_config` resources:
> 
> ```
> # major version of the JDK that is used to build OpenSearch
> build.jdk = 17
> # list of JDK major versions that are used to run OpenSearch
> runtime.jdk = 17,16,15,14,13,12,11,8
> ```

Discussed offline:
The current solution in this PR is a short term workaround while we come up with a more elegant solution. Adding JDK 21 to the `provision_config` resources and adding a JDK 21 installation caused other failures, which we are assuming was caused by the newer JDK version. 

In the meantime, integration tests run with Github Actions will work, but running these tests locally with `make it` will not. An issue was opened [here](https://github.com/opensearch-project/opensearch-benchmark/issues/738#issue-2816994760) with more details.

gkamat (approved):
Should be an interim solution to get the GHA integ tests working.  Still need to fix the actual integ tests when run via command line.

Issue #842, opened 2025-05-19
Refactor AsyncExecutor


OVI3D0:
### Is your feature request related to a problem? Please describe

[AsyncExecutor](https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/worker_coordinator/worker_coordinator.py#L1906) is the class used to define the logic of individual clients in OpenSearch Benchmark (OSB). As OSB's capabilities have grown, so have the capabilities of the AsyncExecutor class. However, the __call__ method for this class has now grown to 150+ lines and has a bunch of if-else statements. This makes the class very hard to follow and contribute to.

### Describe the solution you'd like

The AsyncExecutor class should be refactored so parts of the logic are abstracted out into appropriately named methods that can be invoked from the __call__ method depending on the conditions. This would make the __call__ method much easier to follow and modify in the future.

rustemd02:
Hi! may I work on this?

Issue #480, opened 2024-03-12
[BUG] [Need-Help]Timeout error while creating a custom workload


VinDataR:
We are trying to create a custom workload from a existing OpenSearch cluster using a below command; 

```
opensearch-benchmark create-workload \
--workload="sample-index" \
--target-hosts="opensearch-master:9200" \
--client-options="basic_auth_user:'****',basic_auth_password:'****'" \
--indices="sample-index-5" \
--output-path="/opensearch-benchmark/.benchmark" --offline 
```
This command is constantly ending in timeout error. Default timeout is mentioned as 10 seconds. We are not seeing an explicit flag to increase timeout from default 10s. Does anyone has any idea how to increase the timeout for create-workload command. Attached error screenshot.

![image (4)](https://github.com/opensearch-project/opensearch-benchmark/assets/163123878/06886729-7e1b-45a3-9394-dd7ca989aa52)

IanHoang:
Hi @VinodRakuten, we currently have an [RFC](https://github.com/opensearch-project/opensearch-benchmark/issues/395) for enhancing the `create-workload` feature and mentions this issue and a proposed solution. We'll be implementing these enhancements soon. Feel free to comment on the RFC if you have any comments or questions.

### To better help you:
- What's the command you're running (Please redact any information from the command as we only want to check the flags that are being used)?
- How many indices are you running against?
- How large are these indices in terms of GB?
- What's the average size of the shards?

### Manual Work-around
For very large indices, one work-around that I've had successful experiences with is running the `create-workload` feature on indices independently. This will essentially create separate workloads and you'll have to combine the workload.jsons that are produced. For example, if I want to extract corpora from indices A, B, and C, I'd run the `create-workload` feature separately for each index. Afterwards, I'll have three different workloads -- one for A, one for B, and one for C -- and I'd create a new workload.json that combines elements from all three workloads' workload.json. 

This isn't ideal but it gets the job done. Some of these issues should be resolved once we enhance this feature in the future based on this [RFC](https://github.com/opensearch-project/opensearch-benchmark/issues/395) plan.

VinDataR:
Thank you for your reponse @IanHoang.  We tried the manual workaround by merging the workload.json but it didn't work. Maybe we might missed some pieces while combining them. We'll give it another shot.

IanHoang:
[...]

PR #843, opened 2025-05-20
REFACTOR async def __call__(self, *args, **kwargs)

rustemd02:
### Description
Improved Readability and Maintainability of AsyncExecutor's `__call__` Method

The following private helper methods have been introduced to encapsulate distinct parts of the original logic:

*   **`async def _wait_for_rampup(self, rampup_wait_time)`**:
    *   Handles the logic for waiting during the ramp-up phase.
*   **`def _handle_client_state(self, was_paused, client_state, total_start, expected_scheduled_time)`**:
    *   Manages client state adjustments, particularly relevant for redline testing when clients might be paused or resumed.
*   **`async def _prepare_context_manager(self, params)`**:
    *   Determines and prepares the correct context manager for the outgoing request. This includes specific handling for "produce-stream-message" operations (using `MessageProducerFactory`) and "vector-search" operations (adding client and core counts to params).
*   **`async def _execute_request(self, params, expected_scheduled_time, total_start)`**:
    *   Encapsulates the main request execution flow:
        *   Calculates and applies throughput throttling.
        *   Manages request timing (absolute processing start, actual processing start/end).
        *   Invokes `_prepare_context_manager`.
        *   Calls `execute_single` to perform the actual operation.
        *   Initiates error reporting via `_report_error` if a request fails during redline testing.
        *   Returns a dictionary of timing and result data.
*   **`def _report_error(self, request_meta_data)`**:
    *   Handles the logic for sending error information to the `FeedbackActor` via the `error_queue` if an error occurs during redline testing.
*   **`def _process_results(self, result_data, total_start, client_state, percent_completed)`**:
    *   Takes the results from `_execute_request` and:
        *   Calculates various time metrics (service time, client processing time, processing time, latency).
        *   Calls `schedule_handle.after_request`.
        *   Determines task completion status based on runner state and `self.complete` flag.
        *   Calculates progress.
        *   Adds samples to `self.sampler` if the client is active.
        *   Returns the completion status.
*   **`async def _cleanup(self)`**:
    *   Manages the cleanup of resources after the main loop, specifically stopping the `message_producer` if it was used.

### Issues Resolved
* #842 

### Testing
I've ran redline testing to ensure that method still works

---

### DIFF
--- a/osbenchmark/worker_coordinator/worker_coordinator.py
+++ b/osbenchmark/worker_coordinator/worker_coordinator.py
@@ -2022,24 +2022,32 @@ class AsyncExecutor:
         request_start = request_end = client_request_start = client_request_end = None
         total_ops, total_ops_unit, request_meta_data = 0, "ops", {}
 
-        try:
-            async with context_manager as request_context:
+        async with context_manager as request_context:
+            try:
                 total_ops, total_ops_unit, request_meta_data = await asyncio.wait_for(
                     execute_single(
                         self.runner, self.opensearch, params, self.on_error,
                         redline_enabled=self.redline_enabled, client_enabled=client_state
                     ),
-                    self.base_timeout
+                    timeout=self.base_timeout
                 )
-                request_start = request_context.request_start
-                request_end = request_context.request_end
-                client_request_start = request_context.client_request_start
-                client_request_end = request_context.client_request_end
+            except asyncio.TimeoutError:
+                self.logger.error("Client %s request timed out after %s s", self.client_id, self.base_timeout)
+                request_meta_data = {"success": False, "error-type": "timeout"}
 
-        except asyncio.TimeoutError:
-            self.logger.error("Client %s request timed out after %s s", self.client_id, self.base_timeout)
-            request_meta_data = {"success": False, "error-type": "timeout"}
+                # Simulate full timing lifecycle
+                request_context_holder.on_client_request_start()
+                request_context_holder.on_request_start()
+                request_context_holder.on_request_end()
+                request_context_holder.on_client_request_end()
 
+            # Now safely extract request timings
+            request_start = request_context.request_start
+            request_end = request_context.request_end
+            client_request_start = request_context.client_request_start
+            client_request_end = request_context.client_request_end
+
+        # If request failed or timings weren't properly captured, fall back
         if not request_meta_data.get("success") or None in (request_start, request_end, client_request_start,
                                                             client_request_end):
             if request_start is None:
@@ -2051,6 +2059,7 @@ class AsyncExecutor:
                 request_end = now
             if client_request_end is None:
                 client_request_end = now
+
             if not request_meta_data.get("skipped", False):
                 error_info = {
                     "client_id": self.client_id,
--- a/osbenchmark/worker_coordinator/worker_coordinator.py
+++ b/osbenchmark/worker_coordinator/worker_coordinator.py
@@ -1966,21 +1966,184 @@ class AsyncExecutor:
         self.error_queue = error_queue
         self.queue_lock = queue_lock
         self.redline_enabled = self.cfg.opts("workload", "redline.test", mandatory=False) if self.cfg else False
+        # Variables to keep track of during execution
+        self.expected_scheduled_time = 0
+        self.sample_type = None
+        self.runner = None
+        self.task_completes_parent = False
+
+    async def _wait_for_rampup(self, rampup_wait_time):
+        """Wait for the ramp-up phase if needed."""
+        if rampup_wait_time:
+            self.logger.info("client id [%s] waiting [%.2f]s for ramp-up.", self.client_id, rampup_wait_time)
+            await asyncio.sleep(rampup_wait_time)
+            self.logger.info("Client id [%s] is running now.", self.client_id)
+
+    def _handle_client_state(self, was_paused, client_state, total_start, expected_scheduled_time):
+        """Handle client state for redline testing."""
+        if client_state and was_paused:
+            now = time.perf_counter()
+            total_start = now - expected_scheduled_time
+            was_paused = False
+        elif not client_state:
+            was_paused = True
+        return was_paused, total_start
+
+    async def _prepare_context_manager(self, params):
+        """Prepare the appropriate context manager for the request."""
+        if params is not None and params.get("operation-type") == "produce-stream-message":
+            if self.message_producer is None:
+                self.message_producer = await client.MessageProducerFactory.create(params)
+            params.update({"message-producer": self.message_producer})
+            return self.message_producer.new_request_context()
+        else:
+            context_manager = self.opensearch["default"].new_request_context()
+            # Add parameters for vector-search
+            if params is not None and params.get("operation-type") == "vector-search":
+                available_cores = int(self.cfg.opts("system", "available.cores", mandatory=False,
+                                                    default_value=multiprocessing.cpu_count()))
+                params.update({"num_clients": self.task.clients, "num_cores": available_cores})
+            return context_manager
+
+    async def _execute_request(self, params, expected_scheduled_time, total_start):
+        """Execute a request with timing control."""
+        # Calculate timing and check if throttling is needed
+        absolute_expected_schedule_time = total_start + expected_scheduled_time
+        throughput_throttled = expected_scheduled_time > 0
+
+        if throughput_throttled:
+            rest = absolute_expected_schedule_time - time.perf_counter()
+            if rest > 0:
+                await asyncio.sleep(rest)
+
+        absolute_processing_start = time.time()
+        processing_start = time.perf_counter()
+        self.schedule_handle.before_request(processing_start)
+
+        # Prepare context manager
+        context_manager = await self._prepare_context_manager(params)
+
+        # Execute the request
+        async with context_manager as request_context:
+            total_ops, total_ops_unit, request_meta_data = await execute_single(
+                self.runner, self.opensearch, params, self.on_error, self.redline_enabled)
+            request_start = request_context.request_start
+            request_end = request_context.request_end
+            client_request_start = request_context.client_request_start
+            client_request_end = request_context.client_request_end
+
+            # Handle errors for redline testing
+            if not request_meta_data["success"] and not request_meta_data.get("skipped", False):
+                self._report_error(request_meta_data)
+
+        processing_end = time.perf_counter()
+
+        return {
+            "absolute_processing_start": absolute_processing_start,
+            "processing_start": processing_start,
+            "processing_end": processing_end,
+            "request_start": request_start,
+            "request_end": request_end,
+            "client_request_start": client_request_start,
+            "client_request_end": client_request_end,
+            "total_ops": total_ops,
+            "total_ops_unit": total_ops_unit,
+            "request_meta_data": request_meta_data,
+            "throughput_throttled": throughput_throttled
+        }
+
+    def _report_error(self, request_meta_data):
+        """Report an error to the FeedbackActor."""
+        if self.error_queue is not None:
+            self.logger.error("Real error detected in client %s. Notifying FeedbackActor...", self.client_id)
+            error_info = {
+                "client_id": self.client_id,
+                "task": str(self.task),
+                "error_details": request_meta_data
+            }
+            self.report_error(error_info)
+
+    def _process_results(self, result_data, total_start, client_state, percent_completed):
+        """Process results from a request."""
+        # Calculate execution times
+        service_time = result_data["request_end"] - result_data["request_start"]
+        client_processing_time = (result_data["client_request_end"] - result_data[
+            "client_request_start"]) - service_time
+        processing_time = result_data["processing_end"] - result_data["processing_start"]
+        time_period = result_data["request_end"] - total_start
+
+        # Update statistics
+        self.schedule_handle.after_request(
+            result_data["processing_end"],
+            result_data["total_ops"],
+            result_data["total_ops_unit"],
+            result_data["request_meta_data"]
+        )
+
+        # Check if runner provided throughput
+        throughput = result_data["request_meta_data"].pop("throughput", None)
+
+        # Calculate latency
+        latency = (result_data["request_end"] - (total_start + self.expected_scheduled_time)
+                   if result_data["throughput_throttled"] else service_time)
+
+        # Determine task completion
+        if self.task_completes_parent:
+            completed = self.runner.completed
+        else:
+            completed = self.complete.is_set() or self.runner.completed
+
+        # Determine progress
+        if completed:
+            progress = 1.0
+        elif self.runner.percent_completed:
+            progress = self.runner.percent_completed
+        else:
+            progress = percent_completed
+
+        # Add sample if client is active
+        if client_state:
+            self.sampler.add(
+                self.task, self.client_id, self.sample_type,
+                result_data["request_meta_data"],
+                result_data["absolute_processing_start"],
+                result_data["request_start"],
+                latency, service_time, client_processing_time, processing_time,
+                throughput, result_data["total_ops"], result_data["total_ops_unit"],
+                time_period, progress,
+                result_data["request_meta_data"].pop("dependent_timing", None)
+            )
+
+        return completed
+
+    async def _cleanup(self):
+        """Clean up resources after task execution."""
+        if self.message_producer is not None:
+            await self.message_producer.stop()
+            self.message_producer = None  # Reset for future calls
+
+    def report_error(self, error_info):
+        """Report an error to the error queue."""
+        if self.error_queue is not None:
+            try:
+                self.error_queue.put_nowait(error_info)
+            except queue.Full:
+                self.logger.warning("Error queue full; dropping error from client %s", self.client_id)
 
     async def __call__(self, *args, **kwargs):
-        task_completes_parent = self.task.completes_parent
+        has_run_any_requests = False
+        was_paused = False
+        self.task_completes_parent = self.task.completes_parent
         total_start = time.perf_counter()
-        # lazily initialize the schedule
+
+        # Initialize schedule
         self.logger.debug("Initializing schedule for client id [%s].", self.client_id)
         schedule = self.schedule_handle()
         self.schedule_handle.start()
         rampup_wait_time = self.schedule_handle.ramp_up_wait_time
-        if rampup_wait_time:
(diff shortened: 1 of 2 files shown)


OVI3D0:
Hey @rustemd02 thanks for the contribution? Would you be able to rebase these changes since #849 was just merged in?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1994,113 +2157,65 @@ async def __call__(self, *args, **kwargs):
...
 
+        self.logger.debug("Entering main loop for client id [%s].", self.client_id)
+        try:
+            if self.cfg is not None:
+                client_options_obj = self.cfg.opts("client", "options")
+                client_options = getattr(client_options_obj, "all_client_options", {}) or {}
+            else:
+                client_options = {}
+        except exceptions.ConfigError:
+            client_options = {}
```
OVI3D0: Looks like this block to fetch the client options is repeated twice. Maybe it can be abstracted into another small method like:
```
def _get_client_options(self):
        try:
            if self.cfg is not None:
                client_options_obj = self.cfg.opts("client", "options")
                return getattr(client_options_obj, "all_client_options", {}) or {}
            else:
                return {}
        except exceptions.ConfigError:
            return {}
```

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1966,21 +1966,184 @@ def __init__(self, client_id, task, schedule, opensearch, sampler, cancel, compl
...
+            "request_end": request_end,
+            "client_request_start": client_request_start,
+            "client_request_end": client_request_end,
+            "total_ops": total_ops,
+            "total_ops_unit": total_ops_unit,
+            "request_meta_data": request_meta_data,
+            "throughput_throttled": throughput_throttled
+        }
+
+    def _report_error(self, request_meta_data):
```
OVI3D0: Can we add type hints to all these new methods?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1966,21 +1966,184 @@ def __init__(self, client_id, task, schedule, opensearch, sampler, cancel, compl
...
+        else:
+            context_manager = self.opensearch["default"].new_request_context()
+            # Add parameters for vector-search
+            if params is not None and params.get("operation-type") == "vector-search":
+                available_cores = int(self.cfg.opts("system", "available.cores", mandatory=False,
+                                                    default_value=multiprocessing.cpu_count()))
+                params.update({"num_clients": self.task.clients, "num_cores": available_cores})
+            return context_manager
+
+    async def _execute_request(self, params, expected_scheduled_time, total_start):
```
OVI3D0: Nice, so this method replaces the need for handle_redline and handle_benchmark?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1966,21 +1966,184 @@ def __init__(self, client_id, task, schedule, opensearch, sampler, cancel, compl
...
+        absolute_processing_start = time.time()
+        processing_start = time.perf_counter()
+        self.schedule_handle.before_request(processing_start)
+
+        # Prepare context manager
+        context_manager = await self._prepare_context_manager(params)
+
+        # Execute the request
+        async with context_manager as request_context:
+            total_ops, total_ops_unit, request_meta_data = await execute_single(
```
OVI3D0: I think we should add back the asyncio timer we previously had wrapped around execute_single:
```
            # apply per-request timeout
            total_ops, total_ops_unit, request_meta_data = await asyncio.wait_for(execute_single(runner, self.opensearch,
                                                                                params, self.on_error,
                                                                                redline_enabled=True,
                                                                                client_enabled=client_state), base_timeout)
```

This timeout ensured requests wouldn't hang for super long periods of time before timing out

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1966,21 +1966,184 @@ def __init__(self, client_id, task, schedule, opensearch, sampler, cancel, compl
...
+                self.runner, self.opensearch, params, self.on_error, self.redline_enabled)
```
OVI3D0: Also, looks like we're no longer passing the client_state into the execute_single function, which is essential for determining whether a client should or should not be sending requests to its target host

OVI3D0 (changes requested):
Thanks for the PR @rustemd02 ! Overall, this will make the AsyncExecutor much more readable and maintainable. 

I left a few comments, and now that we have all these new helper methods, we can also easily test and validate each one in isolation. Can we write some basic unit tests for the new helper methods to ensure they're working properly?

Also, can you attach some test results before/after your changes just to ensure we're not affecting benchmark or redline test results?

rustemd02:
Thank you so much for the detailed review! 
I've pushed a new commit that incorporates your suggestions.

Also, as requested, I ran a few redline tests to compare the performance between the main branch (before changes) and my branch.

Before refactoring (`main` branch)
Снимок экрана 2025-06-14 в 17 53 07

After refactoring 
Снимок экрана 2025-06-14 в 17 52 21

Please let me know if there's anything else I can help with

OVI3D0:
Hey @rustemd02 thanks for the update! Just looks like you have some [unit test errors](https://github.com/opensearch-project/opensearch-benchmark/actions/runs/15653214725/job/44196180047?pr=843). You can run your tests locally with `pytest tests/worker_coordinator/worker_coordinator_test.py` or running `make test`

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1966,118 +1966,245 @@ def __init__(self, client_id, task, schedule, opensearch, sampler, cancel, compl
...
+            return {}
+
+    async def _wait_for_rampup(self, rampup_wait_time: float) -> None:
+        """Wait for the ramp-up phase if needed."""
+        if rampup_wait_time:
+            self.logger.info("client id [%s] waiting [%.2f]s for ramp-up.", self.client_id, rampup_wait_time)
+            await asyncio.sleep(rampup_wait_time)
+            self.logger.info("Client id [%s] is running now.", self.client_id)
+
+    def _handle_client_state(self, was_paused: bool, client_state: bool, total_start: float,
```
OVI3D0: We can actually get rid of this method, in the new AsyncExecutor logic (#849) we don't mess around with the scheduling time, instead we just throw a fake request (no-op) if client_enabled/client_state is False (see the current [execute_single](https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/worker_coordinator/worker_coordinator.py#L2161) function)

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1966,118 +1966,245 @@ def __init__(self, client_id, task, schedule, opensearch, sampler, cancel, compl
...
+
+        # Prepare context manager
+        context_manager = await self._prepare_context_manager(params)
+
+        # Get client options and base timeout
+        client_options = self._get_client_options()
+        base_timeout = int(client_options.get("base_timeout", 5))
+
+        # Execute the request with timeout
+        async with context_manager as request_context:
```
OVI3D0, line 2025: I like this, but there's some important error checking in [handle_redline](https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/worker_coordinator/worker_coordinator.py#L2105-L2143) that we're missing here: If we're in redline mode and are checking request latency, we need to be monitoring this and adding any failed requests to the error queue. 

Correct me if I'm wrong though? I do see the `report_error` method being used below, but I want to be sure we're catching all request failures including ones that timeout due to the base_timeout

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1966,118 +1966,245 @@ def __init__(self, client_id, task, schedule, opensearch, sampler, cancel, compl
...
+        """Report an error to the error queue."""
+        if self.error_queue is not None:
+            try:
+                self.error_queue.put_nowait(error_info)
+            except queue.Full:
+                self.logger.warning("Error queue full; dropping error from client %s", self.client_id)
 
     async def __call__(self, *args, **kwargs):
-        task_completes_parent = self.task.completes_parent
+        has_run_any_requests = False
```
OVI3D0: Similar to my above comment, we can get rid of all this was_paused and has_run_any_requests logic, since we should ideally just send a no-op if client_state is false

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1966,118 +1966,245 @@ def __init__(self, client_id, task, schedule, opensearch, sampler, cancel, compl
...
+
+                # Skip if client is paused
+                if not client_state:
+                    processing_end = time.perf_counter()
+                    self.schedule_handle.after_request(
+                        processing_end, 0, "ops", {"success": False, "skipped": True}
+                    )
+
+                    if self.complete.is_set():
+                        break
```
OVI3D0: Same as my other two comments

rustemd02:
I've totally missed that, thank you for pointing out those issues!

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1967,117 +1959,222 @@ def __init__(self, client_id, task, schedule, opensearch, sampler, cancel, compl
...
-        self.logger.debug("Initializing schedule for client id [%s].", self.client_id)
-        schedule = self.schedule_handle()
-        self.schedule_handle.start()
-        rampup_wait_time = self.schedule_handle.ramp_up_wait_time
-        if rampup_wait_time:
-            self.logger.info("client id [%s] waiting [%.2f]s for ramp-up.", self.client_id, rampup_wait_time)
-            await asyncio.sleep(rampup_wait_time)
+        # Client options are fetched once during initialization, not on every request.
+        self.client_options = self._get_client_options()
+        self.base_timeout = int(self.client_options.get("base_timeout", 5))
```
OVI3D0: The historical base timeout for OSB is [10 seconds](https://github.com/opensearch-project/opensearch-benchmark/issues/480#issue-2181280938), let's set it to that instead of 5

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1967,117 +1959,222 @@ def __init__(self, client_id, task, schedule, opensearch, sampler, cancel, compl
...
+                "success": False,
+                "error-type": "timeout",
+                "error-description": f"Request timed out after {self.base_timeout} seconds."
+            }
+            # Ensure request times are set to avoid errors in metrics calculation
+            request_start = processing_start
+            request_end = time.perf_counter()
+            client_request_start = processing_start
+            client_request_end = request_end
+            self._report_error(request_meta_data)
```
OVI3D0, line 2055: same here, we only report to the error queue if in redline mode

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1967,117 +1959,222 @@ def __init__(self, client_id, task, schedule, opensearch, sampler, cancel, compl
...
+    def _report_error(self, request_meta_data: dict) -> None:
+        """Report an error to the FeedbackActor."""
+        if self.error_queue is not None:
+            self.logger.error("Real error detected in client %s. Notifying FeedbackActor...", self.client_id)
+            error_info = {
+                "client_id": self.client_id,
+                "task": str(self.task),
+                "error_details": request_meta_data
+            }
+            self.report_error(error_info)
```
OVI3D0: Let's combine this and the other report_error function? No need to have two methods named the same that are looking to achieve the same purpose

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -2086,74 +2183,13 @@ async def __call__(self, *args, **kwargs):
...
-            request_start = request_context.request_start
-            request_end = request_context.request_end
-            client_request_start = request_context.client_request_start
-            client_request_end = request_context.client_request_end
-        else:
-            # normal path: read timings from context
-            request_start = request_context.request_start
-            request_end = request_context.request_end
-            client_request_start = request_context.client_request_start
-            client_request_end = request_context.client_request_end
```
OVI3D0, line 2132: Can we restore this logic if redline is enabled? Maybe in execute_single? I just tried a redline test with your changes and got this error:
```
Workload: big5
Test Procedure: timed-mode-test-procedure
Cause: Unknown
Error: Cannot run task [range-auto-date-histo]: 'client_request_end'

  File "/Users/mikeovi/.pyenv/versions/3.9.17/lib/python3.9/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)

  File "/Users/mikeovi/workplace/opensearch-benchmark/osbenchmark/worker_coordinator/worker_coordinator.py", line 1845, in __call__
    loop.run_until_complete(self.run())

  File "/Users/mikeovi/.pyenv/versions/3.9.17/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()

  File "/Users/mikeovi/workplace/opensearch-benchmark/osbenchmark/worker_coordinator/worker_coordinator.py", line 1892, in run
    _ = await asyncio.gather(*aws)

  File "/Users/mikeovi/workplace/opensearch-benchmark/osbenchmark/worker_coordinator/worker_coordinator.py", line 2184, in __call__
    raise exceptions.BenchmarkError(f"Cannot run task [{self.task}]: {e}") from None

osbenchmark.exceptions.BenchmarkError: Cannot run task [range-auto-date-histo]: 'client_request_end'
```

OVI3D0, line 2132: I think we should restore this logic in _execute_request or execute_single, I just tried a redline test with your changes and got:
```
Workload: big5
Test Procedure: timed-mode-test-procedure
Cause: Unknown
Error: Cannot run task [range-auto-date-histo]: 'client_request_end'

  File "/Users/mikeovi/.pyenv/versions/3.9.17/lib/python3.9/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)

  File "/Users/mikeovi/workplace/opensearch-benchmark/osbenchmark/worker_coordinator/worker_coordinator.py", line 1845, in __call__
    loop.run_until_complete(self.run())

  File "/Users/mikeovi/.pyenv/versions/3.9.17/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()

  File "/Users/mikeovi/workplace/opensearch-benchmark/osbenchmark/worker_coordinator/worker_coordinator.py", line 1892, in run
    _ = await asyncio.gather(*aws)

  File "/Users/mikeovi/workplace/opensearch-benchmark/osbenchmark/worker_coordinator/worker_coordinator.py", line 2184, in __call__
    raise exceptions.BenchmarkError(f"Cannot run task [{self.task}]: {e}") from None

osbenchmark.exceptions.BenchmarkError: Cannot run task [range-auto-date-histo]: 'client_request_end'
```

Which I used to get typically when the request/client start/end values weren't being populated after a skipped request

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1967,117 +1959,219 @@ def __init__(self, client_id, task, schedule, opensearch, sampler, cancel, compl
...
+        processing_start = time.perf_counter()
+        self.schedule_handle.before_request(processing_start)
+
+        context_manager = await self._prepare_context_manager(params)
+
+        request_start = request_end = client_request_start = client_request_end = None
+        total_ops, total_ops_unit, request_meta_data = 0, "ops", {}
+
+        try:
+            async with context_manager as request_context:
```
OVI3D0: so `_execute_request` works for normal benchmarks but not for redline tests. I noticed here that you moved the `asnyc with` block inside the `try` block. I think this is the root cause of the error, as any timed out request will break the async block early and cause the crash I described in a previous comment.

Here's a working version of this method you can drop in to this PR. Feel free to modify it if you need to:
```
    async def _execute_request(self, params: dict, expected_scheduled_time: float, total_start: float,
                               client_state: bool) -> dict:
        """Execute a request with timing control and error handling."""
        absolute_expected_schedule_time = total_start + expected_scheduled_time
        throughput_throttled = expected_scheduled_time > 0

        if throughput_throttled:
            rest = absolute_expected_schedule_time - time.perf_counter()
            if rest > 0:
                await asyncio.sleep(rest)

        absolute_processing_start = time.time()
        processing_start = time.perf_counter()
        self.schedule_handle.before_request(processing_start)

        context_manager = await self._prepare_context_manager(params)

        request_start = request_end = client_request_start = client_request_end = None
        total_ops, total_ops_unit, request_meta_data = 0, "ops", {}

        async with context_manager as request_context:
            try:
                total_ops, total_ops_unit, request_meta_data = await asyncio.wait_for(
                    execute_single(
                        self.runner, self.opensearch, params, self.on_error,
                        redline_enabled=self.redline_enabled, client_enabled=client_state
                    ),
                    timeout=self.base_timeout
                )
            except asyncio.TimeoutError:
                self.logger.error("Client %s request timed out after %s s", self.client_id, self.base_timeout)
                request_meta_data = {"success": False, "error-type": "timeout"}

                # Simulate full timing lifecycle
                request_context_holder.on_client_request_start()
                request_context_holder.on_request_start()
                request_context_holder.on_request_end()
                request_context_holder.on_client_request_end()

            # Now safely extract request timings
            request_start = request_context.request_start
            request_end = request_context.request_end
            client_request_start = request_context.client_request_start
            client_request_end = request_context.client_request_end

        # If request failed or timings weren't properly captured, fall back
        if not request_meta_data.get("success") or None in (request_start, request_end, client_request_start, client_request_end):
            if request_start is None:
                request_start = processing_start
            if client_request_start is None:
                client_request_start = processing_start
            now = time.perf_counter()
            if request_end is None:
                request_end = now
            if client_request_end is None:
                client_request_end = now

            if not request_meta_data.get("skipped", False):
                error_info = {
                    "client_id": self.client_id,
                    "task": str(self.task),
                    "error_details": request_meta_data
                }
                self.report_error(error_info)

        processing_end = time.perf_counter()

        return {
            "absolute_processing_start": absolute_processing_start,
            "processing_start": processing_start,
            "processing_end": processing_end,
            "request_start": request_start,
            "request_end": request_end,
            "client_request_start": client_request_start,
            "client_request_end": client_request_end,
            "total_ops": total_ops,
            "total_ops_unit": total_ops_unit,
            "request_meta_data": request_meta_data,
            "throughput_throttled": throughput_throttled
        }
```

Issue #724, opened 2025-01-09
[WIP] Open Up OSB to Benchmark Other Offerings and Databases


IanHoang:
### Is your feature request related to a problem? Please describe

There have been discussions of opening up OSB into a tool that's capable of benchmarking other databases and offerings in issues like [this one](https://github.com/opensearch-project/opensearch-benchmark/issues/381#issuecomment-1751096785). 

### Describe the solution you'd like

Ideally, OSB in the future will be agnostic and could use a separate repository as plugins so that it can connect and benchmark against various databases and offerings. Opening up this issue for further discussion.

PR #329, opened 2023-06-12
Adding SigV4 support to opensearch-benchmarks for connecting to test clusters and results publishing

sachetalva:
### Description
This PR adds support in OpenSearch benchmark to connect to OpenSearch clusters using SigV4 for running tests or reporting test results.

New configuration options added with this change:
1. New parameters supported with `--client-options` argument in `execute-test`:
  - `amazon_aws_log_in`: Value can be one of `client-options` or `environment`
    - if `amazon_aws_log_in` is  set as `client-options`, then the below configuration is mandatory in client-options:
      - `aws_access_key_id`: The AWS access key ID to use
      - `aws_secret_access_key`: The AWS secret key to use
      - `service`: `es` for Amazon OpenSearch service
      - `region`: The region in which the cluster is available, example: `us-east-1`
    - if `amazon_aws_log_in` is  set as `environment`, then the below environment variables should be set:
      - `OSB_AWS_ACCESS_KEY_ID`: The AWS access key ID to use
      - `OSB_AWS_SECRET_ACCESS_KEY`: The AWS secret key to use
      - `OSB_SERVICE`: `es` for Amazon OpenSearch service
      - `OSB_REGION`: The region in which the cluster is available, example: `us-east-1`
2. New parameters in `reporting` section of benchmark.ini:
  - `datastore.amazon_aws_log_in`: Value can be one of `config` or `environment`
    - if `datastore.amazon_aws_log_in` is  set as `config`, then below configuration is mandatory in `reporting` section in benchmark.ini:
      -  `datastore.aws_access_key_id`: The AWS access key ID to use
      - `datastore.aws_secret_access_key`:  The AWS secret key to use
      - `datastore.service`: `es` for Amazon OpenSearch service
      - `datastore.region`: The region in which the cluster is available, example: `us-east-1`
    - if `datastore.amazon_aws_log_in` is  set as `environment`, then the below environment variables should be set:
      - `OSB_DATASTORE_AWS_ACCESS_KEY_ID`: The AWS access key ID to use
      - `OSB_DATASTORE_AWS_SECRET_ACCESS_KEY`: The AWS secret key to use
      - `OSB_DATASTORE_SERVICE`: `es` for Amazon OpenSearch service
      - `OSB_DATASTORE_REGION`: The region in which the cluster is available, example: `us-east-1`

### Issues Resolved
https://github.com/opensearch-project/opensearch-benchmark/issues/158

### Testing
- [x] New functionality includes testing
- [x] Executed `make test`
[...]

PR #845, opened 2025-05-21
Make OSB Cloud Agnostic: Refactor OSB Client Logic

IanHoang:
### Description
OSB currently supports AWS SigV4 since [this PR](https://github.com/opensearch-project/opensearch-benchmark/pull/329/files#diff-c0986e3e384cd7383b1f6aca6e5527fd566fb2ee29c755ed9ca6bfed68b575dc) a few years back. During one of the OSB triage and office hours meetings, one community member suggested abstracting cloud provider code so that it's not implemented into the core logic. 

This PR abstracts cloud provide logic into a new module called `cloud_provider`. To add support for other cloud providers, developers just have to add an implementation in the `vendors` directory. The previous implementation was baked directly into core logic and also required that users specify `datastore.user` and `datastore.password` even if their AWS credentials were provided in the benchmark.ini. This PR removes that requirement and adds additional guard clauses. 

### Issues Resolved
https://github.com/opensearch-project/opensearch-benchmark/issues/724

### Testing
- [x] Tested with command line specifying client options with AWS credentials
- [x] Tested with benchmark.ini pointing to OpenSearch datastore with AWS credentials

---


osbenchmark/cloud_providers.py
```diff
@@ -0,0 +1,226 @@
...
+        if (not log_in_dict["aws_access_key_id"] or not log_in_dict["aws_secret_access_key"]
+                or not log_in_dict["service"] or not log_in_dict["region"]):
+            self.logger.error("Invalid amazon aws log in parameters, required input aws_access_key_id, "
+                              "aws_secret_access_key, service and region.")
+            raise exceptions.SystemSetupError(
+                "Invalid amazon aws log in parameters, required input aws_access_key_id, "
+                "aws_secret_access_key, and region."
+            )
+
+        if log_in_dict["service"] not in ['es', 'aoss']:
```
OVI3D0, line 103: this can use the AVAILABLE_SERVICES list already defined at the top of the class

osbenchmark/cloud_providers.py
```diff
@@ -0,0 +1,226 @@
...
+                elif metrics_amazon_aws_log_in == 'config':
+                    missing_aws_credentials_message = "Missing AWS credentials through datastore.aws_access_key_id, " \
+                                                      "datastore.aws_secret_access_key, datastore.region, " \
+                                                      "datastore.service in the config file."
+                else:
+                    missing_aws_credentials_message = "datastore.amazon_aws_log_in can only be one of " \
+                                                      "'environment' or 'config'"
+                raise exceptions.ConfigError(missing_aws_credentials_message) from None
+
+            if (metrics_aws_service not in ['es', 'aoss']):
```
OVI3D0, line 160: This can also AVAILABLE_SERVICES defined up top

osbenchmark/cloud_providers.py
```diff
@@ -0,0 +1,226 @@
...
+            log_in_dict["aws_access_key_id"] = client_options.get("aws_access_key_id")
+            log_in_dict["aws_secret_access_key"] = client_options.get("aws_secret_access_key")
+            log_in_dict["region"] = client_options.get("region")
+            log_in_dict["service"] = client_options.get("service")
+            # optional: applicable only for role-based access
+            log_in_dict["aws_session_token"] = client_options.get("aws_session_token")
+
+        if (not log_in_dict["aws_access_key_id"] or not log_in_dict["aws_secret_access_key"]
+                or not log_in_dict["service"] or not log_in_dict["region"]):
+            self.logger.error("Invalid amazon aws log in parameters, required input aws_access_key_id, "
```
OVI3D0, line 96: curious: where does logger come from in this file? I don't see it defined or injected anywhere

osbenchmark/cloud_providers.py
```diff
@@ -0,0 +1,226 @@
...
+            metrics_aws_region = config.opts("results_publishing", "datastore.region",
+                                                   default_value=None, mandatory=False)
+            metrics_aws_service = config.opts("results_publishing", "datastore.service",
+                                                    default_value=None, mandatory=False)
+        elif metrics_amazon_aws_log_in == 'environment':
+            metrics_aws_access_key_id = os.getenv("OSB_DATASTORE_AWS_ACCESS_KEY_ID", default=None)
+            metrics_aws_secret_access_key = os.getenv("OSB_DATASTORE_AWS_SECRET_ACCESS_KEY", default=None)
+            metrics_aws_session_token = os.getenv("OSB_DATASTORE_AWS_SESSION_TOKEN", default=None)
+            metrics_aws_region = os.getenv("OSB_DATASTORE_REGION", default=None)
+            metrics_aws_service = os.getenv("OSB_DATASTORE_SERVICE", default=None)
```
OVI3D0, line 138: what's the difference between here and what you're doing in lines 77-92 in `parse_log_in_params`? It looks like in both cases you're either grabbing these parameters from the config options or from environment variables

OVI3D0:
Left a couple small comments but overall this is a good idea👌

osbenchmark/cloud_providers.py
IanHoang: I copied this code from the previous `client.py` class and thought I copied the logger. Good call out!

osbenchmark/cloud_providers.py
IanHoang: I wanted to see what everyone's opinion is for this. The code in `parse_log_in_params_for_metrics` existed in `metrics.py` while the code in `parse_log_in_params` existed in `client.py`. The former handles situations when setting up a client that chats with the datastore with AWS credentials while the latter handles moments when an OpenSearch client is provisioned for all other purposes. 

Was debating on whether it would be better as two methods or just one large method. Logic could exist as one method and depending on the purpose of the client (datastore or everything else), it can plug in the variables to fetch (i.e. `OSB_DATASTORE_AWS_ACCESS_KEY_ID` or `OSB_AWS_ACCESS_KEY_ID`). Let me know ya'lls thoughts

IanHoang: Spoke offline with Michael. Refactored into a single method and tested it to confirm it doesn't break anything.

IanHoang (commented):
Condensed parsing of log in params into a single method and tested E2E to ensure it still works

osbenchmark/cloud_providers.py
gkamat: Consolidating into one function would be cleaner.

osbenchmark/cloud_providers.py
gkamat: And the error messages can be consolidated too.

osbenchmark/cloud_providers.py
```diff
@@ -0,0 +1,226 @@
...
+                                                      "datastore.service in the config file."
+                else:
+                    missing_aws_credentials_message = "datastore.amazon_aws_log_in can only be one of " \
+                                                      "'environment' or 'config'"
+                raise exceptions.ConfigError(missing_aws_credentials_message) from None
+
+            if (metrics_aws_service not in ['es', 'aoss']):
+                raise exceptions.ConfigError("datastore.service can only be one of 'es' or 'aoss'") from None
+
+        self.aws_metrics_log_in_dict['metrics_aws_log_in_choice'] = metrics_amazon_aws_log_in
```
gkamat: Again, use a name without "dict".

osbenchmark/cloud_providers.py
```diff
@@ -0,0 +1,226 @@
...
+        # pass in aws access key id, aws secret access key, aws session token, service and region on command
+        if self.aws_metrics_log_in_dict['metrics_aws_log_in_choice'] is not None:
+            client_options["amazon_aws_log_in"] = 'client_option'
+            client_options["aws_access_key_id"] = self.aws_metrics_log_in_dict['metrics_aws_access_key_id']
+            client_options["aws_secret_access_key"] = self.aws_metrics_log_in_dict['metrics_aws_secret_access_key']
+            client_options["service"] = self.aws_metrics_log_in_dict['metrics_aws_service']
+            client_options["region"] = self.aws_metrics_log_in_dict['metrics_aws_region']
+
+            if self.aws_metrics_log_in_dict['metrics_aws_session_token']:
+                client_options["aws_session_token"] = self.aws_metrics_log_in_dict['metrics_aws_session_token']
```
gkamat, line 182: Simply the assignment might suffice, if `None` is acceptable where used.

osbenchmark/cloud_providers.py
```diff
@@ -0,0 +1,226 @@
...
+    def create_async_client(self, hosts, client_options, client_class):
+        credentials = Credentials(access_key=self.aws_log_in_dict["aws_access_key_id"],
+                            secret_key=self.aws_log_in_dict["aws_secret_access_key"],
+                            token=self.aws_log_in_dict["aws_session_token"])
+        aws_auth = opensearchpy.AWSV4SignerAsyncAuth(credentials, self.aws_log_in_dict["region"],
+                                                     self.aws_log_in_dict["service"])
+        return client_class(hosts=hosts,
+                                        connection_class=async_connection.AsyncHttpConnection,
+                                        use_ssl=True, verify_certs=True, http_auth=aws_auth,
+                                        **client_options)
```
gkamat, line 205: Can these be combined into a single function with `async` as a parameter?  Why is the `aws_auth` computation different?  Finally, does the sync client not support client options, or is that handled differently?

gkamat (changes requested):

IanHoang:
Addressed feedback from @gkamat

OVI3D0 (approved):

gkamat (approved):
This is much better organized than it was in the previous version of the PR.

PR #848, opened 2025-05-23
Revise Docker Build Post PR Releases

IanHoang:
### Description
After each PR is merged in, Github Actions is run to release a new development version of OSB. The Github Actions workflow currently produces two images with different architectures (ARM64 and AMD64). To make it easy to copy-over to OSB's production Dockerhub repository, we should produce a single multi-platform image. 

This PR streamlines the current docker-push-release workflow by:
- Revising it to produce a single image containing both ARM64 and AMD64 architectures
- Revising the tag to be `latest-main`
- Renames the workflow to `Docker Push Release` since there are two `Docker Build` Github Action Workflows which can be confusing to users. The other `Docker Build` workflow is related to `docker-build.yml`

---

### DIFF
--- a/.github/workflows/docker-push-release.yml
+++ b/.github/workflows/docker-push-release.yml
@@ -1,4 +1,4 @@
-name: Docker Build
+name: Docker Push Release
 on:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


gkamat (approved):
This should be updated to be usable for release publishing into the official Docker Hub repo rather than only staging.  The latter is intended for incremental releases on commits, so "push-release" is more applicable to the former.

PR #874, opened 2025-06-11
Introduce CPU-based Feedback for Redline Testing

### LINKED ISSUE #839: Add CPU-aware feedback to redline testing

Currently, the new FeedbackActor introduced for Redline testing is able to scale clients up/down based on whether requests to the cluster are failing or not. 

While we already plan to add latency-aware feedback to this new actor, other metrics that could be measured include the CPU.

Add a flag and a new method for the FeedbackActor to read metrics from a clusters' CPU and adjust the client count based on these metrics.

_No response_

### Additional context

_No response_

OVI3D0:
### Description
Introduces CPU-based feedback for redline testing.

Requires the user to have a metrics store, as well as enabling the node-stats telemetry devices.
The lack of either results in test failure:
```
[ERROR] Cannot execute-test. Error in worker_coordinator (CPU-based feedback requires a metrics store. You are using an in-memory metrics store)
```
And
```
[ERROR] Cannot execute-test. Error in worker_coordinator (Node stats not enabled. You must use `--telemetry node-stats` flag for CPU-based feedback)
```

Also requires the use of the `--redline-cpu-max-usage` flag. This flag takes an int and determines the max threshold for the CPU load of any given node during the test.

Additional flags that are not required, include:
`--redline-cpu-window-seconds` - Controls the size of the window from which OSB checks the average load on each individual node in the cluster. Default is 30, so OSB checks the average CPU load per node over the last 30 seconds.
`--redline-cpu-check-interval` - Controls how often the FeedbackActor queries the users’ datastore to check CPU load. Default is 30 as in every 30 seconds.

This feature takes advantage of OSB's node-stats telemetry device and queries the datastore connected to the load generation host every N seconds. It gathers the CPU usage for each node into 'buckets', and reports any that exceed the threshold to the FeedbackActor's error queue.

Only _one_ node needs to exceed the threshold for the FeedbackActor to trigger a scale-down.

### Issues Resolved
#839 

### Testing

Testing with different config flag configurations + a working configuration. Results shown here against a 3.0 cluster with a max CPU usage of 30%:
Screenshot 2025-06-11 at 3 30 59 PM

---

### DIFF
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -692,6 +692,12 @@ def create_arg_parser():
         help="Maximum number of clients to allow during redline testing. If not set, will default to clients defined in the test procedure.",
         default=None
     )
+    test_execution_parser.add_argument(
+        "--redline-max-cpu-usage",
+        type=int,
+        help="Maximum CPU utilization before scaling back client numbers",
+        default=None
+    )
 
     ###############################################################################
     #
@@ -985,6 +991,7 @@ def configure_test(arg_parser, args, cfg):
         cfg.add(config.Scope.applicationOverride, "workload", "redline.scale_down_pct", args.redline_scaledown_percentage)
         cfg.add(config.Scope.applicationOverride, "workload", "redline.sleep_seconds", args.redline_post_scaledown_sleep)
         cfg.add(config.Scope.applicationOverride, "workload", "redline.max_clients", args.redline_max_clients)
+        cfg.add(config.Scope.applicationOverride, "workload", "redline.max_cpu_usage", args.redline_max_cpu_usage)
     cfg.add(config.Scope.applicationOverride, "workload", "latency.percentiles", args.latency_percentiles)
     cfg.add(config.Scope.applicationOverride, "workload", "throughput.percentiles", args.throughput_percentiles)
     cfg.add(config.Scope.applicationOverride, "workload", "randomization.enabled", args.randomization_enabled)
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -809,7 +809,7 @@ def execute_test(cfg, kill_running_processes=False):
     # redline testing: check metrics store type before running cpu based feedback test
     cpu_max = cfg.opts("workload", "redline.max_cpu_usage", default_value=None)
     if cpu_max is not None:
-        store = metrics.metrics_store(cfg, workload="dummy", test_procedure="dummy", read_only=False)
+        store = metrics.metrics_store(cfg, read_only=False)
         try:
             if isinstance(store, metrics.InMemoryMetricsStore):
                 raise exceptions.SystemSetupError(
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -806,6 +806,19 @@ def execute_test(cfg, kill_running_processes=False):
                   f"Otherwise, you need to manually kill them."
             raise exceptions.BenchmarkError(msg)
 
+    # redline testing: check metrics store type before running cpu based feedback test
+    cpu_max = cfg.opts("workload", "redline.max_cpu_usage", default_value=None)
+    if cpu_max is not None:
+        store = metrics.metrics_store(cfg, workload="dummy", test_procedure="dummy", read_only=False)
+        try:
+            if isinstance(store, metrics.InMemoryMetricsStore):
+                raise exceptions.SystemSetupError(
+                    "CPU-based feedback requires a metrics store, but you're using the in-memory store. "
+                    "Specify a metrics store in your benchmark.ini or via CLI to continue."
+                )
+        finally:
+            store.close()
+
     with_actor_system(test_execution_orchestrator.run, cfg)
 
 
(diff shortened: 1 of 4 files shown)


osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -455,6 +491,57 @@ def scale_up(self) -> None:
...
+                        "script": f"params.avgCpu > {self.max_cpu_threshold}"
+                    }
+                    }
+                }
+                }
+            }
+        }
+        resp = self.os_client.search(index=self.metrics_index, body=body)
+        buckets = resp['aggregations']['nodes']['buckets']
+        if buckets:
```
IanHoang, line 532: We should add a logging statement for when no buckets are found

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -455,6 +491,57 @@ def scale_up(self) -> None:
             self.last_scaleup_time = time.perf_counter()
             self.state = FeedbackState.NEUTRAL
 
+    def _check_cpu_usage(self):
+        """
+        Grab the average CPU load per-node in the past N seconds
+        If any exceed the threshold given, report to the error queue
+        """
+        body = {
```
IanHoang, line 499: WDYT of having queries stored in a central location and being accessed from there? I've seen a number of queries accessed from different places across the repository. 

It doesn't add any functional value but might add some maitenance value. This isn't necessary at the moment but let me know your thoughts and we can always open up an issue as a good first issue later.

OVI3D0: I like this. I thought about adding a separate module for redline related functionality, since the feedback actor is starting to look bloated. It would also make things easier to test. Like [here](https://github.com/opensearch-project/opensearch-benchmark/pull/843)

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -295,6 +313,20 @@ def receiveMsg_ConfigureFeedbackScaling(self, msg, sender):
         self.num_clients_to_scale_up = msg.scale_step
         self.percentage_clients_to_scale_down = msg.scale_down_pct
         self.POST_SCALEDOWN_SECONDS = msg.sleep_seconds
+        # CPU feedback related items
+        self.max_cpu_threshold = msg.cpu_max
```
rishabh6788: wouldn't this error out if msg doesn't contain the cpu related info?

rishabh6788: I see that you are sending in ConfigureFeedbackScaling class.

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -997,14 +1084,36 @@ def start_benchmark(self):
                     self.workers.append(worker)
                     worker_id += 1
         if redline_enabled:
+            metrics_index = None
+            test_execution_id = None
+            # we must have a metrics store connected for CPU based feedback
+            cpu_max = self.config.opts("workload", "redline.max_cpu_usage", default_value=None)
+            if cpu_max and isinstance(self.metrics_store, metrics.InMemoryMetricsStore):
```
rishabh6788, line 1104: is it possible to have these validations before actors are initialized?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -997,14 +1084,36 @@ def start_benchmark(self):
...
+                raise exceptions.SystemSetupError("CPU-based feedback requires a metrics store. You are using an in-memory metrics store")
+            elif cpu_max and 'node-stats' not in self.config.opts("telemetry", "devices"):
```
rishabh6788: Can we instead enable it when user is using cpu based feedback instead of asking them to pass telemetry params for better user experience?

osbenchmark/worker_coordinator/worker_coordinator.py
OVI3D0: right the ConfigureFeedbackScaling class gives each of these default values to avoid errors

osbenchmark/worker_coordinator/worker_coordinator.py
OVI3D0: Good point

osbenchmark/worker_coordinator/worker_coordinator.py
OVI3D0: Maybe in benchmark.py after we gather the args we could read from benchmark.ini to determine whether the metrics store will be in-memory or an opensearch data store. And make the check there

osbenchmark/benchmark.py
```diff
@@ -788,6 +806,19 @@ def execute_test(cfg, kill_running_processes=False):
                   f"Otherwise, you need to manually kill them."
             raise exceptions.BenchmarkError(msg)
 
+    # redline testing: check metrics store type before running cpu based feedback test
+    cpu_max = cfg.opts("workload", "redline.max_cpu_usage", default_value=None)
+    if cpu_max is not None:
+        store = metrics.metrics_store(cfg, workload="dummy", test_procedure="dummy", read_only=False)
```
rishabh6788: why dummy? Can we use the actual values passed by the user?

rishabh6788 (approved):

PR #887, opened 2025-07-07
add ability to use benchmark.ini file to customize redline tests

### LINKED ISSUE #885: Add support for redline test customization using benchmark.ini

Currently, users can run a redline test with the `--redline-test` flag. There are also several optional flags that can be used to fine-tune various parts of a redline test, such as how quickly OSB scales up clients, how many clients are paused at a time, and how long OSB waits before continuing to scale up. With the addition of [CPU-based feedback](https://github.com/opensearch-project/opensearch-benchmark/pull/874), there are now 7 optional flags available to users related to redline testing. However, this can lead to a bad user experience if they want to fully customize a redline test with so many flags.

Users should be able to customize their redline test with the config file already available to OSB, benchmark.ini. 
There should be a static [redline] section added to this file, and the new FeedbackActor should be able to pick from either command flag input or the value available in benchmark.ini.

The priority should be command flag -> config value -> default value.

The new redline section in benchmark.ini would look something like:
```
[redline]
scale_step = 30
scaledown_percentage = .50
post_scaledown_sleep = 10
```

_No response_

### Additional context

_No response_

OVI3D0:
### Description
Adds a new `[redline]` section to `benchmark.ini` along with some default values for redline testing. These are equivalent to the default values originally used in the FeedbackActor, and allow the user to customize various parts of redline testing from the config object rather than using command flags.

The priority for which value to use is: 
command flag -> config object values -> default values in the `ConfigureFeedbackScaling` class.

And the values that can be customized in the config object include:
- The scale_step (how many clients are unpaused per tick)
- The scale_down_percentage (what percentage of clients are scaled down upon an error)
- The sleep_seconds (how long the FeedbackActor chooses to rest before pushing more load on the host)
- The cpu_window_seconds (How long of a window the FeedbackActor takes into account when calculating average CPU load)
- The cpu_check_interval (How often the FeedbackActor queries the datastore to check avg CPU load)

### Issues Resolved
#885 

### Testing

Running redline tests with/without various values in the config object

---

### DIFF
--- a/osbenchmark/worker_coordinator/worker_coordinator.py
+++ b/osbenchmark/worker_coordinator/worker_coordinator.py
@@ -25,12 +25,14 @@
 import asyncio
 import collections
 import concurrent.futures
+import configparser
 import datetime
 import itertools
 import json
 import logging
 import math
 import multiprocessing
+import os
 import queue
 import random
 import sys
@@ -204,6 +206,33 @@ class TaskFinished:
         self.metrics = metrics
         self.next_task_scheduled_in = next_task_scheduled_in
 
+def load_redline_config():
+    config = configparser.ConfigParser()
+    benchmark_home = os.environ.get('BENCHMARK_HOME') or os.environ['HOME']
+    benchmark_ini = benchmark_home + '/.benchmark/benchmark.ini'
+    if not os.path.isfile(benchmark_ini):
+        print(f"WARNING: redline config file {benchmark_ini} not found. Proceeding with default values.")
+        return {}
+
+    config.read(benchmark_ini)
+    config_object = {}
+
+    if "redline" in config:
+        redline = config["redline"]
+        for key in [
+            "scale_step",
+            "scaledown_percentage",
+            "post_scaledown_sleep",
+            "max_cpu_usage",
+            "cpu_window_seconds",
+            "cpu_check_interval",
+            "max_clients"
+        ]:
+            if key in redline:
+                config_object[key] = redline[key]
+    
+    return config_object
+
 class ConfigureFeedbackScaling:
     DEFAULT_SLEEP_SECONDS = 30
     DEFAULT_SCALE_STEP = 5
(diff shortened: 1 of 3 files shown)


osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -204,6 +206,33 @@ def __init__(self, metrics, next_task_scheduled_in):
         self.metrics = metrics
         self.next_task_scheduled_in = next_task_scheduled_in
 
+def load_redline_config():
+    config = configparser.ConfigParser()
+    benchmark_home = os.environ.get('BENCHMARK_HOME') or os.environ['HOME']
+    benchmark_ini = benchmark_home + '/.benchmark/benchmark.ini'
+    if not os.path.isfile(benchmark_ini):
+        print(f"WARNING: redline config file {benchmark_ini} not found. Proceeding with default values.")
```
IanHoang: We should `console.println()`
```
from osbenchmark.utils import console
...
console.println(f"WARNING: redline config file {benchmark_ini} not found. Proceeding with default values.")
```

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -204,6 +206,33 @@ def __init__(self, metrics, next_task_scheduled_in):
...
+        redline = config["redline"]
+        for key in [
+            "scale_step",
+            "scaledown_percentage",
+            "post_scaledown_sleep",
+            "max_cpu_usage",
+            "cpu_window_seconds",
+            "cpu_check_interval",
+            "max_clients"
+        ]:
```
IanHoang, line 230: Nit: would make this a variable so we can easily add to this and reuse anywhere else if needed

IanHoang (approved):
LGTM just left one recommendation and one nit

osbenchmark/worker_coordinator/worker_coordinator.py
OVI3D0: where would you place this variable?

Issue #886, opened 2025-07-07
[RFC] Allow OSB to Benchmark Profiling Metrics


neuenfeldttj:
### Is your feature request related to a problem? Please describe

Currently, OSB benchmarks end-to-end latency for all queries. OpenSearch has a profile flag that enables the user to see the latency of individual metrics (weight creation, scoring, etc.). It would be nice to have to ability for users to enable profiling and have the benchmark display the percentiles for specified profile metrics. While profiling severely increases latency (which the user should be alerted), they would now know the specific timings of query execution. This helps them determine whether their code changes truly created an optimization or not.

### Describe the solution you'd like

Profiling is only used in during search, so a profile metrics JSON object can be given inside of an operation object.
For instance, in `geopoint/operations/default.json` it would look similar to:
```
{
      "name": "polygon",
      "operation-type": "search",
      "body": {
        "query": {
          "geo_polygon": {
            "location": {
              "points": [
                [-0.1, 49.0],
                ...
              ]
            }
          }
        }
      }
      "profile-metrics": [
        "create_weight",
        "scorer",
        ...
      ]
    },
```
Then, the output would display the percentiles for each profile metric for the given operation. For instance, this is running vector-search with some new profiling metrics (exact_search, ann_search): 

<img width="714" height="295" alt="Image" src="https://github.com/user-attachments/assets/60f334bb-39b8-4d3f-a7d8-2b7dfe6925c4" />

Since the profiler is returned as a tree structure, a simple summation of each metric in the tree would give us the total time that metric was called.

markwu-sde:
Hi @neuenfeldttj - thanks for the RFC. Can you link the work that you've been working on related to extending the profiler?

neuenfeldttj:
https://github.com/opensearch-project/OpenSearch/issues/18460
https://github.com/opensearch-project/OpenSearch/pull/18656

IanHoang:
@neuenfeldttj By incorporating profiling, are you referring to this [API](https://docs.opensearch.org/docs/latest/api-reference/search-apis/profile/)? 

What are the benefits of running this benchmark profiling versus running manually? In our experience, most OpenSearch users run the `profile` flag manually and do not seem to run it hundreds of times consecutively (as would happen if incorporated into OSB). Also, the profiling API is resource intensive and would skew the latency / service times metrics.

neuenfeldttj:
[...]

PR #891, opened 2025-07-11
Adding query profiling metrics to metrics store

neuenfeldttj:
### Description
This is a high-level implementation of the issue linked below. Basically, a workload can specify query profile metrics and then the benchmarking tool samples the first N queries with a profiler. OSB then traverses through the entire profile tree and sums the profile metrics that were passed in (if present). The queries sampled for profiling are excluded from the operational metrics to avoid skewing the results. This means that a new sampler had to be created for the profile metric samples. Also, there are no checks currently on the sample size given because of how `schedule` is a generator type.

Ideally, it would be nice if samples could be added more easily to a `Sampler` (so we wouldn't have to define a new `Sample` subclass for every single metric we want to get percentiles from). It would also be nice if things could be added more easily to the metrics store instead of the current hard-coded version -- we could create a function to add some sort of metrics group (operational, correctness, profile-metric) to the store.

### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-benchmark/issues/886

### Testing

---

### DIFF
--- a/osbenchmark/metrics.py
+++ b/osbenchmark/metrics.py
@@ -1840,8 +1840,17 @@ class GlobalStatsCalculator:
                         self.single_latency(t, op_type, metric_name="recall@k"),
                         self.single_latency(t, op_type, metric_name="recall@1"),
                         error_rate,
-                        duration,
+                        duration
                     )
+
+                    profile_metrics = task.operation.params.get("profile-metrics", None)
+                    if profile_metrics:
+                        profile_metrics.append("query_time")
+                        result.add_profile_metrics(
+                            t,
+                            task.operation.name,
+                            {name: self.single_latency(t, op_type, metric_name=name) for name in profile_metrics}
+                        )
 
         self.logger.debug("Gathering indexing metrics.")
         result.total_time = self.sum("indexing_total_time")
@@ -2038,6 +2047,7 @@ class GlobalStats:
     def __init__(self, d=None):
         self.op_metrics = self.v(d, "op_metrics", default=[])
         self.correctness_metrics = self.v(d, "correctness_metrics", default=[])
+        self.profile_metrics = self.v(d, "profile_metrics", default=[])
         self.total_time = self.v(d, "total_time")
         self.total_time_per_shard = self.v(d, "total_time_per_shard", default={})
         self.indexing_throttle_time = self.v(d, "indexing_throttle_time")
@@ -2125,20 +2135,24 @@ class GlobalStats:
                     })
             elif metric == "correctness_metrics":
                 for item in value:
-                    if "recall@k" in item:
-                        all_results.append({
-                            "task": item["task"],
-                            "operation": item["operation"],
-                            "name": "recall@k",
-                            "value": item["recall@k"]
-                        })
-                    if "recall@1" in item:
-                        all_results.append({
-                            "task": item["task"],
-                            "operation": item["operation"],
-                            "name": "recall@1",
-                            "value": item["recall@1"]
-                        })
+                    for knn_metric in ["recall@k", "recall@1"]:
+                        if knn_metric in item:
+                            all_results.append({
+                                "task": item["task"],
+                                "operation": item["operation"],
+                                "name": knn_metric,
+                                "value": item[knn_metric]
+                            })
+            elif metric == "profile_metrics":
+                for item in value:
+                    for metric_name in item.keys():
+                        if metric_name not in ["task, operation, error_rate, duration"]:
+                            all_results.append({
+                                "task": item["task"],
+                                "operation": item["operation"],
+                                "name": metric_name,
+                                "value": item[metric_name]
+                            })
             elif metric.startswith("total_transform_") and value is not None:
                 for item in value:
                     all_results.append({
@@ -2189,9 +2203,15 @@ class GlobalStats:
             "recall@k": recall_at_k_stats,
             "recall@1":recall_at_1_stats,
             "error_rate": error_rate,
-            "duration": duration
-            }
-            )
+            "duration": duration,
+            })
+        
+    def add_profile_metrics(self, task, operation, profile_metrics):
+        self.profile_metrics.append({
+            "task": task,
+            "operation": operation,
+            "metrics": profile_metrics
+            })
 
     def tasks(self):
         # ensure we can read test_execution.json files before OSB 0.8.0
--- a/osbenchmark/metrics.py
+++ b/osbenchmark/metrics.py
@@ -2205,7 +2205,7 @@ class GlobalStats:
             "error_rate": error_rate,
             "duration": duration,
             })
-        
+
     def add_profile_metrics(self, task, operation, profile_metrics):
         self.profile_metrics.append({
             "task": task,
--- a/osbenchmark/metrics.py
+++ b/osbenchmark/metrics.py
@@ -1809,21 +1809,21 @@ class GlobalStatsCalculator:
 
         for tasks in self.test_procedure.schedule:
             for task in tasks:
-                t = task.name
+                task_name = task.name
                 op_type = task.operation.type
-                error_rate = self.error_rate(t, op_type)
-                duration = self.duration(t)
+                error_rate = self.error_rate(task_name, op_type)
+                duration = self.duration(task_name)
 
                 if task.operation.include_in_results_publishing or error_rate > 0:
-                    self.logger.debug("Gathering request metrics for [%s].", t)
+                    self.logger.debug("Gathering request metrics for [%s].", task_name)
                     result.add_op_metrics(
-                        t,
+                        task_name,
                         task.operation.name,
-                        self.summary_stats("throughput", t, op_type, percentiles_list=self.throughput_percentiles),
-                        self.single_latency(t, op_type),
-                        self.single_latency(t, op_type, metric_name="service_time"),
-                        self.single_latency(t, op_type, metric_name="client_processing_time"),
-                        self.single_latency(t, op_type, metric_name="processing_time"),
+                        self.summary_stats("throughput", task_name, op_type, percentiles_list=self.throughput_percentiles),
+                        self.single_latency(task_name, op_type),
+                        self.single_latency(task_name, op_type, metric_name="service_time"),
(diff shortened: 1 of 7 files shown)


osbenchmark/metrics.py
```diff
@@ -1840,9 +1840,18 @@ def __call__(self):
                         self.single_latency(t, op_type, metric_name="recall@k"),
                         self.single_latency(t, op_type, metric_name="recall@1"),
                         error_rate,
-                        duration,
+                        duration
                     )
 
+                    profile_metrics = task.operation.params.get("profile-metrics", None)
+                    if profile_metrics:
+                        profile_metrics.append("query_time")
```
OVI3D0, line 1848: do you foresee any consequences to directly mutating the users' workload JSON files?

cc @IanHoang I think you mentioned something related to this in our sync

osbenchmark/metrics.py
```diff
@@ -1840,9 +1840,18 @@ def __call__(self):
...
+                        result.add_profile_metrics(
+                            t,
```
OVI3D0: nit: `t` is pretty opaque, I see the other methods do the same thing but maybe we can specify `task_name` instead of just `t` in these method calls?

osbenchmark/metrics.py
```diff
@@ -2125,20 +2135,24 @@ def op_metrics(op_item, key, single_value=False):
...
+                            all_results.append({
+                                "task": item["task"],
+                                "operation": item["operation"],
+                                "name": knn_metric,
+                                "value": item[knn_metric]
+                            })
+            elif metric == "profile_metrics":
+                for item in value:
+                    for metric_name in item.keys():
+                        if metric_name not in ["task, operation, error_rate, duration"]:
```
OVI3D0: I assume you mean
`["task", "operation", "error_rate", "duration"]`
instead of one big string?

osbenchmark/worker_coordinator/runner.py
```diff
@@ -1167,6 +1167,36 @@ async def __call__(self, opensearch, params):
         # disable eager response parsing - responses might be huge thus skewing results
         opensearch.return_raw_response()
 
+        def get_profile_metrics(response_json: Dict[str, Any], metrics: List) -> Dict[str, int]:
```
OVI3D0: should we change the return type to `Dict[str, float]` since you divide by `1e6` at the end?

osbenchmark/worker_coordinator/runner.py
```diff
@@ -1219,30 +1249,37 @@ async def _search_after_query(opensearch, params):
...
-                }
-            else:
-                return {
-                    "weight": 1,
-                    "unit": "ops",
-                    "success": True
-                }
+                })
+
+            if profile:
```
OVI3D0: curious, what's the difference between here and the `metric_timings` we add below in lines 1547-1550? It seems in that case you add metric_timings as a key-value pair like `result.update({"profile-metrics": metric_timings})`, but here you just do `result.update(metric_timings)`

osbenchmark/worker_coordinator/runner.py
```diff
@@ -1504,6 +1544,10 @@ def _get_should_calculate_recall(params: dict) -> bool:
                 candidates.append(field_value)
             neighbors_dataset = params["neighbors"]
 
+            if profile:
+                metric_timings = get_profile_metrics(response_json, params.get("profile-metrics"))
+                result.update({"profile-metrics": metric_timings})
```
OVI3D0: nit: I know it's only 3 lines, but it seems like we do the same thing at line 1278. Maybe we can extract this into a small helper method?

osbenchmark/worker_coordinator/runner.py
```diff
@@ -1167,6 +1167,36 @@ async def __call__(self, opensearch, params):
...
+            shards = response_json['profile']['shards']
+            for shard in shards:
+                searches = shard['searches']
+                for search in searches:
+                    queries = search['query']
+                    for query in queries:
+                        metric_timings["query_time"] += query["time_in_nanos"]
+                        _get_query_timings(query)
+            metric_timings = {key : value / 1e6 for key, value in metric_timings.items()}
+            return metric_timings
```
OVI3D0: I think it's worth adding some exception handling here incase we get an unexpected format and we get a key error for example. We can log/print an error message and silently return a 0 filled dict instead of potentially blowing the test up

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -986,6 +989,9 @@ def prepare_benchmark(self, t):
                                                          downsample_factor,
                                                          self.workload.meta_data,
                                                          self.test_procedure.meta_data)
+        self.profile_metrics_post_processor = ProfileMetricsSamplePostprocessor(self.metrics_store,
```
OVI3D0: let's only create this postprocessor if profiling is enabled

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1245,12 +1251,14 @@ def close(self):
...
 
-    def update_samples(self, samples):
+    def update_samples(self, samples, profile_samples):
         if len(samples) > 0:
             self.raw_samples += samples
             # We need to check all samples, they will be from different clients
             for s in samples:
                 self.most_recent_sample_per_client[s.client_id] = s
+        if len(profile_samples) > 0:
+            self.raw_profile_samples += profile_samples
```
OVI3D0: I think it's better if we separate this out into `update_samples` and `update_profile_samples`

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1414,6 +1449,76 @@ def merge(self, *args):
...
+        self.metrics_store = metrics_store
+        self.workload_meta_data = workload_meta_data
+        self.test_procedure_meta_data = test_procedure_meta_data
+
+    def __call__(self, raw_samples):
+        if len(raw_samples) == 0:
+            return
+        total_start = time.perf_counter()
+        start = total_start
+        final_sample_count = 0
```
OVI3D0, line 1455: `final_sample_count` is never incremented so the debug will always log `0`

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1414,6 +1449,76 @@ def merge(self, *args):
...
+        self.logger.debug("Flushing the metrics store took [%f] seconds.", (end - start))
+        self.logger.debug("Postprocessing [%d] raw samples (downsampled to [%d] samples) took [%f] seconds in total.",
+                          len(raw_samples), final_sample_count, (end - total_start))
+
+    def merge(self, *args):
+        result = {}
+        for arg in args:
+            if arg is not None:
+                result.update(arg)
+        return result
```
OVI3D0: this method is identical to the `merge()` method in the normal SamplePostProcessor, maybe we can extract it into a shared helper?

osbenchmark/metrics.py
IanHoang: +1

IanHoang:
@neuenfeldttj Can you add unittests for this?

osbenchmark/worker_coordinator/runner.py
neuenfeldttj: That was a mistake. I'll fix and double check in unit tests

osbenchmark/worker_coordinator/worker_coordinator.py
neuenfeldttj: I agree. Do you have suggestions on how we should get that information? I was thinking the easiest thing to do would be to make a flag but that would have to occur in test_procedures but then again it's only dependent upon Search tasks.

osbenchmark/metrics.py
neuenfeldttj: This could be concerning just from the fact that a user could have that exact same key. We can remove it or keep it. I figured it would be nice to see the true query time from the profiler's perspective (no latencies).

osbenchmark/worker_coordinator/runner.py
```diff
@@ -1167,6 +1170,45 @@ async def __call__(self, opensearch, params):
         # disable eager response parsing - responses might be huge thus skewing results
         opensearch.return_raw_response()
 
+        def add_profile_to_results(response_json: Dict[str, Any], params: dict, result: dict):
```
OVI3D0, line 1173: curious, can we get any more specific than `Any`?

neuenfeldttj: I don't think so? The response output has values from an int/float, str, array, or object

osbenchmark/worker_coordinator/worker_coordinator.py
OVI3D0: since we modify the test procedure in `runner.py` you can probably check the test procedure for `profile-metrics` or `query_time`

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1320,8 +1346,8 @@ def __call__(self, raw_samples):
                         )
 
                         self.metrics_store.put_value_cluster_level(
-                            name=recall_metric_name,
-                            value=sample.request_meta_data[recall_metric_name],
+                            name=knn_metric_name,
+                            value=sample.request_meta_data[knn_metric_name],
```
OVI3D0: why the name changes here?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1700,60 +1780,69 @@ def __init__(self, start_timestamp, buffer_size=16384):
...
         except queue.Full:
             self.logger.warning("Dropping sample for [%s] due to a full sampling queue.", task.operation.name)
 
-    @property
-    def samples(self):
-        samples = []
+class ProfileMetricsSampler(Sampler):
+    """
+    Encapsulates management of gathered profile metrics samples.
+    """
```
OVI3D0, line 1823: can you add docstrings to the `DefaultSampler`/`DefaultSample`/`ProfileMetricsSample` as well so future contributors can quickly differentiate between them?

OVI3D0:
left a couple small comments but other than that LGTM

IanHoang (approved):

PR #922, opened 2025-08-07
Initial version of the Ingestion Manager for streaming ingestion

### LINKED ISSUE #896: [Data Streaming] Implement Ingestion Manager
### Ingestion Manager

The Ingestion Manager is primarily comprised of a throttling controller that will manage the rate at which the Data Partitioner downloads data from the Data Producer. If clients are backlogged with ingestion into the target cluster for the chunks they are currently processing, no additional chunks will be downloaded beyond a high watermark, thereby throttling the Data Partitioner and consequently, the Data Producer.

The Ingestion Manager also assigns chunks to clients as they free up from ingestion of data from their previously assigned chunks. It keeps track of chunks in use and the clients that are using those chunks. One consideration is when clients run into errors, fail or get suspended by the redline mechanism. In the latter case, clients will be expected to suspend only after processing their allocated chunk. In a later implementation, the Ingestion Manager will periodically perform health checks on clients and terminate ones that still have an active chunk of documents to process, but do not respond. In the case of a failed client, the Ingestion Manager will reassign its chunk to the next active client that requests chunks, with the caveat that some bulks may be re-ingested. This should be an acceptable solution for a long-running stream.

More details are in the RFC #789.

gkamat:
### Description
Initial version of the Ingestion Manager for streaming ingestion.  Inactive currently, and will be integrated with the Data Producer/Partitioner when the Data Reader changes are completed.

### Issues Resolved
#896 

### Testing
- [X] Unit tests
- [ ] Integration tests will be added with Data Reader

---


osbenchmark/workload/ingestion_manager.py
```diff
@@ -0,0 +1,13 @@
+
```
IanHoang, line 8: Add license?

IanHoang (approved):
LGTM as interim changes

Issue #930, opened 2025-08-11
[SDG] Add Pydantic Support


IanHoang:
### Is your feature request related to a problem? Please describe

Currently, SDG before OSB 2.0 release, has minimum ways of ensuring that the yaml provided adheres to a schema that SDG expects. 

Minor defects in sdg-config.yml that the user provides can cause issues for developers and make it difficult to debug.

### Describe the solution you'd like

We should incorporate a well-backed library like Pydantic that can ensure strong data validation while preserving the current architecture. This should be a simple swap in and replace. 

By doing this, we'll make SDG more robust, have better error handling, and better extensibility for future config development.

PR #931, opened 2025-08-11
Add Pydantic to SDG for stronger validation and extensibility

IanHoang:
### Description
This PR adds pydantic to OSB's SDG feature that will debut with OSB 2.0. Pydantic ensures that data conforms to the format OSB expects before it uses it to generate data. It uses type hints to enforce data validation at runtime and provides robust error handling. This will overall make SDG more robust, prevent bugs, and make it easier to add new configurations that users can toggle with. 

### What's changed?
Not much has changed. I've renamed `types.py` to `models.py`. This file originally contained just a dictionary containing default values for generation settings and a dataclass that tracks SDG metadata. The dataclass has been converted to a pydantic model and there are new models that relate to the sdg-config.yml file users provide. This limits the chance of users introducing subtle bugs.

Ran several E2E tests with different scenarios. SDG performance remains the same. This is an interim PR that lays strong foundation for subsequent PRs related to user sdg-config changes. 

### Issues Resolved
https://github.com/opensearch-project/opensearch-benchmark/issues/930

### Testing

- [ ] Tested with custom sdg with config
- [ ] Tested with custom sdg without config
- [ ] Tested with basic mappings without config
- [ ] Tested with basic mappings with config
- [ ] Tested with complex mappings without config
- [ ] Tested with complex mappings with config
- [ ] Tested rollovers 

---


.pylintrc
```diff
@@ -173,7 +173,15 @@ disable=print-statement,
...
-        raise-missing-from
+        raise-missing-from,
+        R1732,
+        R0801,
+        R0402,
+        E0213,
+        E0012,
+        C0206,
+        W0238,
+        R1729
```
gkamat: Would help to use the text strings here, since they are more readable.

osbenchmark/synthetic_data_generator/helpers.py
```diff
@@ -98,18 +98,26 @@ def host_has_available_disk_storage(sdg_metadata: SyntheticDataGeneratorMetadata
...
+            raise exceptions.ConfigError(f"User provided config with extension [{extension}]. Config must have a {allowed_extensions} extension.")
         else:
             with open(config_path, 'r') as file:
-                return yaml.safe_load(file)
+                config_details = yaml.safe_load(file)
+
+            if config_details is None:
+                return SDGConfig()
+
+            return SDGConfig(**config_details)
```
gkamat:
```
return SDGConfig(**config_details) if config_details else SDGConfig()
```

osbenchmark/synthetic_data_generator/models.py
```diff
@@ -0,0 +1,129 @@
...
+import re
+
+from pydantic import BaseModel, Field, field_validator
+
+GB_TO_BYTES = 1024 ** 3
+
+class SettingsConfig(BaseModel):
+    workers: Optional[int] = Field(default_factory=os.cpu_count)
+    max_file_size_gb: Optional[int] = 40
+    docs_per_chunk: Optional[int] = 10000
```
gkamat: Might want to document the reason these values were chosen.

osbenchmark/synthetic_data_generator/models.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @field_validator('workers', 'max_file_size_gb', 'docs_per_chunk')
+    def validate_values_are_positive_integers(cls, v):
+        if v is not None and v <= 0:
+            raise ValueError('Value in Settings portion must be a positive integer.')
```
gkamat: Printing out the value of `v` will help the user track the reason for the error better.

osbenchmark/synthetic_data_generator/models.py
```diff
@@ -0,0 +1,129 @@
...
+class CustomGenerationValuesConfig(BaseModel):
+    custom_lists: Optional[Dict[str, List[Any]]] = None
+    custom_providers: Optional[List[Any]] = None
+
+    @field_validator('custom_lists')
+    def validate_custom_lists(cls, v):
+        if v is not None:
+            for key, value in v.items():
+                if not isinstance(key, str):
+                    raise ValueError('All keys within custom_lists of CustomGenerationValues section must be strings')
```
gkamat: Printing out the offending key will be helpful.

osbenchmark/synthetic_data_generator/models.py
```diff
@@ -0,0 +1,129 @@
...
+        return v
+
+class MappingGenerationValuesConfig(BaseModel):
+    generator_overrides: Optional[Dict[str, GeneratorParams]] = None
+    field_overrides: Optional[Dict[str, FieldOverride]] = None
+
+    @field_validator('generator_overrides')
+    def validate_generator_types(cls, v):
+        if v is not None:
+            valid_generator_types = ['integer', 'long', 'float', 'double', 'date', 'text', 'keyword']
```
gkamat: Should `short`,`byte`, etc. be in here too?

osbenchmark/synthetic_data_generator/models.py
```diff
@@ -0,0 +1,129 @@
...
+            'generate_text',
+            'generate_keyword',
+            'generate_integer',
+            'generate_long',
+            'generate_short',
+            'generate_byte',
+            'generate_float',
+            'generate_double',
+            'generate_boolean',
+            'genearte_date',
```
gkamat: `generate`

osbenchmark/synthetic_data_generator/models.py
```diff
@@ -0,0 +1,129 @@
...
+                    raise ValueError(f"Invalid Generator Type '{generator_type}. Must be one of: {valid_generator_types}'")
+
+        return v
+
+    @field_validator('field_overrides')
+    def validate_field_names(cls, v):
+        if v is not None:
+            for field_name in v.keys():
+                if not re.match(r'^[a-zA-Z][a-zA-Z0-9_.]*$', field_name):
+                    raise ValueError(f"Invalid Field Name in FieldOverrides: {field_name}")
```
gkamat: literal can contain only alphanumeric characters, underscores and periods.

osbenchmark/synthetic_data_generator/strategies/custom_module_strategy.py
```diff
@@ -16,29 +16,38 @@
...
+                self.custom_lists = self.sdg_config.CustomGenerationValues.custom_lists or {}
+                provider_names = self.sdg_config.CustomGenerationValues.custom_providers or []
+                self.custom_providers = {
+                    name: getattr(self.custom_module, name) for name in provider_names
+                }
+            except AttributeError as e:
+                msg = f"Error when setting up custom lists and custom providers: {e}"
+                raise exceptions.ConfigError(msg)
+            except TypeError:
+                msg = "Synthetic Data Generator Config has custom_lists and custom_providers pointing to null values. Either populate or remove."
```
gkamat, line 49: Can `TypeError` be raised in cases other than `null` values, such as invalid values?

setup.py
```diff
@@ -122,7 +122,9 @@ def str_from_file(name):
     # Licence: BSD-3-Clause
     "dask[distributed]",
     # Licence: BSD-3-Clause
-    "bokeh!=3.0.*,>=2.4.2"
+    "bokeh!=3.0.*,>=2.4.2",
+    # License: MIT
+    "pydantic"
```
gkamat: Should this be pinned to a particular max version, which would get updated periodically, but only after testing?

gkamat (changes requested):

setup.py
IanHoang: That's a good idea. I'll add a pin for pydantic and pydantic_core as I was using those during my testing and development

osbenchmark/synthetic_data_generator/models.py
IanHoang: Good call out. I plan to open a subsequent PR for adding more complications to Mapping SDG but will update this here. I'll add all except for object.

osbenchmark/synthetic_data_generator/helpers.py
IanHoang: Much cleaner

osbenchmark/synthetic_data_generator/strategies/custom_module_strategy.py
IanHoang: Upon quick lookup, it looks like it can. But the models for custom_lists and custom_providers should add extra validation.

gkamat (approved):

Issue #918, opened 2025-08-05
[Data Streaming] Implement Data Reader


gkamat:
### Data Reader

The Data Reader is the OSB component that orchestrates how data is supplied to the clients. Currently, it operates based on locally-stored data corpora files and their associated file offset tables. The offset tables enable the parallel clients to seek into a corpus file and locate the point from which to begin ingestion of documents for the respective partition allocated to them.

The Data Reader will be enhanced to operate with distinct chunk files for each client, in addition to the existing functionality. Each client will operate independently on the chunk file assigned to it by the Ingestion Manager and will delete the file when ingestion of the file is complete. At that point, it will request the Ingestion Manager to assign it the next chunk. There is substantial glue logic that will need to be modified and added in the OSB codebase to support the chunk functionality. Much of this is in the data reader hierarchy, slicing and bulk generator components and will be elaborated further in the implementation design for this component. The current data corpus files and the offset file scheme will be subsumed into the new workflow, so they will continue to work for backward compatibility.

More details are in the RFC #789

Issue #929, opened 2025-08-11
[Streaming Ingestion] Performance and scaling tests


gkamat:
### Description

The [Streaming Ingestion](#850) feature is intended to enable OSB  to scale up to a target throughput of 10 TB/day.  This issue will track performance achieved with various load generation instance types and target cluster capabilities, including serverless implementations.  This will help provide appropriate guidance to users and stakeholders.

PR #935, opened 2025-08-16
Initial set of Data Reader changes for Streaming Ingestion

gkamat:
### Description
Initial set of Data Reader changes for Streaming Ingestion.  This is an interim check-in to facilitate the availability of the feature in OSB 2.0 as an alpha capability.  There are several enhancements to be added yet; these will be made available progressively after the release.  Integration tests are also to be added.

For now, the focus has been to ensure that OSB functionality, when Streaming Ingestion is not enabled, is not affected. 

### Issues Addressed
#918 

### Testing
- [ ] Unit and integ tests

---

### DIFF
--- a/osbenchmark/cloud_provider/vendors/s3_data_producer.py
+++ b/osbenchmark/cloud_provider/vendors/s3_data_producer.py
@@ -15,8 +15,6 @@ from boto3 import client
 
 from osbenchmark import exceptions
 from osbenchmark.data_streaming.data_producer import DataProducer
-from osbenchmark.cloud_provider import CloudProviderFactory
-from osbenchmark.cloud_provider.vendors.aws import AWSProvider
 from osbenchmark.workload.ingestion_manager import IngestionManager
 
 class S3DataProducer(DataProducer):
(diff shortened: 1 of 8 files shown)


osbenchmark/cloud_provider/vendors/s3_data_producer.py
```diff
@@ -39,42 +37,21 @@ def __init__(self, bucket:str, keys, client_options: dict, data_dir=None) -> Non
...
-                self.s3_client = client('s3',
-                                   aws_access_key_id = provider.aws_log_in_config['aws_access_key_id'],
-                                   aws_secret_access_key = provider.aws_log_in_config['aws_secret_access_key'],
-                                   aws_session_token = provider.aws_log_in_config['aws_session_token'],
-                                   )
-            else:
-                # For testing.  Set credentials with environment variables.
-                self.s3_client = client('s3')
+            self.chunk_size = IngestionManager.chunk_size * 1024**2
+            self.num_workers = os.cpu_count() * 2
```
IanHoang, line 41: Will we always want 2x the number of CPU cores or will we ever want to use less?

osbenchmark/workload/loader.py
```diff
@@ -412,6 +415,8 @@ def __init__(self, cfg):
 
     @staticmethod
     def prepare_docs(cfg, workload, corpus, preparator):
+        if corpus.streaming_ingestion:
+            return
```
IanHoang, line 419: Nit: A logging statement here would be useful. Can mention that workload loader detected that user is using streaming ingestion and from which cloud provider (only aws for now)

osbenchmark/workload/params.py
```diff
@@ -776,8 +777,7 @@ def __init__(self, corpora, batch_size, bulk_size, ingest_percentage, id_conflic
         # use a value > 0 so percent_completed returns a sensible value
         self.total_bulks = 1
         self.infinite = False
-        # TBD: obtain value from workload spec.
-        self.streaming_ingestion = False
+        self.streaming_ingestion = corpora[0].streaming_ingestion
```
IanHoang, line 780: With the index 0, will streaming ingestion always only have one corpora?

osbenchmark/workload/params.py
```diff
@@ -1756,55 +1761,63 @@ def __next__(self):
...
-    # Interim function: will be updated in the next set of changes.
     @staticmethod
     def _start_producer():
         client_options_obj = IngestionManager.config.opts("client", "options")
         client_options = getattr(client_options_obj, "all_client_options", {})
-        keys = [ "sorted-http-logs-regenerated_0.json", "sorted-http-logs-regenerated_1.json", "sorted-http-logs-regenerated_2.json" ]
         # pylint: disable = import-outside-toplevel
         from osbenchmark.cloud_provider.vendors.s3_data_producer import S3DataProducer
-        producer = S3DataProducer("big5-corpus", keys, client_options, Slice.data_dir)
+        bucket = re.sub('^s3://', "", Slice.base_url)
```
IanHoang, line 1790: Understand that this is still the initial set of changes for the MVP but what are your thoughts on moving the instantiation of the producer outside of the `_start_producer()`? We could have a separate method dedicated to setting up the producer -- such as `_setup_data_producer()` -- based on the `self.streaming_ingestion` detected in the workload. After that, the producer can be passed into `_start_producer()`. That way, the main workflow will be preserved and we can support other types of data producesr later (e.g. `azure`, `gcp`, etc.)

osbenchmark/cloud_provider/vendors/s3_data_producer.py
gkamat: That is a good question, given that the cores are used for the bulk operations as well.  This is one of the items to be evaluated during the scaling experiments #929, but for now the feature seems to perform generally well with this number.

osbenchmark/workload/loader.py
gkamat: The change needs to have logging added all through (and error-checking), but will include this for clarity.

osbenchmark/workload/params.py
gkamat: Yes, it would not help dividing the bandwidth allocated to the download from cloud storage across multiple document sources, rather than optimizing the speed for the single corpus.  We could look into this as a later improvement, perhaps.

osbenchmark/workload/params.py
gkamat: That would certainly be the right way of modularizing this.  For now, the focus was on getting the feature in prior to the 2.0 release deadline, end-to-end.

PR #938, opened 2025-08-19
Add support for gRPC transport

finnegancarroll:
### Description

Changes provide support for gRPC transport operations in OSB:
- Introduces protobuf runners responsible for parsing params into proto request and response.
- Introduces a unified client which mirrors REST client functionality with additional providers for gRPC stubs
- Introduces instrumentation for gRPC to measure request start/end at the network level (before any client serialization is executed)
- Introduces flag to specify host address for gRPC endpoint

Note that these changes depend on several upstream changes:

1. Schema definitions are taken from opensearch protobufs 0.10.0. To work out of the box OpenSearch core needs to update its server side definitions to this version. See pending PR for this change: https://github.com/opensearch-project/OpenSearch/pull/19007
2. Python library for opensearch-protobufs is currently built and installed manually. Please see this pending PR to publish these libraries to PyPI: https://github.com/opensearch-project/opensearch-protobufs/pull/183
3. No workload currently utilizes gRPC. Please see pending PR in workloads repository to add a gRPC alternative to the big5 workload: https://github.com/opensearch-project/opensearch-benchmark-workloads/pull/689

### Running gRPC benchmarks

To run OSB against the gRPC endpoint please see `transport-grpc` settings for enabling this endpoint in settings:
https://github.com/opensearch-project/OpenSearch/tree/main/modules/transport-grpc

Clone the OSB workloads feature branch and provide it with the `--workload-path`:
https://github.com/opensearch-project/opensearch-benchmark-workloads/pull/689

```
opensearch-benchmark run \
    --pipeline=benchmark-only \
    --workload-path="<repo-path>/opensearch-benchmark-workloads/big5" \
    --target-host=http://localhost:9200 \
    --grpc-target-hosts=http://localhost:9400 \
    --workload-params '{"number_of_shards":"1","number_of_replicas":"0", "ingest_percentage":"1.0"}' \
    --distribution-version="3.3.0" \
    --kill-running-processes

```

With example output:

```
------------------------------------------------------
    _______             __   _____
   / ____(_)___  ____ _/ /  / ___/_________  ________
  / /_  / / __ \/ __ `/ /   \__ \/ ___/ __ \/ ___/ _ \
 / __/ / / / / / /_/ / /   ___/ / /__/ /_/ / /  /  __/
/_/   /_/_/ /_/\__,_/_/   /____/\___/\____/_/   \___/
------------------------------------------------------

|                                                         Metric |              Task |       Value |   Unit |
|---------------------------------------------------------------:|------------------:|------------:|-------:|
|                     Cumulative indexing time of primary shards |                   |     6.97163 |    min |
|             Min cumulative indexing time across primary shards |                   |     6.97163 |    min |
|          Median cumulative indexing time across primary shards |                   |     6.97163 |    min |
|             Max cumulative indexing time across primary shards |                   |     6.97163 |    min |
|            Cumulative indexing throttle time of primary shards |                   |           0 |    min |
|    Min cumulative indexing throttle time across primary shards |                   |           0 |    min |
| Median cumulative indexing throttle time across primary shards |                   |           0 |    min |
|    Max cumulative indexing throttle time across primary shards |                   |           0 |    min |
|                        Cumulative merge time of primary shards |                   |     0.41035 |    min |
|                       Cumulative merge count of primary shards |                   |          16 |        |
|                Min cumulative merge time across primary shards |                   |     0.41035 |    min |
|             Median cumulative merge time across primary shards |                   |     0.41035 |    min |
|                Max cumulative merge time across primary shards |                   |     0.41035 |    min |
|               Cumulative merge throttle time of primary shards |                   |    0.111683 |    min |
|       Min cumulative merge throttle time across primary shards |                   |    0.111683 |    min |
|    Median cumulative merge throttle time across primary shards |                   |    0.111683 |    min |
|       Max cumulative merge throttle time across primary shards |                   |    0.111683 |    min |
|                      Cumulative refresh time of primary shards |                   |   0.0752167 |    min |
|                     Cumulative refresh count of primary shards |                   |          22 |        |
|              Min cumulative refresh time across primary shards |                   |   0.0752167 |    min |
|           Median cumulative refresh time across primary shards |                   |   0.0752167 |    min |
|              Max cumulative refresh time across primary shards |                   |   0.0752167 |    min |
|                        Cumulative flush time of primary shards |                   |   0.0570333 |    min |
|                       Cumulative flush count of primary shards |                   |           3 |        |
|                Min cumulative flush time across primary shards |                   |   0.0570333 |    min |
|             Median cumulative flush time across primary shards |                   |   0.0570333 |    min |
|                Max cumulative flush time across primary shards |                   |   0.0570333 |    min |
|                                        Total Young Gen GC time |                   |        4.29 |      s |
|                                       Total Young Gen GC count |                   |        1991 |        |
|                                          Total Old Gen GC time |                   |           0 |      s |
|                                         Total Old Gen GC count |                   |           0 |        |
|                                                     Store size |                   |    0.249005 |     GB |
|                                                  Translog size |                   | 5.12227e-08 |     GB |
|                                         Heap used for segments |                   |           0 |     MB |
|                                       Heap used for doc values |                   |           0 |     MB |
|                                            Heap used for terms |                   |           0 |     MB |
|                                            Heap used for norms |                   |           0 |     MB |
|                                           Heap used for points |                   |           0 |     MB |
|                                    Heap used for stored fields |                   |           0 |     MB |
|                                                  Segment count |                   |           8 |        |
|                                                 Min Throughput | grpc-index-append |     20314.5 | docs/s |
|                                                Mean Throughput | grpc-index-append |       23556 | docs/s |
|                                              Median Throughput | grpc-index-append |     22851.4 | docs/s |
|                                                 Max Throughput | grpc-index-append |     40423.3 | docs/s |
|                                        50th percentile latency | grpc-index-append |     183.964 |     ms |
|                                        90th percentile latency | grpc-index-append |     250.861 |     ms |
|                                        99th percentile latency | grpc-index-append |     409.044 |     ms |
|                                      99.9th percentile latency | grpc-index-append |      801.04 |     ms |
|                                       100th percentile latency | grpc-index-append |     848.214 |     ms |
|                                   50th percentile service time | grpc-index-append |     183.964 |     ms |
|                                   90th percentile service time | grpc-index-append |     250.861 |     ms |
|                                   99th percentile service time | grpc-index-append |     409.044 |     ms |
|                                 99.9th percentile service time | grpc-index-append |      801.04 |     ms |
|                                  100th percentile service time | grpc-index-append |     848.214 |     ms |
|                                                     error rate | grpc-index-append |           0 |      % |
|                                                 Min Throughput |    grpc-match-all |        2.01 |  ops/s |
|                                                Mean Throughput |    grpc-match-all |        2.01 |  ops/s |
|                                              Median Throughput |    grpc-match-all |        2.01 |  ops/s |
|                                                 Max Throughput |    grpc-match-all |        2.01 |  ops/s |
|                                        50th percentile latency |    grpc-match-all |     7.13467 |     ms |
|                                        90th percentile latency |    grpc-match-all |     10.6535 |     ms |
|                                        99th percentile latency |    grpc-match-all |      11.535 |     ms |
|                                       100th percentile latency |    grpc-match-all |     17.7223 |     ms |
|                                   50th percentile service time |    grpc-match-all |     5.52638 |     ms |
|                                   90th percentile service time |    grpc-match-all |     8.48873 |     ms |
|                                   99th percentile service time |    grpc-match-all |     9.38595 |     ms |
|                                  100th percentile service time |    grpc-match-all |     15.5186 |     ms |
|                                                     error rate |    grpc-match-all |           0 |      % |
|                                                 Min Throughput |         grpc-term |        2.01 |  ops/s |
|                                                Mean Throughput |         grpc-term |        2.01 |  ops/s |
|                                              Median Throughput |         grpc-term |        2.01 |  ops/s |
|                                                 Max Throughput |         grpc-term |        2.01 |  ops/s |
|                                        50th percentile latency |         grpc-term |     6.22912 |     ms |
|                                        90th percentile latency |         grpc-term |     9.92944 |     ms |
|                                        99th percentile latency |         grpc-term |     14.6624 |     ms |
|                                       100th percentile latency |         grpc-term |     14.7964 |     ms |
|                                   50th percentile service time |         grpc-term |     4.49331 |     ms |
|                                   90th percentile service time |         grpc-term |      7.2344 |     ms |
|                                   99th percentile service time |         grpc-term |     12.2952 |     ms |
|                                  100th percentile service time |         grpc-term |     12.3938 |     ms |
|                                                     error rate |         grpc-term |           0 |      % |


-----------------------------------
[INFO] ✅ SUCCESS (took 418 seconds)
-----------------------------------
```

### Issues Resolved
N/A

### Testing
~~- [ ] New functionality includes testing~~

Tested manually.

---

### DIFF
--- a/osbenchmark/worker_coordinator/proto_helpers/ProtoQueryHelper.py
+++ b/osbenchmark/worker_coordinator/proto_helpers/ProtoQueryHelper.py
@@ -1,12 +1,16 @@
 from opensearch.protobufs.schemas import search_pb2
 from opensearch.protobufs.schemas import common_pb2
 
+# In some cases (KNN) we set stored fields explicitly to "_none_" to disable
+# https://github.com/opensearch-project/OpenSearch/blob/3.3/server/src/main/java/org/opensearch/search/fetch/StoredFieldsContext.java#L59
+STORED_FIELDS_NONE = "_none_"
+
 def is_true(value):
     if isinstance(value, str):
         return value.lower() == "true"
     return bool(value)
 
-def _get_relation(relation):
+def get_relation(relation):
     match relation:
         case 0:
             return "TOTAL_HITS_RELATION_UNSPECIFIED"
@@ -17,7 +21,7 @@ def _get_relation(relation):
         case _:
             return "TOTAL_HITS_RELATION_UNSET"
 
-def _get_terms_dict(query):
+def get_terms_dict(query):
     terms = {}
     for key, value in query.items():
         terms[key] = []
@@ -35,7 +39,7 @@ class ProtoQueryHelper:
     # Term query supports a single term on single field.
     @staticmethod
     def term_query_to_proto(query):
-        term = _get_terms_dict(query)
+        term = get_terms_dict(query)
         if len(term.keys()) > 1:
             raise Exception("Error parsing query - Term query contains multiple distinct fields: " + str(query))
         if len(term.values()) > 1:
@@ -136,11 +140,11 @@ class ProtoQueryHelper:
         source_config = common_pb2.SourceConfigParam(bool=fetch_source)
         timeout = params.get("request-timeout")
 
-        if body.get("stored_fields") is None or "_none_":
-            stored_fields = ["_none_"]
+        if body.get("stored_fields") is None or body.get("stored_fields") is STORED_FIELDS_NONE:
+            stored_fields = [STORED_FIELDS_NONE]
         else:
             stored_fields = body.get("stored_fields")
-        
+
         if isinstance(params.get("cache"), bool):
             cache = params.get("cache")
         elif isinstance(params.get("cache"), str):
@@ -189,7 +193,7 @@ class ProtoQueryHelper:
                 "unit": "ops",
                 "success": True,
                 "hits": response.hits.total.total_hits.value,
-                "hits_relation": _get_relation(response.hits.total.total_hits.relation),
+                "hits_relation": get_relation(response.hits.total.total_hits.relation),
                 "timed_out": response.timed_out,
                 "took": response.took,
             }


finnegancarroll:
Big5 term query.
Minor decrease in latency, ~3.5ms to ~3.4ms with gRPC/protobuf.

image

finnegancarroll:
Note: These previous latency benchmarks are not accurate. Previously we were measuring serialization time for the gRPC client but not the REST client. Moved gRPC request start/end timers to execute in an intercepter such that we are measuring the time closer to the network. Throughput calculations remain unchanged.

osbenchmark/benchmark.py
```diff
@@ -587,6 +587,11 @@ def add_workload_source(subparser):
         "--worker-ips",
         help="Define a comma-separated list of hosts which should generate load (default: localhost).",
         default="localhost")
+    test_run_parser.add_argument(
+        "--grpc-target-hosts",
+        help="Define a comma-separated list of host:port pairs for gRPC endpoints "
+             "(default: localhost:9400).",
+        default="")
```
gkamat, line 594: What is the intention for supporting multiple host:port pairs?  Will they be handled similarly to the existing target hosts?  Are the gRPC runners going to be targeting multiple clusters concurrently?

osbenchmark/client.py
```diff
@@ -167,9 +175,11 @@ def create(self):
         if self.provider:
             self.logger.info("Creating OpenSearch client with provider %s", self.provider)
             return self.provider.create_client(self.hosts, self.client_options)
-
         else:
-            return opensearchpy.OpenSearch(hosts=self.hosts, ssl_context=self.ssl_context, **self.client_options)
+            import opensearchpy
+            class BenchmarkOpenSearch(opensearchpy.OpenSearch, RequestContextHolder):
+                pass
+            return BenchmarkOpenSearch(hosts=self.hosts, ssl_context=self.ssl_context, **self.client_options)
```
gkamat: Why is this class not at the top level like the other one that was moved there (`BenchmarkAsyncOpenSearch`)?  OSB now uses only the async interface AFAIK, is there some reason the earlier interface needs to be resuscitated?

osbenchmark/client.py
```diff
@@ -278,3 +285,134 @@ async def create(params):
...
+        """
+        # pylint: disable=import-outside-toplevel
+        import grpc
+        from opensearch.protobufs.services.document_service_pb2_grpc import DocumentServiceStub
+        from opensearch.protobufs.services.search_service_pb2_grpc import SearchServiceStub
+        
+        if not self.grpc_hosts:
+            return {}
+            
+        stubs = {}
```
gkamat: Might be clearer to return `stubs` just once at the end of the function, and set it accordingly here.

osbenchmark/client.py
```diff
@@ -278,3 +285,134 @@ async def create(params):
...
+
+        for cluster_name, hosts in self.grpc_hosts.all_hosts.items():
+            if hosts:
+                # Just use the first host - No load balancing support
+                host = hosts[0]
```
gkamat: Probably easier to understand if the hosts for the 'default' key are selected, and the first host selected from that cluster.  Is there some reason the load balancing is skipped here?

osbenchmark/client.py
```diff
@@ -278,3 +285,134 @@ async def create(params):
...
+                grpc_addr = f"{host['host']}:{host['port']}"
+                self.logger.info(f"Creating gRPC channel for cluster '{cluster_name}' at {grpc_addr}")
+                
+                options = [
+                    ('grpc.max_concurrent_streams', 100),
+                    ('grpc.so_reuseport', 0),
+                    ('grpc.use_local_subchannel_pool', 1),
+                    ('grpc.max_send_message_length', 10 * 1024 * 1024),  # 10 MB
+                    ('grpc.max_receive_message_length', 10 * 1024 * 1024)  # 10 MB
+                ]
```
gkamat: Would be better to have these defaults specified at the top of the module, or in a separate module.

osbenchmark/client.py
```diff
@@ -278,3 +285,134 @@ async def create(params):
...
+                channel = grpc.aio.insecure_channel(
+                    target=grpc_addr,
+                    options=options,
+                    compression=None
+                )
+
+                stubs[cluster_name] = {
+                    'document_service': DocumentServiceStub(channel),
+                    'search_service': SearchServiceStub(channel),
+                    '_channel': channel
```
gkamat: Why is only one key prefixed with an underscore?  Should all of them be so?

osbenchmark/client.py
```diff
@@ -278,3 +285,134 @@ async def create(params):
...
+
+class UnifiedClient:
+    """
+    Unified client that wraps both OpenSearch REST client and gRPC stubs.
+    This provides a single interface for runners to access both protocols.
+    Acts as a transparent proxy to the OpenSearch client while adding gRPC capabilities.
+    """
+    def __init__(self, opensearch_client, grpc_stubs=None):
+        self._opensearch = opensearch_client
+        self._grpc_stubs = grpc_stubs or {}
```
gkamat: Wouldn't 
```
self._grpc_stubs = grpc_stubs
```
do the same?

osbenchmark/client.py
```diff
@@ -278,3 +285,134 @@ async def create(params):
...
+        
+    def __getattr__(self, name):
+        """Delegate all unknown attributes to the underlying OpenSearch client."""
+        return getattr(self._opensearch, name)
+        
+    def document_service(self, cluster_name="default"):
+        """Get the gRPC DocumentService stub for the specified cluster."""
+        if cluster_name in self._grpc_stubs:
+            return self._grpc_stubs[cluster_name].get('document_service')
+        return None
```
gkamat:
```
return self._grpc_stubs[cluster_name].get('document_service') if self.has_grpc_support(cluster_name) else None
```

osbenchmark/context.py
```diff
@@ -120,19 +124,23 @@ def update_client_request_end(cls, new_client_request_end):
 
     @classmethod
     def on_client_request_start(cls):
-        cls.update_client_request_start(time.perf_counter())
+        timestamp = time.perf_counter()
+        cls.update_client_request_start(timestamp)
```
gkamat: Is this clearer than the previous version?

osbenchmark/worker_coordinator/proto_helpers/ProtoBulkHelper.py
```diff
@@ -0,0 +1,87 @@
+from opensearch.protobufs.schemas import document_pb2
+
+def _parse_docs_from_body(body):
+    lineSplitBody = body.decode('utf-8').split('\n')
+    lineList = []
+    opList = []
+    docList = []
```
gkamat: Variables typically are in snake_case in the OSB codebase.  Would be better to conform to that scheme.

osbenchmark/worker_coordinator/proto_helpers/ProtoBulkHelper.py
```diff
@@ -0,0 +1,87 @@
...
+    indexPattern = '{"index":'
```
gkamat: What would happen if the metadata line contained `'{ "index" :`?  Wouldn't it be better to assume every alternate line is a metadata line, and check accordingly?  Where are the documents coming from -- generated by OSB or from somewhere else?

osbenchmark/worker_coordinator/proto_helpers/ProtoBulkHelper.py
```diff
@@ -0,0 +1,87 @@
...
+    for lineBody in lineSplitBody:
+        lineList.append(lineBody)
+        if indexPattern in lineBody:
+            opList.append(lineBody)
+        else:
+            docList.append(lineBody)\
```
gkamat: Why the backslash at EOL?

osbenchmark/worker_coordinator/proto_helpers/ProtoBulkHelper.py
```diff
@@ -0,0 +1,87 @@
...
+    * ``index``: index name
+    """
+    @staticmethod
+    def build_proto_request(params):
+        index = params.get("index")
+        body = params.get("body")
+        doc_list = _parse_docs_from_body(body)
+        request  = document_pb2.BulkRequest()
+        request.index = index
+        # All bulk request here are index ops
```
gkamat: All bulk requests ..

osbenchmark/worker_coordinator/proto_helpers/ProtoBulkHelper.py
```diff
@@ -0,0 +1,87 @@
...
+        op_cont = document_pb2.OperationContainer()
```
gkamat: `op_container` may be a clearer name for this variable.

osbenchmark/worker_coordinator/proto_helpers/ProtoBulkHelper.py
```diff
@@ -0,0 +1,87 @@
...
+    Consumed from params dictionary:
+    * ``body``: JSON body of bulk ingest request
+    * ``index``: index name
+    """
+    @staticmethod
+    def build_proto_request(params):
+        index = params.get("index")
+        body = params.get("body")
+        doc_list = _parse_docs_from_body(body)
+        request  = document_pb2.BulkRequest()
```
gkamat: Extra space before the equals character.

osbenchmark/worker_coordinator/proto_helpers/ProtoBulkHelper.py
```diff
@@ -0,0 +1,87 @@
...
+        request.index = index
+        # All bulk request here are index ops
+        op_cont = document_pb2.OperationContainer()
+        op_cont.index.CopyFrom(document_pb2.IndexOperation())
+        for doc in doc_list:
+            requestBody = document_pb2.BulkRequestBody()
```
gkamat: snake_case request_body?

osbenchmark/worker_coordinator/proto_helpers/ProtoBulkHelper.py
```diff
@@ -0,0 +1,87 @@
...
+    * ``index``: index name
+    * ``bulk-size``: documents per bulk request
+    * ``unit``: in the case of bulk always 'ops'
+    * ``detailed-results``: return detailed results - Unsupported by proto bulk -> Raise exception
+    """
+    @staticmethod
+    def build_stats(response, params):
+        which_field = response.WhichOneof('response')
+        if which_field == 'bulk_error_response':
+            raise Exception("Server responded with error: " + str(which_field))
```
gkamat: Isn't `which_field` already a `str` here?

osbenchmark/worker_coordinator/proto_helpers/ProtoBulkHelper.py
```diff
@@ -0,0 +1,87 @@
...
+        if not isinstance(response.bulk_response_body, document_pb2.BulkResponseBody):
+            raise Exception("Unknown response proto: " + str(type(response)))
+
+        if params.get("detailed-results"):
+            raise Exception("Detailed stats not supported for proto bulk index.")
+
+        success_count = 0
+        error_count = 0
+        for item in response.bulk_response_body.items:
+            if item.index.status > 299:
```
gkamat: What does 299 signify here?  Is it a convention from the client?

osbenchmark/worker_coordinator/proto_helpers/ProtoKNNQueryHelper.py
```diff
@@ -0,0 +1,95 @@
...
+        cache = params.get("cache")
+        detailed_results = params.get("detailed-results")
+        calc_recall = params.get("calculate-recall")
+        resp_compression = params.get("response-compression-enabled")
+
+        req_params = params.get("request-params")
+        source = None
+        allow_partial_search_results = None
+        if req_params is not None:
+            source = req_params.get("_source") if "_source" in req_params else None
```
gkamat: `dict.get()` returns None if key not present.

osbenchmark/worker_coordinator/proto_helpers/ProtoKNNQueryHelper.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    Parse stats from protobuf response.
+    Consumed from params dictionary:
+    * ``detailed-results``: return detailed results, hits, took, hits_relation
+    """
+    @staticmethod
+    def build_stats(response, params):
+        which_field = response.WhichOneof('response')
+        if which_field == 'error_4xx_response' or which_field == 'error_5xx_response':
+            raise Exception("Server responded with error: " + str(which_field))
```
gkamat: The `str()` is probably not needed here.

osbenchmark/worker_coordinator/proto_helpers/ProtoQueryHelper.py
```diff
@@ -0,0 +1,145 @@
...
+from opensearch.protobufs.schemas import common_pb2
+
+def _get_relation(relation):
+    if relation == 0:
+        return "TOTAL_HITS_RELATION_UNSPECIFIED"
+    elif relation == 1:
+        return "TOTAL_HITS_RELATION_EQ"
+    elif relation == 2:
+        return "TOTAL_HITS_RELATION_GTE"
+    else:
```
gkamat: Probably better to use the `match` statement here -- we plan to drop support for Python 3.8 and 3.9 shortly.  For now, please leave it as is.

osbenchmark/worker_coordinator/proto_helpers/ProtoQueryHelper.py
```diff
@@ -0,0 +1,145 @@
...
+        return "TOTAL_HITS_RELATION_UNSET"
+
+def _get_terms_dict(query):
+    terms = {}
+    for key, value in query.items():
+        terms[key] = []
+    for key, value in query.items():
+        if type(value) is list:
+            for item in value:
+                terms[key].append(item)
```
gkamat: Would `extend()` be better here?

osbenchmark/worker_coordinator/proto_helpers/ProtoQueryHelper.py
```diff
@@ -0,0 +1,145 @@
...
+        elif type(value) is dict:
+            for ignore, term_value in value.items():
```
gkamat: Perhaps `update()` will be preferable here?

osbenchmark/worker_coordinator/proto_helpers/ProtoQueryHelper.py
```diff
@@ -0,0 +1,145 @@
...
+"""
+def _parse_term_from_query(query):
+    term = _get_terms_dict(query)
+    if len(term.keys()) > 1:
+        raise Exception("Error parsing query - Term query contains multiple distinct fields: " + str(query))
+    if len(term.values()) > 1:
+        raise Exception("Error parsing query - Term query contains multiple terms: " + str(query))
+
+    term_field = next(iter(term.keys()))
+    term_value = next(iter(term[term_field]))
```
gkamat: Are these iterating through, or just operating on a single item each?

gkamat (changes requested):
Please add appropriate unit tests and integ tests for this feature.

Some additional logging lines may be helpful for debugging.

Please convert to a regular PR when ready to be merged.

osbenchmark/client.py
finnegancarroll: Thanks. Did not realize non async usage was fully deprecated.

osbenchmark/client.py
finnegancarroll: Yea this is pretty much the same. Moving the "not configured" exceptions here as these service functions should never be called without --grpc-target-hosts.

osbenchmark/client.py
finnegancarroll: Moving this to just throw an exception here if user is trying to use a gRPC service when the cluster does not have stubs initialized.

osbenchmark/client.py
finnegancarroll: I was thinking `_channel` is a little special as it's only here so the unified client has a reference when it shuts down. It is otherwise not intended to be part of the client interface. Adding a comment for clarity.

osbenchmark/benchmark.py
finnegancarroll: It looks like the REST client passes these hosts directly to `opensearchpy` which puts them in a `connection_pool` and [fetches a new connection](https://github.com/opensearch-project/opensearch-py/blob/main/opensearchpy/transport.py#L264-L272) on each new request. This seems to select connections with [round robin](https://github.com/opensearch-project/opensearch-py/blob/main/opensearchpy/connection_pool.py#L136C1-L137C1) while additionally checking the health of connections and marking unhealthy nodes as dead.

On this initially version I was thinking to leave `--grpc-target-hosts` as is and throw an unimplemented exception if multiple hosts are provided with the plan to mirror the above REST functionality in a follow up. Lmk what you think.

osbenchmark/worker_coordinator/proto_helpers/ProtoQueryHelper.py
finnegancarroll: > we plan to drop support for Python 3.8 and 3.9 shortly

The published python libraries for [opensearch-protobufs](https://pypi.org/project/opensearch-protobufs/) have >3.10 requirement. Does this PR depend on dropping python3.8/9?

finnegancarroll:
Hi @gkamat, it looks like the integration tests for OSB run against a [few select versions](https://github.com/opensearch-project/opensearch-benchmark/blob/main/it/__init__.py#L40) and source OS from the published release .tar. I expect the gRPC apis in this PR will largely be compatible with versions >3.3 of OpenSource.

I'm wondering if I need to wait for 3.3 to be published to introduce ITs for these changes?

osbenchmark/worker_coordinator/proto_helpers/ProtoBulkHelper.py
finnegancarroll: These are OSB provided bulk request bodies as would get passed to the REST client. In the case of the REST client they are passed through directly to the `opensearchpy` client so I think they should all conform to the REST JSON structure.

Good catch this is pretty messy. Simplifying to just grab every other line which we know to be a document.

osbenchmark/worker_coordinator/proto_helpers/ProtoBulkHelper.py
finnegancarroll: Woops

osbenchmark/worker_coordinator/proto_helpers/ProtoBulkHelper.py
finnegancarroll: In this version status for each response item mirrors http error codes:
https://github.com/opensearch-project/opensearch-protobufs/blob/b6f889416da83b7dc4a0408347965e7820bd61d0/protos/schemas/document.proto#L217-L219

Checking > 299 is then checking for 4xx 5xx errors in the response items.

osbenchmark/worker_coordinator/proto_helpers/ProtoQueryHelper.py
finnegancarroll: I think the key here is ignored. Big5 for example gives term queries of the form:
```
"term": {
    "log.file.path": {
        "value" : ["/var/log/messages/birdknight", "/var/log/messages/otherterm"]
    }
}
```
where "value" is completely ignored.

osbenchmark/client.py
finnegancarroll: The load balancing for multiple hosts seems to come from opensearchpy client:
https://github.com/opensearch-project/opensearch-benchmark/pull/938/files#r2408278949

At this stage i'm planning on throwing exceptions here and later adding multi-cluster/multi-host support.

osbenchmark/worker_coordinator/proto_helpers/ProtoQueryHelper.py
finnegancarroll: Swapped these if statements for match.

osbenchmark/worker_coordinator/proto_helpers/ProtoQueryHelper.py
finnegancarroll: For term query there will always be one key in the dictionary and the term will be a single item in a list. Lmk if you have a suggestion on a better way to write this.

osbenchmark/benchmark.py
gkamat: Sounds like a plan.

osbenchmark/client.py
```diff
@@ -278,3 +280,124 @@ async def create(params):
...
         else:
             raise ValueError(f"Unsupported ingestion source type: {producer_type}")
+
+
+class GrpcClientFactory:
+    """
+    Factory for creating gRPC client stubs.
+    Note gRPC channels must default `use_local_subchannel_pool` to true.
+    Sub channels manage the underlying connection with the server. When the global sub channel pool is used gRPC will
+    re-use sub channels and their underlying connections which does not appropriately reflect a multi client scenario.
```
gkamat, line 290: Probably could be clearer: using local subchannels permits additional connections to be created each with their own pools, which can improve performance.

osbenchmark/worker_coordinator/proto_helpers/ProtoQueryHelper.py
```diff
@@ -0,0 +1,205 @@
...
+
+def get_terms_dict(query):
+    terms = {}
+    for key, value in query.items():
+        terms[key] = []
+        if isinstance(value, list):
+            terms[key].extend(value)
+        elif isinstance(value, dict):
+            for _, term_value in value.items():
+                terms[key].append(term_value)
```
gkamat, line 32:
```
terms[key].extend(value.values())
```
might be simpler here.

osbenchmark/worker_coordinator/proto_helpers/ProtoQueryHelper.py
```diff
@@ -0,0 +1,205 @@
...
+    @staticmethod
+    def term_query_to_proto(query):
+        term = get_terms_dict(query)
+        if len(term.keys()) > 1:
+            raise Exception("Error parsing query - Term query contains multiple distinct fields: " + str(query))
+        if len(term.values()) > 1:
+            raise Exception("Error parsing query - Term query contains multiple terms: " + str(query))
+
+        # Term query body gives field/value as lists
+        term_field = next(iter(term.keys()))
```
gkamat, line 49: nit: `next(iter(term))` should suffice.

osbenchmark/worker_coordinator/proto_helpers/ProtoQueryHelper.py
```diff
@@ -0,0 +1,205 @@
...
+    # Build protobuf SearchRequest.
+    # Consumed from params dictionary:
+    # ``body``: query body as loaded from workload - Contains `_size` and `source`
+    # ``index``: index name
+    # ``request-timeout``: request timeout
+    # ``cache``: enabled request cache
+    @staticmethod
+    def build_proto_request(params):
+        body = params.get("body")
+        size = body.get("size") if "size" in body else None
```
gkamat, line 87:
```
size = body.get("size")
```
should suffice.

tests/worker_coordinator/proto_query_helper_test.py
```diff
@@ -0,0 +1,321 @@
...
+        }
+
+        result = ProtoQueryHelper.build_proto_request(params)
+
+        self.assertIsInstance(result, search_pb2.SearchRequest)
+        self.assertTrue(result.request_body.query.HasField("term"))
+        self.assertEqual(result.request_body.query.term.field, "log.file.path")
+        self.assertEqual(result.request_body.query.term.value.string, "/var/log/messages/birdknight")
+
+    def test_build_proto_request_term_query_multi_field_fails(self):
```
gkamat, line 78: Perhaps another test for the error case where there are multiple keys?

gkamat (approved):
None of the suggestions need to go in right away, perhaps in a future PR.

PR #964, opened 2025-09-24
Add 'fire and forget' mode to OSB

### LINKED ISSUE #958: [PROPOSAL] Add 'Unhinged mode' to OSB

Several users have expressed the want for a sustained throughput mode in OpenSearch Benchmark (OSB).

OSB's clients' currently run in a closed loop model, where individual clients send a request -> await a response -> record metrics -> send the next request during a load test

This closed loop model can cause throughput to suffer at higher loads due to response times' naturally taking longer and longer, which causes clients to fall behind schedule, which means OSB often has trouble sustaining high throughput numbers over a period of time.

OSB should have a second mode where clients' send requests in an open loop model, or 'fire and forget'. In this mode, OSB would continuously send the requested TPS without worrying about things like increasing latency or failed requests.

The drawback here is that because we are not waiting for responses, we'd be unable to record metrics. This mode would purely be for users who want to generate a sustained TPS against their cluster over some period of time. 

Async processes are able to fire requests and finish without needing to wait for the response:
```
    async def fire_request(self, request_id: int):
        """Fire a single request without waiting for response"""
        try:
            if self.session:
                # Fire and forget - don't await the response
                asyncio.create_task(self.session.get(self.target_url))
        except Exception as e:
            print(f"Error firing request {request_id}: {e}")
```

We could add a new class to [worker_coordinator.py](https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/worker_coordinator/worker_coordinator.py) which is meant for this second mode. 

This class would be similar to the AsyncExecutor but much more bare-bones, since it won't need any of the functionality to wait for responses or record metrics. All it would do is fire queries at the cluster.

We would also need to modify the [runner](https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/worker_coordinator/runner.py) class since this encapsulates the logic for running operations against clusters.

Users could activate the 'unhinged' mode with a new flag like `--fire-and-forget` or something similar, which OSB would take and then utilize this new client/runner combination to sustain whatever throughput the user defines for the duration of the test.

_No response_

### Additional context

_No response_

OVI3D0:
### Description
Adds new 'fire and forget' mode to OSB. 

Rather than OSB's traditional client model of 'fire request -> await request -> record metrics -> move on', this mode allows each client to fire requests without needing to worry about awaiting responses or recording metrics. This allows OSB to easily sustain high throughput values when load testing cluster, and is intended for those who don't necessarily want precise measurements on each request, but would rather test their clusters against very high sustained throughput levels.

The drawback here is of course that there is no information returned to the user on how their cluster is performing aside from outside forms of polling/measurements, like the performance charts than can be viewed in the AWS console for their cluster.

The PR introduces a new flag, `--fire-and-forget`, which tells OSB to choose the [DeterministicScheduler](https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/worker_coordinator/scheduler.py#L230-L246). Unlike the UnitAwareScheduler used for most benchmarks, this scheduler doesn't care about request metadata, it only calculates throughput for each client and tells them to send requests at a certain rate.

The flag also tells OSB to make use of a new request executor, called the UnhingedExecutor. This executor, unlike the AsyncExecutor, creates separate asynchronous tasks to send requests without awaiting them, ensuring requests are sent at the specified rate no matter the latency or failure rates. To sum it up, if each executor needs to send 2 RPS, this mode tells them to create 2 async tasks per second, each of which will send the request, rather than trying to send 1 request every 0.5 seconds on its own.

This mode can consume resources very quickly, and users should ensure their hardware is able to handle the thousands of async processes that are created when using this flag. It's also likely they will run into an OSerror `too many open files` with all of the connections being established.

The max number of network sockets can be checked with `ulimit -n` and changed with the same command, like: `ulimit -n 2048`

### Issues Resolved
#958 

### Testing

New unit tests + running tests in 'fire and forget' mode against OS cluster at 500 TPS:

Screenshot 2025-09-24 at 11 48 18 AM
Screenshot 2025-09-24 at 11 47 55 AM

Unit tests produce this warning since we don't await the async tasks:
```
sys:1: RuntimeWarning: coroutine 'UnhingedExecutor._fire_and_forget_request.<locals>.fire_and_forget_runner' was never awaited
Coroutine created at (most recent call last)
  File "/Users/mikeovi/.pyenv/versions/3.8.12/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/Users/mikeovi/.pyenv/versions/3.8.12/lib/python3.8/unittest/mock.py", line 1325, in patched
    return func(*newargs, **newkeywargs)
  File "/Users/mikeovi/workplace/opensearch-benchmark/tests/__init__.py", line 35, in async_wrapper
    asyncio.run(t(*args, **kwargs), debug=True)
  File "/Users/mikeovi/.pyenv/versions/3.8.12/lib/python3.8/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/Users/mikeovi/.pyenv/versions/3.8.12/lib/python3.8/asyncio/base_events.py", line 603, in run_until_complete
    self.run_forever()
  File "/Users/mikeovi/.pyenv/versions/3.8.12/lib/python3.8/asyncio/base_events.py", line 570, in run_forever
    self._run_once()
  File "/Users/mikeovi/.pyenv/versions/3.8.12/lib/python3.8/asyncio/base_events.py", line 1851, in _run_once
    handle._run()
  File "/Users/mikeovi/.pyenv/versions/3.8.12/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/Users/mikeovi/workplace/opensearch-benchmark/tests/worker_coordinator/worker_coordinator_test.py", line 2454, in test_fire_and_forget_request_no_throttling_needed
    await executor._fire_and_forget_request({}, 1.0, 0.0)  # expected_scheduled_time = 1.0
  File "/Users/mikeovi/workplace/opensearch-benchmark/osbenchmark/worker_coordinator/worker_coordinator.py", line 2603, in _fire_and_forget_request
    task = asyncio.create_task(fire_and_forget_runner())
```

---

### DIFF
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -792,6 +792,12 @@ def create_arg_parser():
     help="Path where the HTML visualization should be saved when --visualize is enabled. If not specified, it will be saved in the test run directory.",
     default=None
     )
+    test_run_parser.add_argument(
+        "--no-await",
+        help="Enable unhinged mode for maximum throughput without response handling or metrics collection (search queries only).",
+        action="store_true",
+        default=False
+    )
 
     ###############################################################################
     #
@@ -1141,6 +1147,7 @@ def configure_test(arg_parser, args, cfg):
     cfg.add(config.Scope.applicationOverride, "workload", "randomization.alpha", args.randomization_alpha)
     cfg.add(config.Scope.applicationOverride, "workload", "visualize", args.visualize)
     cfg.add(config.Scope.applicationOverride, "workload", "visualize.output.path", args.visualize_output_path)
+    cfg.add(config.Scope.applicationOverride, "worker_coordinator", "no_await", args.no_await)
     configure_workload_params(arg_parser, args, cfg)
     configure_connection_params(arg_parser, args, cfg)
     configure_telemetry_params(args, cfg)
--- a/osbenchmark/worker_coordinator/worker_coordinator.py
+++ b/osbenchmark/worker_coordinator/worker_coordinator.py
@@ -47,6 +47,7 @@ import thespian.actors
 
 from osbenchmark.utils import opts
 from osbenchmark import actor, config, exceptions, metrics, workload, client, paths, PROGRAM_NAME, telemetry
+from osbenchmark.context import RequestContextHolder
 from osbenchmark.worker_coordinator import runner, scheduler
 from osbenchmark.database.factory import DatabaseClientFactory
 from osbenchmark.database.registry import DatabaseType, get_client_factory
@@ -2244,7 +2245,8 @@ class AsyncIoAdapter:
                 client_id, task, schedule, opensearch, self.sampler, self.profile_sampler, self.cancel, self.complete,
                 task.error_behavior(self.abort_on_error), self.cfg, self.shared_states, self.feedback_actor, self.error_queue, self.queue_lock)
             final_executor = AsyncProfiler(async_executor) if self.profiling_enabled else async_executor
-            aws.append(final_executor())
+            executor_task = final_executor()
+            aws.append(executor_task)
         run_start = time.perf_counter()
         try:
             _ = await asyncio.gather(*aws)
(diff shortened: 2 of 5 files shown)


osbenchmark/benchmark.py
```diff
@@ -787,6 +787,12 @@ def add_workload_source(subparser):
     help="Path where the HTML visualization should be saved when --visualize is enabled. If not specified, it will be saved in the test run directory.",
     default=None
     )
+    test_run_parser.add_argument(
+        "--fire-and-forget",
```
rishabh6788: nit: can we come up with some other name, like `--no-await` or `--sustain` etc?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -2509,6 +2522,137 @@ async def __call__(self, *args, **kwargs):
                 self.complete.set()
             await self._cleanup()
 
+class UnhingedExecutor:
```
rishabh6788: Nit: Same here, can we change this to `AsyncNoAwaitExecutor` or something similar to maintain naming convention?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -2509,6 +2522,137 @@ async def __call__(self, *args, **kwargs):
...
+
+    async def _wait_for_rampup(self, rampup_wait_time: float) -> None:
+        """Wait for the ramp-up phase if needed."""
+        if rampup_wait_time:
+            self.logger.info("client id [%s] waiting [%.2f]s for ramp-up.", self.client_id, rampup_wait_time)
+            await asyncio.sleep(rampup_wait_time)
+            self.logger.info("Client id [%s] is running now.", self.client_id)
+
+
+    async def _fire_and_forget_request(self, params: dict, expected_scheduled_time: float, total_start: float) -> None:
```
rishabh6788: nit: same here for the naming.

rishabh6788:
LGTM apart from some naming conventions. Can you run test in timed mode, may be for 15-30 mins and share some more results, good idea to test it with ramp up mode as well.

OVI3D0:
> LGTM apart from some naming conventions. Can you run test in timed mode, may be for 15-30 mins and share some more results, good idea to test it with ramp up mode as well.

Here's a screenshot after running combined with the ramp-up test procedure property:
Screenshot 2025-10-15 at 3 25 00 PM

OVI3D0:
confirmed still working. OSB will not collect metrics in this mode but the requests are sent. In this case we sent 10,000 queries:
```
[ec2-user@ip-10-0-137-227 ~]$ curl -s 'http://localhost:9200/_nodes/stats/indices/search?human' | python3 -m json.tool | grep -E "query_total|query_time"
                    "query_total": 60006,
                    "query_time": "1.2m",
                    "query_time_in_millis": 76019,
                    "concurrent_query_total": 60002,
                    "concurrent_query_time": "1.2m",
                    "concurrent_query_time_in_millis": 75958,
                    "startree_query_total": 0,
                    "startree_query_time": "0s",
                    "startree_query_time_in_millis": 0,
[ec2-user@ip-10-0-137-227 ~]$ curl -s 'http://localhost:9200/_nodes/stats/indices/search?human' | python3 -m json.tool | grep -E "query_total|query_time"
                    "query_total": 60019,
                    "query_time": "1.2m",
                    "query_time_in_millis": 76038,
                    "concurrent_query_total": 60015,
                    "concurrent_query_time": "1.2m",
                    "concurrent_query_time_in_millis": 75977,
                    "startree_query_total": 0,
                    "startree_query_time": "0s",
                    "startree_query_time_in_millis": 0,
[ec2-user@ip-10-0-137-227 ~]$ curl -s 'http://localhost:9200/_nodes/stats/indices/search?human' | python3 -m json.tool | grep -E "query_total|query_time"
                    "query_total": 70006,
                    "query_time": "1.4m",
                    "query_time_in_millis": 88315,
                    "concurrent_query_total": 70002,
                    "concurrent_query_time": "1.4m",
                    "concurrent_query_time_in_millis": 88254,
                    "startree_query_total": 0,
                    "startree_query_time": "0s",
                    "startree_query_time_in_millis": 0,
[ec2-user@ip-10-0-137-227 ~]$ curl -s 'http://localhost:9200/_nodes/stats/indices/search?human' | python3 -m json.tool | grep -E "query_total|query_time"
                    "query_total": 70006,
                    "query_time": "1.4m",
                    "query_time_in_millis": 88315,
                    "concurrent_query_total": 70002,
                    "concurrent_query_time": "1.4m",
                    "concurrent_query_time_in_millis": 88254,
                    "startree_query_total": 0,
                    "startree_query_time": "0s",
                    "startree_query_time_in_millis": 0,
[ec2-user@ip-10-0-137-227 ~]$ curl -s 'http://localhost:9200/_nodes/stats/indices/search?human' | python3 -m json.tool | grep -E "query_total|query_time"
                    "query_total": 70006,
                    "query_time": "1.4m",
                    "query_time_in_millis": 88315,
                    "concurrent_query_total": 70002,
                    "concurrent_query_time": "1.4m",
                    "concurrent_query_time_in_millis": 88254,
                    "startree_query_total": 0,
                    "startree_query_time": "0s",
                    "startree_query_time_in_millis": 0,
[ec2-user@ip-10-0-137-227 ~]$
```

PR #968, opened 2025-10-16
Add looped mode support for bulk ingestion operation. 

rishabh6788:
### Description
This PR added `looped` mode parameter for bulk operations. Using this users will be able run ingestion even when the corpus is not big enough to sustain long term indexing at any specified rate. 

If not `time-period` is defined and looped mode is set to true then ingestion would run indefinitely. 

Example:
```
{
  "name": "bulk-index",
  "operation": {
    "operation-type": "bulk",
    "bulk-size": 5000,
    "looped": true
  },
  "time-period": 900,
  "clients": 10
}
```

### Issues Resolved

### Testing

---


gkamat (approved):
It might be useful for "looped" to be an integer instead, which would indicate the number of iterations to loop.  That would operate as in iteration-based mode and the user could specify "1000" for a thousand iterations, and "0" for infinite -- the current behaviour.

PR #978, opened 2025-11-04
Changes to aid integration between SDG and Streaming ingestion

gkamat:
### Description
Changes to aid integration between SDG and Streaming ingestion and some fixes.

### Testing
- [ ] Ran unit and integ tests.

---

### DIFF
--- a/osbenchmark/cloud_provider/vendors/s3_data_producer.py
+++ b/osbenchmark/cloud_provider/vendors/s3_data_producer.py
@@ -8,6 +8,7 @@
 
 import os
 import sys
+import time
 import logging
 
 from concurrent.futures import ThreadPoolExecutor, wait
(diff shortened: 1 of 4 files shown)


osbenchmark/cloud_provider/vendors/s3_data_producer.py
```diff
@@ -44,20 +45,33 @@ def __init__(self, bucket:str, keys, client_options: dict, data_dir=None) -> Non
...
+            while True:
+                response = self.s3_client.list_objects(Bucket=self.bucket, Prefix=self.keys[:-2])
+                for object in response['Contents']:
+                    key = object['Key']
+                    if key not in processed_keys:
+                        processed_keys.add(key)
+                        size = self.s3_client.head_object(Bucket=self.bucket, Key=key)['ContentLength']
+                        if size == 0:
+                            return
+                        yield key
```
IanHoang, line 60: should we add a comment here that the streaming terminates here when empty file is encountered?

osbenchmark/cloud_provider/vendors/s3_data_producer.py
```diff
@@ -44,20 +45,33 @@ def __init__(self, bucket:str, keys, client_options: dict, data_dir=None) -> Non
...
             print(f"Error: {e}")
 
-    def _get_keys(self):
-        rsl = list()
-        if self.keys[-1] == '*':
-            keys = self.s3_client.list_objects(Bucket=self.bucket, Prefix=self.keys[:-1])
-            for key in keys['Contents']:
-                rsl.append(key['Key'])
+    def _get_next_key(self):
+        if len(self.keys) > 2 and self.keys.endswith('**'):
```
IanHoang, line 49: Why does length of self.keys need to be at least 2?
```
len(self.keys) > 2
```

IanHoang: So that readers can better understand what's happening, we can abstract these clauses into methods that explain the action and return booleans
```
len(self.keys) > 2 and self.keys.endswith('**')
```
and 
```
len(self.keys) > 1 and self.keys[-1] == '*':
```

IanHoang (approved):
Left a few comments.

osbenchmark/cloud_provider/vendors/s3_data_producer.py
gkamat: Yes, that would be appropriate.  Right now, the user model might change based on feedback, but once confirmed, this will be refactored.

osbenchmark/cloud_provider/vendors/s3_data_producer.py
gkamat: Will add this in the next update.

PR #988, opened 2025-11-18
Decoupling Metric Store from Worker Coordinator

ajleong623:
### Description
There is a new actor called `SamplePostProcessorActor`. However the `SamplePostProcessorActor` acts as an actor for processing samples directly from the `Worker` actors as well as any task involving the metrics store. '

`SamplePostProcessorActor` initialization:

The new actor is initialized in the coordinator. This happens in the `prepare_benchmark` method. The `SamplePostProcessorActor` is initialized after receiving the `StartSamplePostProcessorActor` message which has the configurations for creating the metrics store, the metrics sample and profile sample post processor objects, and the telemetry collectors. `StartTelemetry` and `StopTelemetry` messages are now used to start and stop the telemetry collection threads through the coordinator

Which methods were changed:

- `send_samples`: In the worker class, instead of just sending samples to the coordinator, a `ProcessSamples` message is sent to the new `SamplePostProcessorActor` which is then sent to the `SamplePostprocessor` instance inside of the `SamplePostProcessorActor`. 
- `to_externalizable`: In the metric store, the `to_externalizable` method is used to send the results of running the workload to the coordinator. We will now need to call the method through the `SamplePostProcessorActor` and the `GetExternalizableMetricsStore` message.
- `close`: When the coordinator is closed, the message `CloseMetricsStore` will now be used to signal to the new actor to close the metrics store. 
- `reset_relative_time`: The `ResetRelativeTimeRequest` message will now be used instead to reset the relative time of the metric store in the new actor. 

One of the concerns if with synchronization. A lot of methods involving the metrics store that were synchronous are now handled through an asynchronous message to the `SamplePostProcessorActor` which holds the metric store. 

### Issues Resolved

### Testing

---


ajleong623:
I have finished adding some unit tests. Please let me know if anyone has any questions, comments, or suggestions.

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1845,6 +1941,7 @@ def send_samples(self):
             samples = self.sampler.samples
             if len(samples) > 0:
                 self.send(self.master, UpdateSamples(self.worker_id, samples, self.profile_sampler.samples))
+                self.send(self.sample_post_processor_actor, ProcessSamples(samples, self.profile_sampler.samples))
```
OVI3D0: how come we send samples to master and the new actor?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1196,6 +1212,10 @@ def start_benchmark(self):
 
         self.update_progress_message()
 
+    def index_name(self, test_run_timestamp):
+        ts = time.from_is8601(test_run_timestamp)
```
OVI3D0: LLM caught a typo here :) should be `from_iso8601()` no?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1347,20 +1362,99 @@ def update_progress_message(self, task_finished=False):
...
+                                                   test_procedure=msg.test_procedure.name,
+                                                   read_only=False)
+        self.sample_post_processor = DefaultSamplePostprocessor(self.metrics_store,
+                                                         msg.downsample_factor,
+                                                         msg.workload.meta_data,
+                                                         msg.test_procedure.meta_data)
+        self.profile_metrics_post_processor = ProfileMetricsSamplePostprocessor(self.metrics_store,
+                                                                            msg.workload.meta_data,
+                                                                            msg.test_procedure.meta_data)
+        os_clients = self.create_os_clients()
```
OVI3D0, line 1431: why do we create more OS clients here?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1347,20 +1362,99 @@ def update_progress_message(self, task_finished=False):
...
-        self.sample_post_processor(raw_samples)
-        profile_samples = self.raw_profile_samples
-        self.raw_profile_samples = []
-        if len(profile_samples) > 0:
-            if self.profile_metrics_post_processor is None:
-                self.profile_metrics_post_processor = ProfileMetricsSamplePostprocessor(self.metrics_store,
-                                                                                    self.workload.meta_data,
-                                                                                    self.test_procedure.meta_data)
-            self.profile_metrics_post_processor(profile_samples)
+class SamplePostProcessorActor(actor.BenchmarkActor):
```
OVI3D0, line 1403: let's add an ActorExitRequest method to this actor for a clean shutdown on benchmark completion

OVI3D0:
left a few comments/questions but this is shaping up pretty well! nice work

osbenchmark/worker_coordinator/worker_coordinator.py
ajleong623: Although the metrics collector is now in a separate actor, the progress message still needs the samples. https://github.com/opensearch-project/opensearch-benchmark/blob/9004daa9e3dfcec506e7ecd03071b4c3ee19a00f/osbenchmark/worker_coordinator/worker_coordinator.py#L1317. Should the progress bar also be handled in the new actor?

osbenchmark/worker_coordinator/worker_coordinator.py
ajleong623: It is also like that in the metrics file https://github.com/opensearch-project/opensearch-benchmark/blob/9004daa9e3dfcec506e7ecd03071b4c3ee19a00f/osbenchmark/time.py#L47. I think that LLM is correct, though, and both functions should be changed.

osbenchmark/worker_coordinator/worker_coordinator.py
ajleong623: For the new actor, telemetry has been moved over due to its dependence on the metric store. OS clients were a requirement for telemetry.

osbenchmark/worker_coordinator/worker_coordinator.py
ajleong623: Yes. Thank you for catching that. An edge case that has been bugging me is what happens when the coordinator shuts down before the `SamplePostProcessorActor` closes the metric store. I think that the clean up steps should be managed through looking out for that request. ~Additionally, maybe I should have named the new actor `MetricStoreActor` as well for readability and to tie it to the scope of its responsibilities.~ I will actually want to keep the current name.

ajleong623: Upon a second look, I noticed that whenever an exit request is received for the coordinator, the close method is called which signals to the new actor to close the metric store which I think handles a clean shutdown. I added the ActorExitRequest method either way for logging.

ajleong623:
Ok, I made another change where now the only information sent between the worker and coordinator is just a mapping between each client id to the latest progress that client has made, represented by a tuple. This hopefully reduces the amount of memory sent between the worker and coordinator since the coordinator did not need the full samples.

ajleong623:
A future idea could be also allowing the progress publisher to have its own actor, however, I that can be a separate PR.

ajleong623:
Here are the most recent results with the command:
```
opensearch-benchmark run   --pipeline=benchmark-only   --workload-path=~/opensearch-benchmark-workloads/big5/   --target-host=https://search-osb-metrics-testing-56ekd6uqkbq4yme7tbzuud7za4.aos.us-west-2.on.aws   --test-procedure=timed-mode-test-procedure   --workload-params='target_throughput:2000'   --redline-test   --client-options 'basic_auth_user:admin,basic_auth_password:Metrics0525!,verify_certs:false'   --kill-running-processes
```

Without any changes, I received

```
------------------------------------------------------
    _______             __   _____
   / ____(_)___  ____ _/ /  / ___/_________  ________
  / /_  / / __ \/ __ `/ /   \__ \/ ___/ __ \/ ___/ _ \
 / __/ / / / / / /_/ / /   ___/ / /__/ /_/ / /  /  __/
/_/   /_/_/ /_/\__,_/_/   /____/\___/\____/_/   \___/
------------------------------------------------------
            
|                                                         Metric |   Task |       Value |   Unit |
|---------------------------------------------------------------:|-------:|------------:|-------:|
|                     Cumulative indexing time of primary shards |        |     247.106 |    min |
|             Min cumulative indexing time across primary shards |        |     46.8489 |    min |
|          Median cumulative indexing time across primary shards |        |     48.7746 |    min |
|             Max cumulative indexing time across primary shards |        |     54.7089 |    min |
|            Cumulative indexing throttle time of primary shards |        |           0 |    min |
|    Min cumulative indexing throttle time across primary shards |        |           0 |    min |
| Median cumulative indexing throttle time across primary shards |        |           0 |    min |
|    Max cumulative indexing throttle time across primary shards |        |           0 |    min |
|                        Cumulative merge time of primary shards |        |     314.347 |    min |
|                       Cumulative merge count of primary shards |        |         202 |        |
|                Min cumulative merge time across primary shards |        |     59.7553 |    min |
|             Median cumulative merge time across primary shards |        |      63.448 |    min |
|                Max cumulative merge time across primary shards |        |     65.4447 |    min |
|               Cumulative merge throttle time of primary shards |        |     212.008 |    min |
|       Min cumulative merge throttle time across primary shards |        |     40.8501 |    min |
|    Median cumulative merge throttle time across primary shards |        |     42.0425 |    min |
|       Max cumulative merge throttle time across primary shards |        |     44.4238 |    min |
|                      Cumulative refresh time of primary shards |        |     34.1781 |    min |
|                     Cumulative refresh count of primary shards |        |         959 |        |
|              Min cumulative refresh time across primary shards |        |     6.44107 |    min |
|           Median cumulative refresh time across primary shards |        |     6.92882 |    min |
|              Max cumulative refresh time across primary shards |        |     7.06172 |    min |
|                        Cumulative flush time of primary shards |        |     8.52178 |    min |
|                       Cumulative flush count of primary shards |        |         249 |        |
|                Min cumulative flush time across primary shards |        |     1.62967 |    min |
|             Median cumulative flush time across primary shards |        |     1.67273 |    min |
|                Max cumulative flush time across primary shards |        |      1.8499 |    min |
|                                        Total Young Gen GC time |        |       8.152 |      s |
|                                       Total Young Gen GC count |        |         410 |        |
|                                          Total Old Gen GC time |        |           0 |      s |
|                                         Total Old Gen GC count |        |           0 |        |
|                                                      Doc count |        |    1.16e+08 |        |
|                                                     Store size |        |     176.041 |     GB |
|                                                  Translog size |        | 7.68341e-07 |     GB |
|                                         Heap used for segments |        |           0 |     MB |
|                                       Heap used for doc values |        |           0 |     MB |
|                                            Heap used for terms |        |           0 |     MB |
|                                            Heap used for norms |        |           0 |     MB |
|                                           Heap used for points |        |           0 |     MB |
|                                    Heap used for stored fields |        |           0 |     MB |
|                                                  Segment count |        |          93 |        |
|                                                 Min Throughput |  range |           0 |  ops/s |
|                                                Mean Throughput |  range |        9.35 |  ops/s |
|                                              Median Throughput |  range |       11.12 |  ops/s |
|                                                 Max Throughput |  range |       13.17 |  ops/s |
|                                        50th percentile latency |  range |     106.522 |     ms |
|                                        90th percentile latency |  range |     398.062 |     ms |
|                                        99th percentile latency |  range | 1.49195e+06 |     ms |
|                                      99.9th percentile latency |  range | 3.30313e+06 |     ms |
|                                     99.99th percentile latency |  range | 4.70798e+06 |     ms |
|                                       100th percentile latency |  range | 5.30169e+06 |     ms |
|                                   50th percentile service time |  range |     93.2246 |     ms |
|                                   90th percentile service time |  range |     332.353 |     ms |
|                                   99th percentile service time |  range | 1.49195e+06 |     ms |
|                                 99.9th percentile service time |  range | 3.30313e+06 |     ms |
|                                99.99th percentile service time |  range | 4.70798e+06 |     ms |
|                                  100th percentile service time |  range | 5.30169e+06 |     ms |
|                                                     error rate |  range |        3.32 |      % |

[WARNING] Error rate is 3.32 for operation 'range'. Please check the logs.
[INFO] Redline test finished. Maximum stable client number reached: 4793
```

Then with my changes, I received 

```
[Redline] Active clients: 4821
------------------------------------------------------
    _______             __   _____
   / ____(_)___  ____ _/ /  / ___/_________  ________
  / /_  / / __ \/ __ `/ /   \__ \/ ___/ __ \/ ___/ _ \
 / __/ / / / / / /_/ / /   ___/ / /__/ /_/ / /  /  __/
/_/   /_/_/ /_/\__,_/_/   /____/\___/\____/_/   \___/
------------------------------------------------------
            
|                                                         Metric |   Task |       Value |   Unit |
|---------------------------------------------------------------:|-------:|------------:|-------:|
|                     Cumulative indexing time of primary shards |        |     339.554 |    min |
|             Min cumulative indexing time across primary shards |        |           0 |    min |
|          Median cumulative indexing time across primary shards |        |     0.68265 |    min |
|             Max cumulative indexing time across primary shards |        |     54.7089 |    min |
|            Cumulative indexing throttle time of primary shards |        |           0 |    min |
|    Min cumulative indexing throttle time across primary shards |        |           0 |    min |
| Median cumulative indexing throttle time across primary shards |        |           0 |    min |
|    Max cumulative indexing throttle time across primary shards |        |           0 |    min |
|                        Cumulative merge time of primary shards |        |     335.244 |    min |
|                       Cumulative merge count of primary shards |        |        3253 |        |
|                Min cumulative merge time across primary shards |        |           0 |    min |
|             Median cumulative merge time across primary shards |        |   0.0269083 |    min |
|                Max cumulative merge time across primary shards |        |     65.4447 |    min |
|               Cumulative merge throttle time of primary shards |        |     223.251 |    min |
|       Min cumulative merge throttle time across primary shards |        |           0 |    min |
|    Median cumulative merge throttle time across primary shards |        |           0 |    min |
|       Max cumulative merge throttle time across primary shards |        |     44.4238 |    min |
|                      Cumulative refresh time of primary shards |        |     46.9055 |    min |
|                     Cumulative refresh count of primary shards |        |       40227 |        |
|              Min cumulative refresh time across primary shards |        |           0 |    min |
|           Median cumulative refresh time across primary shards |        |    0.114192 |    min |
|              Max cumulative refresh time across primary shards |        |     7.06172 |    min |
|                        Cumulative flush time of primary shards |        |      11.034 |    min |
|                       Cumulative flush count of primary shards |        |         487 |        |
|                Min cumulative flush time across primary shards |        |           0 |    min |
|             Median cumulative flush time across primary shards |        | 0.000133333 |    min |
|                Max cumulative flush time across primary shards |        |      1.8499 |    min |
|                                        Total Young Gen GC time |        |        8.35 |      s |
|                                       Total Young Gen GC count |        |         425 |        |
|                                          Total Old Gen GC time |        |           0 |      s |
|                                         Total Old Gen GC count |        |           0 |        |
|                                                      Doc count |        | 3.63249e+08 |        |
|                                                     Store size |        |     227.834 |     GB |
|                                                  Translog size |        | 1.44951e-05 |     GB |
|                                         Heap used for segments |        |           0 |     MB |
|                                       Heap used for doc values |        |           0 |     MB |
|                                            Heap used for terms |        |           0 |     MB |
|                                            Heap used for norms |        |           0 |     MB |
|                                           Heap used for points |        |           0 |     MB |
|                                    Heap used for stored fields |        |           0 |     MB |
|                                                  Segment count |        |         403 |        |
|                                                 Min Throughput |  range |           0 |  ops/s |
|                                                Mean Throughput |  range |       72.36 |  ops/s |
|                                              Median Throughput |  range |       26.76 |  ops/s |
|                                                 Max Throughput |  range |      164.95 |  ops/s |
|                                        50th percentile latency |  range |     95.8603 |     ms |
|                                        90th percentile latency |  range |      297.68 |     ms |
|                                        99th percentile latency |  range | 1.20073e+06 |     ms |
|                                      99.9th percentile latency |  range | 2.42907e+06 |     ms |
|                                     99.99th percentile latency |  range | 3.74694e+06 |     ms |
|                                       100th percentile latency |  range |  4.6101e+06 |     ms |
|                                   50th percentile service time |  range |     86.6219 |     ms |
|                                   90th percentile service time |  range |     266.132 |     ms |
|                                   99th percentile service time |  range | 1.20073e+06 |     ms |
|                                 99.9th percentile service time |  range | 2.42907e+06 |     ms |
|                                99.99th percentile service time |  range | 3.74694e+06 |     ms |
|                                  100th percentile service time |  range |  4.6101e+06 |     ms |
|                                                     error rate |  range |        3.09 |      % |

[WARNING] Error rate is 3.09 for operation 'range'. Please check the logs.
[INFO] Redline test finished. Maximum stable client number reached: 4821
```

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1348,20 +1360,104 @@ def update_progress_message(self, task_finished=False):
...
+        self.metrics_store.reset_relative_time()
+
+    def receiveMsg_ActorExitRequest(self, msg, sender):
+        self.logger.info("SamplePostProcessorActor has received ActorExitRequest and will close the metrics store.")
+        self.close()
+
+    def close(self):
+        if self.metrics_store and self.metrics_store.opened:
+            self.metrics_store.close()
+        self.logger.info("Metrics store close request has been processed.")
```
OVI3D0: I would be careful here that we're not dropping samples at the end

In the old code, the `post_process_samples()` method is called one last time before shutdown and is run synchronously, which guarantees that the last samples are sent out and processed before shutting down. Here we just shut down the metrics store so we might be silently dropping samples in the end

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1348,20 +1360,104 @@ def update_progress_message(self, task_finished=False):
...
+        if msg.samples:
+            self.sample_post_processor(msg.samples)
+        if msg.profile_samples:
+            self.profile_metrics_post_processor(msg.profile_samples)
+
+    def receiveMsg_StartTelemetry(self, msg, sender):
+        self.telemetry.on_benchmark_start()
+
+    def receiveMsg_StopTelemetry(self, msg, sender):
+        self.telemetry.on_benchmark_stop()
```
OVI3D0: Because OSB is adding support for other databases we'll need a check to ensure we're connected to an OpenSearch instance before turning telemetry on

there should be a check in the original code, something like `if database_type.lower() == "opensearch"`, we should add that here

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1348,20 +1360,104 @@ def update_progress_message(self, task_finished=False):
...
+
+        os_default = opensearch["default"]
+
+        if enable:
+            devices = [
+                telemetry.NodeStats(telemetry_params, opensearch, self.metrics_store),
+                telemetry.ExternalEnvironmentInfo(os_default, self.metrics_store),
+                telemetry.ClusterEnvironmentInfo(os_default, self.metrics_store),
+                telemetry.JvmStatsSummary(os_default, self.metrics_store),
+                telemetry.IndexStats(os_default, self.metrics_store),
```
OVI3D0: this had an index names arg in the original code let's put it back

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1217,23 +1235,19 @@ def joinpoint_reached(self, worker_id, worker_local_timestamp, task_allocations)
...
                 self.logger.info("All steps completed.")
                 # Some metrics store implementations return None because no external representation is required.
                 # pylint: disable=assignment-from-none
-                m = self.metrics_store.to_externalizable(clear=True)
+                self.target.send(self.target.sample_post_processor_actor, GetExternalizableMetricsStore(True, reason=ReasonForExternalizableRequest.BENCHMARK_COMPLETED))
                 self.logger.debug("Closing metrics store...")
-                self.metrics_store.close()
+                self.close_metric_store()
                 # immediately clear as we don't need it anymore and it can consume a significant amount of memory
                 self.metrics_store = None
```
OVI3D0: can we make sure we're not duplicating code here, because there is a method in the new post processor actor that also closes the metrics store right?

tests/worker_coordinator/worker_coordinator_test.py
```diff
@@ -251,7 +251,9 @@ def test_client_reaches_join_point_which_completes_parent(self):
 
         # this requires at least Python 3.6
         # target.on_task_finished.assert_called_once()
-        self.assertEqual(1, target.on_task_finished.call_count)
+        # on PR https://github.com/opensearch-project/opensearch-benchmark/pull/988,
+        # target.on_task_finished is now delegated to an extenal actor.
+        #self.assertEqual(1, target.on_task_finished.call_count)
```
OVI3D0: we can delete this if not necessary or just write a new test for it if needed

osbenchmark/worker_coordinator/worker_coordinator.py
ajleong623: That is true. In this code, `post_process_samples()` is no longer called. However, the samples are sent directly to the new actor. So, it looks like we should somehow program the messaging to make sure that all the workers have sent in their samples before shutting down or that all `ProcessSample` messages are received before the `CloseMetricsStore` or `ActorExitRequest` messages are processed? I will have to think about a solution for a little bit.

osbenchmark/worker_coordinator/worker_coordinator.py
ajleong623: Great catch. I noticed that this is being called in the [`prepare_telemetry` method](https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/worker_coordinator/worker_coordinator.py#L987-L1020) which has been moved to the new actor rather than the coordinator.

osbenchmark/worker_coordinator/worker_coordinator.py
ajleong623: Yes. This is in the worker coordinator end and should be signaling a message pass to the new actor to close the metric store. However, I did notice that a reference to the metric store in the coordinator is no longer needed. I also noticed that after closing, the metric store reference should be cleared.

ajleong623:
Ok, so to make sure the metric store collected was not closed before samples are processed, I made sure the join point logic is only submitted after the processing logic. The join point area is where the logic around finishing the task is handled.

ajleong623:
Summary of changes:

The workers will now send the samples they processed to a new actor called` SamplePostProcessorActor`. 

New messages:

- `ProcessSamples`: Information about samples that should be processed by `SamplePostProcessorActor` sent from workers.
- `UpdateProgressSamples`: The information sent from worker to worker coordinator for updating the progress publisher (The worker coordinator is still behind publishing progress messages)
- `CloseMetricsStore`: Passed from coordinator to new `SamplePostProcessorActor` about closing the metric store
- `GetExternalizableMetricsStore`: Sent from coordinator to `SamplePostProcessorActor` which tells the
- `SamplePostProcessorActor` to send a snapshot of the metric store to the coordinator. One thing to note is that with making sending snapshots asynchronous, the worker coordinator will have to receive the messages of `TaskFinished` and `BenchmarkComplete`.
- `ResetRelativeTimeRequest`: Since the metric store is now asynchronous, the method call of `reset_relative_time()` now needs to be sent through a message
- `StartSamplePostProcessorActor`: Tells the `SamplePostProcessorActor` to initialize and is sent during the prepare_benchmark method in the coordinator
- `StartTelemetry`/`StopTelemetry`: Is now sent from the coordinator to `SamplePostProcessorActor` for signaling the method calls of `self.telemetry.on_benchmark_start()` and `self.telemetry.on_benchmark_stop()`

Codeflow changes:

- Signalling the joinpoint message to the coordinator is no longer just handled by the workers. The message is forwarded from the `SamplePostProcessorActor` instead so that methods around joinpoint reached is handled after processing samples
- Workers still have to forward their progress to the coordinator for publishing progress, however, the samples themselves are still forwarded to the new actor.
- The new actor is initialized in the `prepare_benchmark` method in the coordinator.
- When the benchmark is cancelled, the `ActorExitRequest` message is now forwarded to the `SamplePostProcessorActor`

osbenchmark/worker_coordinator/worker_coordinator.py
ajleong623: In the new changes, I have synced the join point phase to happen after the samples are also processed. The join point phase is where the logic behind the coordinator actor closing the metric store happens without a cancellation.

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1378,20 +1388,115 @@ def update_progress_message(self, task_finished=False):
...
+        if enable:
+            # Only enable OpenSearch-specific telemetry for OpenSearch databases
+            if database_type.lower() == "opensearch":
+                devices = [
+                    telemetry.NodeStats(telemetry_params, opensearch, self.metrics_store),
+                    telemetry.ExternalEnvironmentInfo(os_default, self.metrics_store),
+                    telemetry.ClusterEnvironmentInfo(os_default, self.metrics_store),
+                    telemetry.JvmStatsSummary(os_default, self.metrics_store),
+                    telemetry.IndexStats(os_default, self.metrics_store,
+                                         index_names=[idx.name for idx in self.workload.indices] if self.workload.indices else None),
```
OVI3D0, line 1608: As it stands, this line is crashing any benchmark run with `AttributeError: 'SamplePostProcessorActor' object has no attribute 'workload'`

Basically what's happening is `self.workload` is never assigned under `SamplePostProcessorActor` so when you try to grab `self.workload.indices` it crashes any benchmark run. Try assigning `self.workload` under `receiveMsg_StartSamplePostProcessorActor` when you read it from the msg object

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1196,14 +1208,16 @@ def start_benchmark(self):
...
             elif cpu_max and "node-stats" not in self.config.opts("telemetry", "devices"):
                 raise exceptions.SystemSetupError("Node stats telemetry not enabled — this is required for CPU-based redline feedback.")
-            elif cpu_max and isinstance(self.metrics_store, metrics.OsMetricsStore):
+            elif cpu_max and metrics_store_type is metrics.OsMetricsStore:
                 # pass over the index and test run ID so the feedbackActor can query the datastore
-                metrics_index = self.metrics_store.index
-                test_run_id = self.metrics_store.test_run_id
+                test_run_timestamp = time.to_iso8601(self.config.opts("system", "time.start"))
+                metrics_index = self.index_name(test_run_timestamp)
+                test_run_id = self.config.opts("system", "test_run.id")
```
OVI3D0: can we double check that this will still function in the same manner? Redline testing needs the index name to query for CPU based feedback.

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -617,6 +666,7 @@ def receiveMsg_BenchmarkFailure(self, msg, sender):
     def receiveMsg_BenchmarkCancelled(self, msg, sender):
         self.logger.info("Main worker_coordinator received a notification that the benchmark has been cancelled.")
         self.coordinator.close()
+        self.send(self.sample_post_processor_actor, thespian.actors.ActorExitRequest())
```
OVI3D0: let's add a quick 'if' check here like below with the feedback actor, if for some reason this came in before `prepare_benchmark` ran then it could lead to issues

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -960,9 +1017,7 @@ def __init__(self, target, config, os_client_factory_class=client.OsClientFactor
         self.allocations = None
         self.raw_samples = []
         self.raw_profile_samples = []
```
OVI3D0: nit but looks like these aren't used anymore so we can clean these up

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1196,14 +1208,16 @@ def start_benchmark(self):
...
+            if cpu_max and metrics_store_type is metrics.InMemoryMetricsStore:
                 raise exceptions.SystemSetupError("CPU-based feedback requires a metrics store. You are using an in-memory metrics store")
             elif cpu_max and "node-stats" not in self.config.opts("telemetry", "devices"):
                 raise exceptions.SystemSetupError("Node stats telemetry not enabled — this is required for CPU-based redline feedback.")
-            elif cpu_max and isinstance(self.metrics_store, metrics.OsMetricsStore):
+            elif cpu_max and metrics_store_type is metrics.OsMetricsStore:
                 # pass over the index and test run ID so the feedbackActor can query the datastore
-                metrics_index = self.metrics_store.index
-                test_run_id = self.metrics_store.test_run_id
+                test_run_timestamp = time.to_iso8601(self.config.opts("system", "time.start"))
```
OVI3D0: Just wondering is `time.start` not already formatted as a `datetime` object?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1871,11 +1977,19 @@ def current_tasks_and_advance(self):
...
+            profile_samples = self.profile_sampler.samples
+            # Map client ids to their latest task progress, e.g. {0: (0.5, "%")}.
+            latest_progress_per_client = {}
+            for s in samples:
+                if s.task_progress is not None:
+                    latest_progress_per_client[s.client_id] = s.task_progress
+            if latest_progress_per_client:
+                self.send(self.master, UpdateProgressSamples(latest_progress_per_client))
+            if samples or profile_samples or joinpoint_reached:
+                self.send(self.sample_post_processor_actor, ProcessSamples(samples, profile_samples, joinpoint_reached=joinpoint_reached))
```
OVI3D0: So it looks like now we send 2 messages per wake up instead of 1, i just want to make sure that this will not flood the actor mailbox's and run us into the same issues. Might be worth measuring here if you'd like to give it a try?

OVI3D0:
Left a few comments but right now it looks like any benchmark run will crash with these changes, also it seems like your CI checks are failing

osbenchmark/worker_coordinator/worker_coordinator.py
ajleong623: Got it. At the time, the CI did not run all the workflows. I think my permissions reset after the CIs went down.

osbenchmark/worker_coordinator/worker_coordinator.py
ajleong623: Right. Are you referring to the UpdateProgressSamples message? This is only needed for the progress publishing logic. I think that the logic could also be moved to its new actor as a future improvement. That way the worker coordinator would not receive any direct messages from the workers.

osbenchmark/worker_coordinator/worker_coordinator.py
ajleong623: I logged it, and it is formatted as a date time object, however, the conversion will turn it into iso which removes the delimiters.

osbenchmark/worker_coordinator/worker_coordinator.py
ajleong623: https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/metrics.py#L852-L854

This is the logic I am copying from. Additionally, I logged the values just to make sure and here are the results.

Here is the original `index_name` and `test_run_id`:
Screenshot 2026-06-16 at 10 04 55 PM


Here are the new values as well as the timestamp formats:
Screenshot 2026-06-16 at 10 01 19 PM

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1247,23 +1271,17 @@ def joinpoint_reached(self, worker_id, worker_local_timestamp, task_allocations)
...
-            self.logger.debug("Postprocessing samples...")
-            self.post_process_samples()
             if self.finished():
-                self.telemetry.on_benchmark_stop()
+                self.target.send(self.target.sample_post_processor_actor, StopTelemetry())
                 self.logger.info("All steps completed.")
                 # Some metrics store implementations return None because no external representation is required.
                 # pylint: disable=assignment-from-none
-                m = self.metrics_store.to_externalizable(clear=True)
+                self.target.send(self.target.sample_post_processor_actor, GetExternalizableMetricsStore(True, reason=ReasonForExternalizableRequest.BENCHMARK_COMPLETED))
```
ajleong623: There is a potential race condition here. We are sending the `GetExternalizableMetricStore` which prepares the metric store entries to be send. We are also sending a message to close the metric store. What if the metric store close happens  while `metric_store.to_externalizable()` has not completed?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1378,20 +1393,121 @@ def update_progress_message(self, task_finished=False):
...
+
+    def receiveMsg_ResetRelativeTimeRequest(self, msg, sender):
+        self.metrics_store.reset_relative_time()
+
+    def receiveMsg_ActorExitRequest(self, msg, sender):
+        self.logger.info("SamplePostProcessorActor has received ActorExitRequest and will close the metrics store.")
+        self.close()
+
+    def close(self):
+        if self.metrics_store and self.metrics_store.opened:
```
ajleong623: Another potential issue is that what if this is called twice because we are receiving async messages. I think it will be better to make this idempotent. Also make sure telemetry is closed when this actor is closed.

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1378,20 +1393,121 @@ def update_progress_message(self, task_finished=False):
...
+                                                                            msg.test_procedure.meta_data)
+        os_clients = self.create_os_clients()
+        uses_static_responses = self.config.opts("client", "options").uses_static_responses
+
+        # Avoid issuing any requests to the target cluster when static responses are enabled. The results
+        # are not useful and attempts to connect to a non-existing cluster just lead to exception traces in logs.
+        self.prepare_telemetry(os_clients, enable=not uses_static_responses)
+        self.worker_coordinator_actor = sender
+
+    def receiveMsg_ProcessSamples(self, msg, sender):
```
ajleong623, line 1442: What happens if there are late samples during a premature closure?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -617,6 +666,9 @@ def receiveMsg_BenchmarkFailure(self, msg, sender):
     def receiveMsg_BenchmarkCancelled(self, msg, sender):
         self.logger.info("Main worker_coordinator received a notification that the benchmark has been cancelled.")
         self.coordinator.close()
```
ajleong623, line 688: We are closing the metric store twice here by the `coordinator.close()` and the `ActorExitRequest`. This will be okay as long as the closing logic is idempotent for the SamplePostProcessingActor.

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -675,30 +727,39 @@ def receiveMsg_WorkloadPrepared(self, msg, sender):
...
-        self.coordinator.update_profile_samples(msg.profile_samples)
+    def receiveMsg_UpdateProgressSamples(self, msg, sender):
+        self.coordinator.update_samples(msg.latest_progress_per_client)
 
     @actor.no_retry("worker_coordinator")  # pylint: disable=no-value-for-parameter
     def receiveMsg_WakeupMessage(self, msg, sender):
         if msg.payload == WorkerCoordinatorActor.RESET_RELATIVE_TIME_MARKER:
             self.coordinator.reset_relative_time()
         elif not self.coordinator.finished():
-            self.post_process_timer += WorkerCoordinatorActor.WAKEUP_INTERVAL_SECONDS
```
ajleong623, line 690: Before, there would be a timer and an interval for when post_processing happens, however, now it happens on every sample. This might save temporary memory, however, when there is a remote metric store, there will be more flushes since every sample will trigger a flush.  https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/worker_coordinator/worker_coordinator.py#L1541. This is problematic as there will be extra network latency. To be safe, let us have the new metric store keep a timer and accumulate like what the original coordinator code did. There was also a whole 30 seconds between intervals. This is legacy code, but there might be a reason why the waits were that long.

tests/worker_coordinator/worker_coordinator_test.py
```diff
@@ -2476,3 +2507,464 @@ def test_task_hash_includes_ramp_down(self):
...
+        latest_progress_per_client = {
+            0: (0.5, "%")
+        }
+
+        actor.receiveMsg_UpdateProgressSamples(worker_coordinator.UpdateProgressSamples(latest_progress_per_client), sender=None)
+
+        actor.coordinator.update_samples.assert_called_once_with(latest_progress_per_client)
+
+    def test_worker_coordinator_updates_latest_progress_per_client(self):
+        coordinator = worker_coordinator.WorkerCoordinator.__new__(worker_coordinator.WorkerCoordinator)
```
ajleong623: I would prefer if the tests followed the existing pattern for how coordinators are initialized.

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1378,20 +1391,186 @@ def update_progress_message(self, task_finished=False):
...
+            self.logger.debug("Ignoring samples received after SamplePostProcessorActor has closed.")
+            return
+        if msg.samples:
+            self.raw_samples.extend(msg.samples)
+        if msg.profile_samples:
+            self.raw_profile_samples.extend(msg.profile_samples)
+        if msg.joinpoint_reached:
+            self.post_process_samples()
+            self.logger.debug("Join point reached message received in SamplePostProcessorActor. Notifying WorkerCoordinatorActor...")
+            self.send(self.worker_coordinator_actor, msg.joinpoint_reached)
```
OVI3D0: just to be safe can we wrap this in a try/finally just to make sure the benchmark never hangs and this actor always sends the joinpoint reached message?

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1284,8 +1301,10 @@ def move_to_next_task(self, workers_curr_step):
             waiting_period = 1.0
         # Some metrics store implementations return None because no external representation is required.
         # pylint: disable=assignment-from-none
-        m = self.metrics_store.to_externalizable(clear=True)
-        self.target.on_task_finished(m, waiting_period)
+        self.target.send(self.target.sample_post_processor_actor,
+            GetExternalizableMetricsStore(True,
+            reason=ReasonForExternalizableRequest.TASK_FINISHED, waiting_period=waiting_period)
+        )
```
ajleong623: There is a race condition here because the workers are being reactivated while the metric store results are in the process of being sent to the parent actor of `worker_coordinator`. This would mean that `ProcessSample` messages could be sent to the SPP actor while thematic store results are still being externalized.

ajleong623:
`TaskBoundaryFlushed` is sent from SPP actor to worker coordinator when the externalized metric store is ready to be submitted to the parent of the worker coordinator actor during the completion of a task.
`FlushAndForwardJoinPoint` sent from worker to SPP to post process the samples and also send the join point.
`FlushForTaskBoundary` sent from worker coordinator to SPP to tell SPP to post process samples and send an externalized metric store.
`FlushAndClose` sent from worker coordinator when the task is finished and a BenchmarkComplete needs to be sent with the metric store externalized. 

The reason for these design choices are to further separate the different reasons for sending an externalized metric store and separating the join point forwarding logic instead of having it embedded within the ProcessSamples logic.

The new method `drive_workers_for_next_task` contains the separated logic from the original `move_to_next_task` method. This makes sure that the SPP actor has completed its post_processing and externalizing the metric store before the new workers are set when a join point is reached.

Additionally, try/catch blocks around `post_process_samples` method calls are added to make sure failures there do not stall the rest of the commands. 

Let me know if these changes are sufficient. I still have to clean up the logging statements.

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -617,6 +684,9 @@ def receiveMsg_BenchmarkFailure(self, msg, sender):
     def receiveMsg_BenchmarkCancelled(self, msg, sender):
         self.logger.info("Main worker_coordinator received a notification that the benchmark has been cancelled.")
         self.coordinator.close()
+        if hasattr(self, "sample_post_processor_actor"):
+            self.logger.info("Shutting down SamplePostProcessorActor due to benchmark cancellation.")
+            self.send(self.sample_post_processor_actor, thespian.actors.ActorExitRequest())
         # shut down FeedbackActor if it's active
         # we do this manually in the workercoordinator since it's fully responsible for the feedback actor
         if hasattr(self, "feedback_actor"):
```
OVI3D0: the two `hasattr` checks here are essentially no-ops since `sample_post_processor_actor` is set to `None` above. Easiest fix would be to say: 
```python
if hasattr(self, "sample_post_processor_actor") is not None:
```

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -675,30 +745,44 @@ def receiveMsg_WorkloadPrepared(self, msg, sender):
...
             self.coordinator.update_progress_message()
             self.wakeupAfter(datetime.timedelta(seconds=WorkerCoordinatorActor.WAKEUP_INTERVAL_SECONDS))
 
+    @actor.no_retry("worker_coordinator")  # pylint: disable=no-value-for-parameter
+    def receiveMsg_BenchmarkComplete(self, msg, sender):
+        self.on_benchmark_complete(msg.metrics)
+
+    @actor.no_retry("worker_coordinator")  # pylint: disable=no-value-for-parameter
+    def receiveMsg_TaskFinished(self, msg, sender):
+        self.on_task_finished(msg.metrics, msg.next_task_scheduled_in)
```
OVI3D0: seems like this is now dead code and can be removed

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -675,30 +745,44 @@ def receiveMsg_WorkloadPrepared(self, msg, sender):
...
+
+    @actor.no_retry("worker_coordinator")  # pylint: disable=no-value-for-parameter
+    def receiveMsg_TaskBoundaryFlushed(self, msg, sender):
+        self.on_task_finished(msg.metrics, msg.next_task_scheduled_in)
+        self.coordinator.drive_workers_for_next_task(msg.workers_curr_step, msg.next_task_scheduled_in)
```
OVI3D0, line 774: I think it's a good idea here to add a guard incase the benchmark is aborted

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1378,20 +1405,233 @@ def update_progress_message(self, task_finished=False):
...
+    @actor.no_retry("sample post processor")  # pylint: disable=no-value-for-parameter
+    def receiveMsg_CloseMetricsStore(self, msg, sender):
+        self.close()
+
+    @actor.no_retry("sample post processor")  # pylint: disable=no-value-for-parameter
+    def receiveMsg_FlushForTaskBoundary(self, msg, sender):
+        metric_results = None
+        if self.closed:
+            self.logger.debug("Ignoring task-boundary flush after SamplePostProcessorActor has closed.")
+            self.send(self.worker_coordinator_actor, TaskBoundaryFlushed(metric_results, msg.waiting_period, msg.workers_curr_step))
```
OVI3D0: if `self.closed` here let's just emit the debug log and not send anything

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1378,20 +1405,233 @@ def update_progress_message(self, task_finished=False):
...
+        try:
+            self.post_process_samples()
+            # Some metrics store implementations return None because no external representation is required.
+            # pylint: disable=assignment-from-none
+            metric_results = self.metrics_store.to_externalizable(clear=msg.clear)
+        except BaseException as e:
+            self.logger.exception("Could not flush samples at task boundary.")
+            self.send(self.worker_coordinator_actor, actor.BenchmarkFailure("Error in sample post processor ({})".format(str(e))))
+        finally:
+            self.send(self.worker_coordinator_actor, TaskBoundaryFlushed(metric_results, msg.waiting_period, msg.workers_curr_step))
```
OVI3D0, line 1500: I think here in the event of an Exception we would send a failure but then proceed to send TaskBoundaryFlushed. This could lead to the workers being re-driven after declaring a failure

osbenchmark/worker_coordinator/worker_coordinator.py
```diff
@@ -1378,20 +1405,233 @@ def update_progress_message(self, task_finished=False):
...
+            # Some metrics store implementations return None because no external representation is required.
+            # pylint: disable=assignment-from-none
+            metric_results = self.metrics_store.to_externalizable(clear=msg.clear)
+            self.close(process_samples=False, stop_telemetry=False)
+        except BaseException as e:
+            self.logger.exception("Could not flush and close samples at benchmark completion.")
+            self.send(self.worker_coordinator_actor, actor.BenchmarkFailure("Error in sample post processor ({})".format(str(e))))
+            self.close(process_samples=False, stop_telemetry=False, suppress_errors=True)
+        finally:
+            self.send(self.worker_coordinator_actor, BenchmarkComplete(metric_results))
```
OVI3D0, line 1521: Same as my previous comment about double sending, here the orchestrator would see BenchmarkFailure but then BenchmarkComplete which I think could lead to confusion.

Let's only send `TaskBoundaryFlushed` or `BenchmarkComplete` messages on the success path

OVI3D0:
once you rebase the unit tests failures about `stream_writer` should stop, but looks like there is one `too many public methods` error also

osbenchmark/worker_coordinator/worker_coordinator.py
OVI3D0: Edit: my suggested fix would actually still be a no-op since `hasattr(self, "sample_post_processor_actor")` would return `True` and `True is not None` would be the same thing. Try this instead:
```python
if self.sample_post_processor_actor is not None:
```

osbenchmark/worker_coordinator/worker_coordinator.py
ajleong623: For this, I added a status variable to denote when the worker coordinator is exiting.

OVI3D0 (approved):
This LGTM and will be included with OSB's 3.0 release.

ajleong623:
I think the last merge conflict came from registering the new telemetry device in https://github.com/opensearch-project/opensearch-benchmark/pull/1102. I resolved it by making sure that the new device is registered in the updated prepare_telemetry location (That method was moved by this change).

PR #1007, opened 2026-01-22
Add database abstraction layer for multi-database support

### LINKED ISSUE #1006: [MULTI-DB] Create database abstraction layer
We should create an abstraction layer in OSB. Rather than just having separate files for each DB, OSB should be able to support multiple DB's through a unified interface.

This will enable OSB to benchmark any DB that supports similar operations (bulk indexing, search, etc) while reusing existing workloads if possible, maintaining the same user experience and metrics collection, and keeping the codebase maintainable.

There should be abstract base classes (contracts) that each DB should follow, with clients, a set of operations/API calls, and runners. 

OpenSearch organizes operations into namespaces (e.g., `client.indices.create()`) so we can preserve that same pattern:
```
class IndicesNamespace(ABC):
    """Operations on indices/collections."""

    @abstractmethod
    async def create(self, index, body=None, params=None):
        pass

    @abstractmethod
    async def delete(self, index, params=None):
        pass

    @abstractmethod
    async def exists(self, index, params=None):
        pass


class ClusterNamespace(ABC):
    """Cluster-level operations."""

    @abstractmethod
    async def health(self, index=None, params=None):
        pass
```

and provide factories for creating DB clients without knowing which implementation to use.

OVI3D0:
### Description
Sets up abstraction layer in OSB and moves OS client/runner logic into appropriate folders

The goal is to set up OSB that can allow it to support any DB that supports similar operations. This PR creates the package structure, defines interfaces and contracts, refactors the existing OS code into the new structure, and updates imports. There are no functional changes to OSB here, just a reorganization of the existing code. 

Here we introduce:
`registry.py`, which defines the pattern for managing different DB's and their factories
`factory.py`, which is the interface for creating DB clients without needing to know which implementation to use
A couple classes in clients/opensearch/opensearch.py, which just wrap the current opensearchpy client so it conforms to the new DatabaseClient interface as well as provide an `OpenSearchclientFactory` class to create the OS clients.

Switches around a few things like checking for DB when loading config, client factory selection with a default of OS client, removes hardcoding for using OS client, skips telemetry for non OS clients. 

### Issues Resolved
#1006 

### Testing

Tested with big5 workload to ensure no breaking changes

---

### DIFF
--- a/it/__init__.py
+++ b/it/__init__.py
@@ -33,7 +33,8 @@ import datetime
 
 import pytest
 
-from osbenchmark import client, config, version, paths
+from osbenchmark.database.clients.opensearch import opensearch as client
+from osbenchmark import config, version, paths
 from osbenchmark.utils import process
 
 CONFIG_NAMES = ["in-memory-it", "os-it"]
--- a/osbenchmark/builder/builder.py
+++ b/osbenchmark/builder/builder.py
@@ -34,7 +34,8 @@ from collections import defaultdict
 import thespian.actors
 from opensearchpy.exceptions import NotFoundError
 
-from osbenchmark import (PROGRAM_NAME, actor, client, config, exceptions,
+from osbenchmark.database.clients.opensearch import opensearch as client
+from osbenchmark import (PROGRAM_NAME, actor, config, exceptions,
                          metrics, paths)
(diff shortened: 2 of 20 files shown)


OVI3D0:
@coderabbitai review

PR #1021, opened 2026-03-02
Update OSB with 3.5 gRPC schema changes

finnegancarroll:
### Description
Update OSB to the newest protobuf schema - 1.2.0.
Add compatibility matrix.

2.0.0 -> OpenSearch 3.3, 3.4
2.1.0 -> OpenSearch 3.5

### Issues Resolved
N/A

### Testing

Updated unit tests.

Manual testing with httplogs, big5, vectorsearch gRPC test procedures.
```
opensearch-benchmark run \
    --pipeline=benchmark-only \
    --target-hosts=localhost:9200 \
    --grpc-target-hosts=localhost:9400 \
    --workload="http_logs" \
    --test-procedure="grpc-append-no-conflicts-index-only" \
    --workload-params '{"number_of_shards":"1","number_of_replicas":"0", "ingest_percentage":"1.0", "bulk_size":"500"}' \
    --kill-running-processes
```

```
opensearch-benchmark run \
    --pipeline=benchmark-only \
    --target-hosts=${HOST}:9200 \
    --grpc-target-hosts=${HOST}:9400 \
    --workload="big5" \
    --test-procedure="grpc-big5" \
    --workload-params '{"number_of_shards":"1","number_of_replicas":"0", "ingest_percentage":"1.0", "bulk_size":"500"}' \
    --kill-running-processes
```

```
PARAMS_FILE=~/projs/test-params-cohere-1k.json
opensearch-benchmark run \
    --pipeline=benchmark-only \
    --target-hosts=localhost:9200 \
    --grpc-target-hosts=localhost:9400 \
    --test-procedure="grpc-no-train-test" \
    --workload-params="${PARAMS_FILE}" \
    --workload=vectorsearch \
    --kill-running-processes
```

---

### DIFF
--- a/osbenchmark/context.py
+++ b/osbenchmark/context.py
@@ -43,19 +43,22 @@ class RequestContextManager:
 
     @property
     def request_start(self):
-        return self.ctx["request_start"]
+        return self.ctx.get("request_start")
 
     @property
     def request_end(self):
-        return max((value for value in self.ctx["request_end_list"] if value < self.client_request_end))
+        client_end = self.client_request_end
+        if client_end is None or "request_end_list" not in self.ctx:
+            return None
+        return max((value for value in self.ctx["request_end_list"] if value < client_end))
 
     @property
     def client_request_start(self):
-        return self.ctx["client_request_start"]
+        return self.ctx.get("client_request_start")
 
     @property
     def client_request_end(self):
-        return self.ctx["client_request_end"]
+        return self.ctx.get("client_request_end")
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         # propagate earliest request start and most recent request end to parent
--- a/docs/grpc-support.md
+++ b/docs/grpc-support.md
@@ -1,0 +1,42 @@
+---
+layout: default
+title: gRPC Support
+parent: OSB
+nav_order: 20
+---
+
+# gRPC Support
+
+OpenSearch Benchmark supports benchmarking OpenSearch clusters via gRPC transport using the `--grpc-target-hosts` option. gRPC support requires the [transport-grpc](https://github.com/opensearch-project/OpenSearch/tree/main/modules/transport-grpc) module to be enabled on the OpenSearch cluster.
+
+## Compatibility Matrix
+
+| OpenSearch Benchmark Version | OpenSearch Version | opensearch-protobufs Version | Notes |
+|------------------------------|-------------------|------------------------------|-------|
+| 2.0.0                        | 3.3, 3.4          | 0.19.0                       | Initial gRPC support |
+| 2.1.0                        | 3.5+              | 1.2.0                        | Updated for OpenSearch 3.5 protobuf changes |
+
+## Supported Operations
+
+- Bulk ingestion
+- Match queries
+- Term queries
+- KNN queries
+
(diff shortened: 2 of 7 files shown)


finnegancarroll:
Hi @OVI3D0, I have a gRPC schema update for 3.5. Adding a compatibility matrix here as well to document these breaking changes going forward. Can you take a look when you have a chance? Thanks!

osbenchmark/context.py
```diff
@@ -43,19 +43,22 @@ async def __aenter__(self):
 
     @property
     def request_start(self):
-        return self.ctx["request_start"]
+        return self.ctx.get("request_start")
```
OVI3D0, line 46: just curious why change these to `.get`?

osbenchmark/context.py
```diff
@@ -43,19 +43,22 @@ async def __aenter__(self):
...
 
     @property
     def request_end(self):
-        return max((value for value in self.ctx["request_end_list"] if value < self.client_request_end))
+        client_end = self.client_request_end
+        if client_end is None or "request_end_list" not in self.ctx:
```
OVI3D0, line 51: curious, so it's the case that sometimes GRPC won't populate `request_end` ?

OVI3D0:
left a couple questions

osbenchmark/context.py
finnegancarroll: `get()` returns `None` if the key doesn't exist, where `ctx["request_start"]` throws a key error. I think this is slightly safer and I see that, at least for gRPC, this hides the true error as OSB eats the original exception and then reports the missing key  (`'client_request_end'`) as the root cause. This might only be the case for gRPC where we manually measure timings with the `RequestContextHolder`.

If I inject some test exception in a gRPC runner I see the following error message behavior.

Main:
```
[ERROR] ❌ Cannot run. Benchmark operation failed:
Worker ID: 0
Task: grpc-index-append
Workload: http_logs
Test Procedure: grpc-append-no-conflicts-index-only
Cause: Unknown
Error: Cannot run task [grpc-index-append]: 'client_request_end'
        Cannot run task [grpc-index-append]: 'client_request_end'
```

Feature branch:
```
[ERROR] ❌ Cannot run. Benchmark operation failed:
Worker ID: 5
Task: grpc-index-append
Workload: http_logs
Test Procedure: grpc-append-no-conflicts-index-only
Cause: Unknown
Error: Cannot run task [grpc-index-append]: An error occurred because of some_condition.
        Cannot run task [grpc-index-append]: An error occurred because of some_condition.

```

osbenchmark/context.py
finnegancarroll: Yea this is related to the above comment where we can fail to set `request_end` if an exception is thrown within the gRPC runner. These changes avoid throwing an exception within context.py so we can surface the true error.

rishabh6788:
LGTM, will let @OVI3D0 do the final check and merge.

rishabh6788 (approved):

osbenchmark/context.py
OVI3D0: Makes sense! I see this issue occasionally as well. May have to add a similar fix or see if there is somewhere upstream we can fix this.

PR #1030, opened 2026-03-25
fix: prevent __aexit__ from masking original runner exceptions

### LINKED ISSUE #1028: RequestContextManager.__aexit__ masks original runner exceptions with KeyError
## Bug Description

When a runner throws an exception before completing its request (e.g., HTTP 404, connection error, gRPC failure), the original error message is replaced by a misleading `KeyError: 'client_request_end'`.

### Why this happens

`RequestContextManager` wraps each runner execution in an `async with` block. The runner populates timing keys in `self.ctx` as it progresses:

1. `on_client_request_start()` → sets `ctx["client_request_start"]`
2. `on_request_start()` → sets `ctx["request_start"]`
3. **Runner does work (HTTP call, etc.)**
4. `on_request_end()` → sets `ctx["request_end_list"]`
5. `on_client_request_end()` → sets `ctx["client_request_end"]`

If the runner throws at step 3, steps 4-5 never run — `ctx` is missing those keys. Python guarantees `__aexit__` still fires, and it tries to read all 4 values to propagate timing to the parent context:

```python
async def __aexit__(self, exc_type, exc_val, exc_tb):
    client_request_start = self.client_request_start
    client_request_end = self.client_request_end  # KeyError here
    request_start = self.request_start
    request_end = self.request_end
    ...
```

The `KeyError` from `__aexit__` **replaces** the original exception, so the user sees `'client_request_end'` instead of the actual error.

### How to reproduce

Any runner that fails mid-request will trigger this. For example:

1. Create a custom runner that throws an exception after `on_client_request_start()` but before `on_client_request_end()`
2. Register it and run a workload that uses it

```python
class FailingRunner(Runner):
    async def __call__(self, client, params):
        request_context_holder.on_client_request_start()
        request_context_holder.on_request_start()
        raise Exception("actual error message")
        # on_request_end() and on_client_request_end() never called
```

**Actual output:**
```
Error: Cannot run task [my-task]: 'client_request_end'
```

**Expected output:**
```
Error: Cannot run task [my-task]: actual error message
```

This also occurs naturally when a `transport.perform_request()` call fails (e.g., connection timeout, HTTP error) before the timing context is fully populated.

### Suggested fix

Guard `__aexit__` to skip timing propagation when the context is incomplete, rather than making all properties lenient:

```python
async def __aexit__(self, exc_type, exc_val, exc_tb):
    if "client_request_end" not in self.ctx or "request_start" not in self.ctx:
        self.ctx_holder.restore_context(self.token)
        self.token = None
        return False
    # ... existing propagation logic
```

This preserves `KeyError` behavior on the properties (catching real bugs in normal flow) while allowing original exceptions to propagate correctly on the error path.

### Affected file

- `osbenchmark/context.py` — `RequestContextManager.__aexit__` (line 63)

thakoreh:
Fixes #1028

When a runner throws before completing its request (HTTP 404,
connection error, etc.), timing keys like client_request_start and
client_request_end may not be set in the context. The __aexit__
method attempts to propagate these timing values, which can raise
a KeyError that replaces the original exception from the runner.

This wraps the timing propagation in a try/except so context cleanup
failures never mask the real error. The context token is always
restored in a finally block to prevent context leaks.

### DIFF
--- a/osbenchmark/context.py
+++ b/osbenchmark/context.py
@@ -62,18 +62,30 @@ class RequestContextManager:
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         # propagate earliest request start and most recent request end to parent
-        client_request_start = self.client_request_start
-        client_request_end = self.client_request_end
-        request_start = self.request_start
-        request_end = self.request_end
-        self.ctx_holder.restore_context(self.token)
-        # don't attempt to restore these values on the top-level context as they don't exist
-        if self.token.old_value != contextvars.Token.MISSING:
-            self.ctx_holder.update_request_start(request_start)
-            self.ctx_holder.update_request_end(request_end)
-            self.ctx_holder.update_client_request_start(client_request_start)
-            self.ctx_holder.update_client_request_end(client_request_end)
-        self.token = None
+        try:
+            client_request_start = self.client_request_start
+            client_request_end = self.client_request_end
+            request_start = self.request_start
+            request_end = self.request_end
+
+            self.ctx_holder.restore_context(self.token)
+            # don't attempt to restore these values on the top-level context as they don't exist
+            if self.token.old_value != contextvars.Token.MISSING:
+                self.ctx_holder.update_request_start(request_start)
+                self.ctx_holder.update_request_end(request_end)
+                self.ctx_holder.update_client_request_start(client_request_start)
+                self.ctx_holder.update_client_request_end(client_request_end)
+        except Exception:
+            # Never mask the original exception from the runner.
+            # If context propagation fails, ensure the context is still restored.
+            pass
+        finally:
+            if self.token is not None:
+                try:
+                    self.ctx_holder.restore_context(self.token)
+                except Exception:
+                    pass
+                self.token = None
         return False
 
 
--- a/osbenchmark/context.py
+++ b/osbenchmark/context.py
@@ -63,13 +63,17 @@ class RequestContextManager:
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         # propagate earliest request start and most recent request end to parent
+        client_request_start = self.client_request_start
+        client_request_end = self.client_request_end
+        request_start = self.request_start
+        request_end = self.request_end
+        self.ctx_holder.restore_context(self.token)
         # don't attempt to restore these values on the top-level context as they don't exist
         if self.token.old_value != contextvars.Token.MISSING:
-            self.ctx_holder.update_request_start(self.request_start)
-            self.ctx_holder.update_request_end(self.request_end)
-            self.ctx_holder.update_client_request_start(self.client_request_start)
-            self.ctx_holder.update_client_request_end(self.client_request_end)
-        self.ctx_holder.restore_context(self.token)
+            self.ctx_holder.update_client_request_start(client_request_start)
+            self.ctx_holder.update_request_start(request_start)
+            self.ctx_holder.update_request_end(request_end)
+            self.ctx_holder.update_client_request_end(client_request_end)
         self.token = None
         return False
 
--- a/osbenchmark/context.py
+++ b/osbenchmark/context.py
@@ -50,7 +50,8 @@ class RequestContextManager:
         client_end = self.client_request_end
         if client_end is None or "request_end_list" not in self.ctx:
             return None
-        return max((value for value in self.ctx["request_end_list"] if value < client_end))
+        request_ends = [value for value in self.ctx["request_end_list"] if value is not None and value < client_end]
+        return max(request_ends) if request_ends else None
 
     @property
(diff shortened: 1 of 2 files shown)


osbenchmark/context.py
```diff
@@ -62,18 +62,30 @@ def client_request_end(self):
...
-            self.ctx_holder.update_client_request_start(client_request_start)
-            self.ctx_holder.update_client_request_end(client_request_end)
-        self.token = None
+        try:
+            client_request_start = self.client_request_start
+            client_request_end = self.client_request_end
+            request_start = self.request_start
+            request_end = self.request_end
+
+            self.ctx_holder.restore_context(self.token)
```
OVI3D0: looks like we're calling restore twice, once here, and again in the 'finally' block. We should probably remove it here and keep it in the finally block to avoid any potential issues.

osbenchmark/context.py
```diff
@@ -62,18 +62,28 @@ def client_request_end(self):
...
+                self.ctx_holder.update_client_request_start(client_request_start)
+                self.ctx_holder.update_client_request_end(client_request_end)
+        except Exception:
+            # Never mask the original exception from the runner.
+            # If context propagation fails, ensure the context is still restored.
+            pass
+        finally:
+            if self.token is not None:
+                try:
+                    self.ctx_holder.restore_context(self.token)
```
OVI3D0: i think burying this in the exception block could lead to issues later on

but looking back it might be even better to fix update_request_end / update_client_request_end / request_end to handle None gracefully

this way there wouldn't even need to be a try/except block here and we can also add a unit test for the failing scenario

thakoreh:
Updated in 2ca84572480c800b4c3483899e88d181b637c5ab: addressed the review by moving the None handling into the request context helpers/properties instead of burying it in a broad exception handler. Specifically, request_end now ignores None values, the context update helpers no-op for None, and __aexit__ no longer catches propagation exceptions. Added unit coverage for the original runner exception preservation scenario and the None-handling helpers. Local validation: py_compile passed for osbenchmark/context.py and tests/client_test.py, plus a direct context smoke test passed. Full pytest setup was blocked locally by cbor2/Rust target build failure on macOS, but DCO is now green.

OVI3D0:
Looks pretty good! can you just unblock this test `test_propagates_nested_context` as i believe it tests this exact functionality

I also think with these changes that the whole try/catch block in `__aexit__` can go away now

OVI3D0:
Tested this locally on a clean checkout off main.

Your v3 looks good and the 4 tests you added all pass. I added one more test where the parent doesn't call any `on_*_start/end` itself, so it has to rely entirely on the child propagating values up via `__aexit__`:

```python
async def test_nested_context_propagates_to_parent(self):
    test_client = client.RequestContextHolder()
    async with test_client.new_request_context() as top_level_ctx:
        async with test_client.new_request_context():
            test_client.on_client_request_start()
            test_client.on_request_start()
            await asyncio.sleep(0.01)
            test_client.on_request_end()
            test_client.on_client_request_end()
        self.assertIsNotNone(top_level_ctx.client_request_start)
        self.assertIsNotNone(top_level_ctx.client_request_end)
        self.assertIsNotNone(top_level_ctx.request_start)
        self.assertIsNotNone(top_level_ctx.request_end)
```

This one fails on the current branch. After the child exits the parent ctx is still empty, so all four assertions hit None.

The cause is the order of operations in `__aexit__`. `update_*()` reads `cls.request_context.get()`, which returns whichever ContextVar is currently bound. In v3 `restore_context` runs after the `update_*` calls, so the child is still bound when the writes happen. The writes land in the child dict, which then gets discarded.

Two changes get the test passing:

1. Move `restore_context` so it runs before the `update_*` calls. You need to snapshot the timing values into locals first since the `self.X` properties read from the child's `self.ctx`.

2. Reorder the `update_*` calls so `update_client_request_start` runs before `update_request_start`. `update_request_start` only writes when `client_request_start` is already in the parent's ctx (the existing scroll-handling guard), so if it runs first it silently does nothing.

```python
async def __aexit__(self, exc_type, exc_val, exc_tb):
    client_request_start = self.client_request_start
    client_request_end = self.client_request_end
    request_start = self.request_start
    request_end = self.request_end
    self.ctx_holder.restore_context(self.token)
    if self.token.old_value != contextvars.Token.MISSING:
        self.ctx_holder.update_client_request_start(client_request_start)
        self.ctx_holder.update_request_start(request_start)
        self.ctx_holder.update_request_end(request_end)
        self.ctx_holder.update_client_request_end(client_request_end)
    self.token = None
    return False
```

With that, the full suite passes (1422 tests, 0 failures). That should be the final change needed. Thanks for your patience here!

OVI3D0:
hey @thakoreh , looks like the CI checks are failing, can you just rebase and they should be green?

PR #1052, opened 2026-04-29
bug-fix:fix wrong doc count when retries enabled

rishabh6788:
### Description
Fix BulkVectorDataSet reporting wrong success count after doc-level retries. 

Summary

  - Fix a bug where BulkVectorDataSet runner only reported the success count from the final retry attempt, silently dropping docs that succeeded on earlier attempts
  - For example, a 500-doc bulk with 90 failures retried successfully would report success-count: 90 and weight: 90 instead of 500
  - Track cumulative successes across retry iterations and use the original total doc count for size and weight
  - Add tests covering: all-success (no retry), full recovery after retry, partial recovery, and no-retry-configured paths
### Issues Resolved
[List any issues this PR will resolve]

### Testing
- [ ] New functionality includes testing

[Describe how this change was tested]

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1053, opened 2026-04-29
fix vectorsearch retry doc counting

OVI3D0:
### Description
VectorSearch Bulk retry logic was miscounting how many docs were being ingested by the host. So if OSB sent 500 docs -> 410 succeed but 90 fail -> retry the 90 docs -> all 90 succeed, this would only count as 90 total docs ingested.

This was throwing off the number of successful docs which in turn throws off throughput calculations. This PR captures the total docs number from the original body *before* any retry loop, and uses that for weight and size. We also accumulate the number of successful docs across attempts which overwrites the per-attempt `success-count`

### Issues Resolved

### Testing

Adds 2 new unit tests specific to these changes

---


OVI3D0:
closing as this is a duplicate of #1052

PR #1073, opened 2026-05-21
Add radial search (max_distance/min_score) recall metrics support

Likhoram:
- Make k optional for radial search queries (max_distance/min_score)
- Route to correct ground truth context (MAX_DISTANCE_NEIGHBORS/MIN_SCORE_NEIGHBORS)
- Add validation that at least one of k/max_distance/min_score is specified
- Handle variable-length neighbor lists (filter -1 padding) for radial recall
- Wire up radial recall metrics through worker coordinator, metrics, and publisher
- Set size=10000 for radial queries to capture all results within threshold
- Add unit tests for radial search query construction and neighbor filtering

---

### DIFF
--- a/osbenchmark/workload/params.py
+++ b/osbenchmark/workload/params.py
@@ -1100,6 +1100,8 @@ class VectorSearchPartitionParamSource(VectorDataSetPartitionParamSource):
     PARAMS_NAME_OVERSAMPLE_FACTOR = "oversample_factor"
     PARAMS_NAME_RESCORE = "rescore"
     PARAMS_NAME_SPACE_TYPE = "space_type"
+    PARAMS_NAME_MAX_DISTANCE = "max_distance"
+    PARAMS_NAME_MIN_SCORE = "min_score"
 
     def __init__(self, workloads, params, query_params, **kwargs):
         super().__init__(workloads, params, Context.QUERY, **kwargs)
@@ -1116,15 +1118,12 @@ class VectorSearchPartitionParamSource(VectorDataSetPartitionParamSource):
         operation_type = parse_string_parameter(self.PARAMS_NAME_OPERATION_TYPE, params,
                                                 self.PARAMS_VALUE_VECTOR_SEARCH)
         self.oversample_factor = params.get(self.PARAMS_NAME_OVERSAMPLE_FACTOR)
-        has_radial = ("max_distance" in params or "min_score" in params
-                      or "max_distance" in query_params or "min_score" in query_params)
+        has_radial = (self.PARAMS_NAME_MAX_DISTANCE in params or self.PARAMS_NAME_MIN_SCORE in params
+                      or self.PARAMS_NAME_MAX_DISTANCE in query_params or self.PARAMS_NAME_MIN_SCORE in query_params)
         if self.oversample_factor and has_radial:
             raise exceptions.InvalidSyntax(
                 "'oversample_factor' cannot be used with 'max_distance' or 'min_score'. "
                 "Rescoring is only supported with top-k search.")
-        if "max_distance" in params and "min_score" in params:
-            raise exceptions.InvalidSyntax(
-                "'max_distance' and 'min_score' cannot be used together.")
         self.query_params = query_params
         self.query_params.update({
             self.PARAMS_NAME_OPERATION_TYPE: operation_type,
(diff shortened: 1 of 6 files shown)


osbenchmark/workload/params.py
```diff
@@ -1116,16 +1116,29 @@ def __init__(self, workloads, params, query_params, **kwargs):
         operation_type = parse_string_parameter(self.PARAMS_NAME_OPERATION_TYPE, params,
                                                 self.PARAMS_VALUE_VECTOR_SEARCH)
         self.oversample_factor = params.get(self.PARAMS_NAME_OVERSAMPLE_FACTOR)
-        if self.oversample_factor and ("max_distance" in query_params or "min_score" in query_params):
+        has_radial = ("max_distance" in params or "min_score" in params
+                      or "max_distance" in query_params or "min_score" in query_params)
```
shatejas: we need constants for max_distance and min_score

osbenchmark/workload/params.py
```diff
@@ -1116,16 +1116,29 @@ def __init__(self, workloads, params, query_params, **kwargs):
...
+        if self.oversample_factor and has_radial:
             raise exceptions.InvalidSyntax(
                 "'oversample_factor' cannot be used with 'max_distance' or 'min_score'. "
                 "Rescoring is only supported with top-k search.")
+        if "max_distance" in params and "min_score" in params:
+            raise exceptions.InvalidSyntax(
+                "'max_distance' and 'min_score' cannot be used together.")
```
shatejas: do we need this? cluster should throw an error right?

osbenchmark/workload/params.py
```diff
@@ -1116,16 +1116,29 @@ def __init__(self, workloads, params, query_params, **kwargs):
...
         self.query_params.update({
-            self.PARAMS_NAME_K: self.k,
             self.PARAMS_NAME_OPERATION_TYPE: operation_type,
             self.PARAMS_NAME_ID_FIELD_NAME: params.get(self.PARAMS_NAME_ID_FIELD_NAME),
         })
+        if self.k is not None:
+            self.query_params[self.PARAMS_NAME_K] = self.k
+        if "max_distance" in params:
+            self.query_params["max_distance"] = params["max_distance"]
+        if "min_score" in params:
```
shatejas: can we stub this out in a method and add a switch case instead?

osbenchmark/workload/params.py
```diff
@@ -1116,16 +1116,29 @@ def __init__(self, workloads, params, query_params, **kwargs):
...
+            self.query_params["min_score"] = params["min_score"]
+        if self.k is None and "max_distance" not in self.query_params and "min_score" not in self.query_params:
+            raise exceptions.InvalidSyntax(
+                "Must specify at least one of 'k', 'max_distance', or 'min_score'.")
```
shatejas: let cluster handle this maybe?

osbenchmark/workload/params.py
```diff
@@ -1166,7 +1174,10 @@ def _update_body_params(self, vector):
         # accept body params if passed from workload, else, create empty dictionary
         body_params = self.query_params.get(self.PARAMS_NAME_BODY) or dict()
         if self.PARAMS_NAME_SIZE not in body_params:
-            body_params[self.PARAMS_NAME_SIZE] = self.k
+            if self.k is not None:
+                body_params[self.PARAMS_NAME_SIZE] = self.k
+            elif "max_distance" in self.query_params or "min_score" in self.query_params:
+                body_params[self.PARAMS_NAME_SIZE] = 10000
```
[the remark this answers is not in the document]
shatejas: 10000 is pretty high the default size is 10, lets not add size parameter at all if its not provided. This will use Opensearch defaults

osbenchmark/workload/params.py
```diff
@@ -1191,8 +1207,15 @@ def partition(self, partition_index, total_partitions):
         if not self.neighbors_data_set_path:
             self.neighbors_data_set_path = self.data_set_path
         # add neighbor instance to partition
+        if "max_distance" in self.query_params:
+            neighbors_context = Context.MAX_DISTANCE_NEIGHBORS
+        elif "min_score" in self.query_params:
+            neighbors_context = Context.MIN_SCORE_NEIGHBORS
```
shatejas: I was just wondering if we need 2 separate datasets? for instance we can have a dataset for minScore 100 and then we should be able to use the same for maxDistance right

osbenchmark/workload/params.py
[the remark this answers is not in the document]
Likhoram: Removed the hardcoded size from the framework. Users can set it explicitly in query_body in their param file. I'll add "size": 10000 to the template param files in the workloads repo so radial search benchmarks get full results by default.

osbenchmark/workload/params.py
Likhoram: Both keys live in the same HDF5 file. neighbors_data_set_path points to one file containing both max_distance_neighbors and min_score_neighbors. The routing just reads the appropriate key.

shatejas: > Both keys live in the same HDF5 file. neighbors_data_set_path points to one file containing both max_distance_neighbors and min_score_neighbors. The routing just reads the appropriate key.

I understand that.  My question is more on why do we need 2 sets for ground truth. For instance if we have max distance of -160 or min score of 161 both should return the same set of neighbors right? can't we just keep one set so we don't duplicate

osbenchmark/workload/params.py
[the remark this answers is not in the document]
shatejas: > 'll add "size": 10000 to the template param files in the workloads repo so radial search benchmarks get full results by default.

For template as we 100 should be good

OVI3D0 (approved):
@shatejas if all looks good to you i can merge

osbenchmark/workload/params.py
Likhoram: Makes sense. Consolidated them into a single key radial_neighbors

OVI3D0:
@Likhoram lint check failing:
`osbenchmark/async_connection.py:69:8: E1125: Missing mandatory keyword argument 'stream_writer' in method call (missing-kwoa)`

shatejas (approved):

shatejas:
@Likhoram @OVI3D0 Functionally it looks good to me

OVI3D0:
> @Likhoram lint check failing: `osbenchmark/async_connection.py:69:8: E1125: Missing mandatory keyword argument 'stream_writer' in method call (missing-kwoa)`

not related to these changes, will fix separately. Thanks @Likhoram :)

PR #1087, opened 2026-07-15
Add per-query radial search threshold support

Likhoram:
### Description
- Adds `radial_search_type` parameter that enables per-query radial thresholds read from an enriched HDF5 dataset
- Each query gets its own `max_distance` or `min_score` value from a pre-computed threshold dataset, instead of using a fixed value for all queries
- Uses explicit `radial_engine` param ("faiss" or "lucene") to select the correct threshold dataset
- Coexists with existing fixed-threshold radial search (no breaking changes)

### Testing
- [x] All 35 existing vector search unit tests pass
- [x] 3 new unit tests pass
- [x] Faiss HNSW benchmark: recall=0.96, p50=9.6ms (k=100, cohere-10M)
- [x] Lucene HNSW benchmark: recall=0.85, p50=18ms (k=100, cohere-10M)
- [x] Faiss varying k: k=200 (0.95), k=500 (0.92), k=1000 (0.88)
- [x] Manual 10K-query validation — Faiss: 0.9599 (matches OSB 0.96)
- [x] Manual 10K-query validation — Lucene: 0.8447 (matches OSB 0.85)

---

### DIFF
--- a/osbenchmark/workload/params.py
+++ b/osbenchmark/workload/params.py
@@ -1164,11 +1164,7 @@ class VectorSearchPartitionParamSource(VectorDataSetPartitionParamSource):
         self.space_type = params.get(self.PARAMS_NAME_SPACE_TYPE, "l2")
 
         if self.radial_search_type:
-            index_body = params.get("target_index_body", "")
-            if "lucene" in index_body.lower():
-                self.radial_engine = "lucene"
-            else:
-                self.radial_engine = "faiss"
+            self.radial_engine = params.get("radial_engine", "faiss")
 
 
         if self.PARAMS_NAME_FILTER in params:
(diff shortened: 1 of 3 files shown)


osbenchmark/workload/params.py
```diff
@@ -1129,7 +1152,7 @@ def __init__(self, workloads, params, query_params, **kwargs):
             self.PARAMS_NAME_OPERATION_TYPE: operation_type,
             self.PARAMS_NAME_ID_FIELD_NAME: params.get(self.PARAMS_NAME_ID_FIELD_NAME),
         })
-        if self.k is not None:
+        if self.k is not None and self.radial_search_type is None:
```
kotwanikunal, line 1155: Wouldn't this break just the k based non radial approach?

Likhoram: No. When radial_search_type is not set (normal k-NN), self.radial_search_type is None evaluates to True, so the condition simplifies to if self.k is not None: same behavior as before.

kotwanikunal: Ahh. Makes sense

So we do k based search only when radial is not set - re-reading it makes sense. Thanks

kotwanikunal (approved):

rishabh6788 (approved):

PR #1096, opened 2026-07-25
Fix aggregate crashing on operations that report null metrics

### LINKED ISSUE #1093: [Bug]: aggregate fails with TypeError when an operation reports null metrics
### Describe the bug

`opensearch-benchmark aggregate` crashes for any set of test runs in which an operation has `null` values in its `min`/`max` metric fields:

```
[ERROR] ❌ Cannot aggregate. '<' not supported between instances of 'NoneType' and 'NoneType'.
[INFO] ❌ FAILURE (took 1 seconds)
```

`calculate_weighted_average` (`osbenchmark/aggregator.py:225`, `:227`) reduces `min`/`max` with `value.get(metric_field, 0)`. The `0` default only applies when the key is **absent** — it does not help when the key is **present with value `None`**, which is what the stored results contain. `min()` then compares `None` to `None` and raises.

An operation that produced no valid samples reports its metrics as null, so any campaign containing one cannot be aggregated. The `max` branch has the same problem, and the percentile/median branch below it (`value * iterations` after a `.get(..., 0)`) raises on `None` for the same reason.

There is a second, independent occurrence in `calculate_rsd` (`osbenchmark/aggregator.py:239`), on the path from `build_aggregated_results_dict` that passes the per-run mean values:

```
TypeError: can't convert type 'NoneType' to numerator/denominator
```

Fixing only the first site moves the crash rather than removing it. I found the second one only by running against real data after unit tests for the first were already green.

Aggregating N runs is the documented way to get a trustworthy result out of a noisy environment, so in practice this blocks multi-run campaigns whenever any operation comes back with no samples.

### To reproduce

1. Have two or more test runs where at least one operation reports null metrics, e.g.

   ```json
   { "task": "node-stats",
     "error_rate": 1.0,
     "throughput": { "min": null, "mean": null, "median": null, "max": null, "unit": "ops/s" } }
   ```

2. Aggregate them:

   ```
   opensearch-benchmark aggregate --test-runs=<id1>,<id2>
   ```

3. It fails immediately, and nothing is written.

Minimal reproduction against the real method, if it helps — the code under test is not mocked, only the cfg/store `__init__` wiring is bypassed:

```python
from osbenchmark.aggregator import Aggregator
agg = Aggregator.__new__(Aggregator)
agg.test_runs = {"run1": None, "run2": None}
agg.accumulated_iterations = {"run1": {"op1": 1}, "run2": {"op1": 1}}
bad = {"min": None, "mean": None, "median": None, "max": None, "unit": "ops/s"}
agg.calculate_weighted_average({"throughput": [bad, bad]}, "op1")   # TypeError

agg.calculate_rsd([None, None], "throughput")                        # TypeError, second site
```

A healthy operation with numeric values aggregates fine, which is what points at `None` as the trigger rather than anything about the environment.

### Expected behavior

`aggregate` completes and produces aggregated results. For an operation with no valid samples I would expect the aggregate to carry `null` through for that metric — honest, it says "not measured" — rather than substituting `0`, which would read as a real measurement of zero throughput. RSD over no values would report `NA`, as it already does for the single-value case.

Note that `aggregate_json_by_key` in the same file already skips nulls deliberately (`next((obj for obj in json_elements if obj is not None), None)`), so handling null metrics is an established pattern in this file; `calculate_weighted_average` just doesn't do it.

### Host / Environment

- OpenSearch Benchmark `main` at `a5880562` (2.4.0), installed from source, Python 3.12
- macOS 15 (arm64)
- Reproducible with any two test runs containing a null-metric operation

### Additional context

`git log -L225,225:osbenchmark/aggregator.py` traces these lines to #638 ("Add aggregate command", `c6ec0a11`) and #692 ("Change min/max to overall_min/overall_max", `2d14ed40`).

Worth flagging for the project generally: `tests/aggregator_test.py` never constructs a real `TestRun` and passes bare `Mock()` test runs throughout, so the suite is green (1422 passed, 5 skipped on `main`) over a command that cannot complete. That matters more for the related attribute defect I'm reporting separately, but it is why neither was caught.

I hit this in Apache solr-orbit, which is a Python port of OSB — its `aggregator.py` is byte-identical to this one apart from the import module name, so all of this is inherited rather than port-specific. Reported there as apache/solr-orbit#55.

### Relevant log output

```shell
  File ".../osbenchmark/benchmark.py", line 1189, in dispatch_sub_command
    aggregator_instance.aggregate()
  File ".../osbenchmark/aggregator.py", line 284, in aggregate
    aggregated_results = self.build_aggregated_results()
  File ".../osbenchmark/aggregator.py", line 176, in build_aggregated_results
    aggregated_results = self.build_aggregated_results_dict()
  File ".../osbenchmark/aggregator.py", line 129, in build_aggregated_results_dict
    aggregated_task_metrics = self.calculate_weighted_average(task_metrics, task)
  File ".../osbenchmark/aggregator.py", line 225, in calculate_weighted_average
    weighted_metrics[metric]['overall_min'] = min(value.get(metric_field, 0) for value in values)
TypeError: '<' not supported between instances of 'NoneType' and 'NoneType'
```

serhiy-bzhezytskyy:
### Description

`calculate_weighted_average` reduced `min`/`max` with `value.get(metric_field, 0)`. That default only applies when the key is *absent*, so a key present with value `None` reached `min()`/`max()` and raised `TypeError: '<' not supported between instances of 'NoneType' and 'NoneType'`. The percentile/median branch just below had the same flaw in a different shape — `value * iterations` on `None`.

`calculate_rsd` is a second, independent site, reached from `build_aggregated_results_dict` with the per-run mean values (`TypeError: can't convert type 'NoneType' to numerator/denominator`). Fixing only the first one moves the crash rather than removing it — that is how I found the second, by running against real data after unit tests for the first were already green.

The change:

- extracts the weighted mean into a `weighted_mean` helper. The divisor now depends on which runs contributed a value, so it can no longer be one total summed up front, and the dict and scalar branches were computing it two different ways;
- leaves nulls out of both the weighted sum and its divisor, so a run that measured nothing does not drag the mean toward zero;
- propagates `None` when no run contributed a value, rather than substituting `0`. `0` would read as "throughput was zero", which is a different claim from "not measured". This matches how `aggregate_json_by_key` in the same file already treats nulls;
- returns `NA` from `calculate_rsd` when no values remain, as it already does for the single-value case.

### Issues Resolved

Resolves #1093

### Testing

- Two new tests in `tests/aggregator_test.py`: all-null metric fields, and partially null — a metric with samples in one run but not another, where the valid values must still aggregate, weighted only by the runs that contributed them.
- Both verified red without the fix, by restoring `main`'s `aggregator.py` under the new tests rather than by inspection.
- Full suite: `1424 passed, 5 skipped` (baseline on `main` is 1422). `pylint` clean.
- End-to-end through the real CLI, on test-run files written by OSB's own `TestRun.as_dict()`: `aggregate` goes from `❌ FAILURE` to `✅ SUCCESS`, and the output is correct rather than merely non-crashing — a healthy operation aggregates to `overall_min 40000.0, mean 42250.0, median 42150.0, overall_max 44500.0` with `mean_rsd 0.8368`, while the null operation reports `null` throughput and `mean_rsd "NA"`.

The `calculate_rsd` site has no dedicated unit test here; it is covered by the end-to-end run, where it was the crash that surfaced once the first site was fixed. I can add a direct one if you'd prefer it in the suite.

### Notes

Found while using Apache solr-orbit, a Python port of OSB whose `aggregator.py` is byte-identical to this one apart from the import module name, to run a multi-configuration benchmark campaign. The same fix is proposed there as apache/solr-orbit#58. Reported here because the defect is upstream, not port-specific.

### DIFF
--- a/osbenchmark/aggregator.py
+++ b/osbenchmark/aggregator.py
@@ -213,7 +213,6 @@ class Aggregator:
         # Get iterations for each test run
         iterations_per_run = [self.accumulated_iterations[test_id][task_name]
                                     for test_id in self.test_runs.keys()]
-        total_iterations = sum(iterations_per_run)
 
         for metric, values in task_metrics.items():
             if isinstance(values[0], dict):
@@ -222,23 +221,44 @@ class Aggregator:
                     if metric_field == 'unit':
                         weighted_metrics[metric][metric_field] = values[0][metric_field]
                     elif metric_field == 'min':
-                        weighted_metrics[metric]['overall_min'] = min(value.get(metric_field, 0) for value in values)
+                        item_values = [value.get(metric_field) for value in values]
+                        weighted_metrics[metric]['overall_min'] = min(
+                            (value for value in item_values if value is not None), default=None)
                     elif metric_field == 'max':
-                        weighted_metrics[metric]['overall_max'] = max(value.get(metric_field, 0) for value in values)
+                        item_values = [value.get(metric_field) for value in values]
+                        weighted_metrics[metric]['overall_max'] = max(
+                            (value for value in item_values if value is not None), default=None)
                     else:
                         # for items like median or containing percentile values
-                        item_values = [value.get(metric_field, 0) for value in values]
-                        weighted_sum = sum(value * iterations for value, iterations in zip(item_values, iterations_per_run))
-                        weighted_metrics[metric][metric_field] = weighted_sum / total_iterations
+                        item_values = [value.get(metric_field) for value in values]
+                        weighted_metrics[metric][metric_field] = self.weighted_mean(item_values, iterations_per_run)
             else:
-                weighted_sum = sum(value * iterations for value, iterations in zip(values, iterations_per_run))
-                weighted_metrics[metric] = weighted_sum / total_iterations
+                weighted_metrics[metric] = self.weighted_mean(values, iterations_per_run)
 
         return weighted_metrics
+
+    @staticmethod
+    def weighted_mean(values: List[Any], iterations_per_run: List[int]) -> Any:
+        """
+        Weights each test run's value by that run's iteration count.
+        Operations that produced no valid samples report their metrics as None; those are left out
+        of both the sum and the divisor, and the result is None when no run contributed a value,
+        so that "not measured" stays distinct from a measured zero
+        """
+        contributions = [(value, iterations) for value, iterations in zip(values, iterations_per_run)
+                         if value is not None]
+        total_iterations = sum(iterations for _, iterations in contributions)
+        if not total_iterations:
+            return None
+        return sum(value * iterations for value, iterations in contributions) / total_iterations
 
     def calculate_rsd(self, values: List[Union[int, float]], metric_name: str) -> Union[float, str]:
         if not values:
             raise ValueError(f"Cannot calculate RSD for metric '{metric_name}': empty list of values")
+        # operations that produced no valid samples report None, which cannot contribute to a deviation
+        values = [value for value in values if value is not None]
+        if not values:
+            return "NA"  # no test run measured this metric
         if len(values) == 1:
             return "NA"  # RSD is not applicable for a single value
         mean = statistics.mean(values)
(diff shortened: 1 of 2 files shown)


OVI3D0 (approved):
@serhiy-bzhezytskyy tested and verified this works, LGTM, but looks like there is just a conflict on the test file if you can do a quick rebase

serhiy-bzhezytskyy:
Rebased onto main — the conflict was in `tests/aggregator_test.py`, where #1097 and this PR each appended a test after the same anchor. Kept both; the suite is at 1425 passed / 5 skipped locally.

Thanks for reviewing and merging #1097.

PR #939, opened 2025-08-19
fixes for aggregate command

OVI3D0:
### Description
Fixes to aggregate command to prevent key errors + update function name for storing aggregated results

### Issues Resolved
[List any issues this PR will resolve]

### Testing
- [x] New functionality includes testing

running aggregate command with changes applied

---
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1097, opened 2026-07-25
Fix aggregate reading a TestRun attribute that does not exist

### LINKED ISSUE #1094: [Bug]: aggregate always fails — update_config_object reads a TestRun attribute that does not exist
### Describe the bug

`opensearch-benchmark aggregate` fails for **any** input:

```
[ERROR] ❌ Cannot aggregate. 'TestRun' object has no attribute 'cluster_config_instance_params'.
[INFO] ❌ FAILURE (took 0 seconds)
```

`update_config_object` (`osbenchmark/aggregator.py:169`) reads `test_run.cluster_config_instance_params`. `TestRun` defines `cluster_config_params` (`osbenchmark/metrics.py:1343`). The name being read exists nowhere else in the repository — nothing assigns it:

```
$ grep -rn 'cluster_config_instance_params' . --include='*.py'
osbenchmark/aggregator.py:169        # the only occurrence, and it is a read
```

This looks like a half-applied rename. #882 (`2c85d505`, "Rebase renamed-components onto latest developments") renamed the attribute on `TestRun` to `cluster_config_params` but renamed this read site to `cluster_config_instance_params`. #939 (`e10342d1`, "fixes for aggregate command") later corrected the *config key* on the same line, from `cluster_config_instance.params` to `cluster_config.params`, and left the attribute name as it was.

Unlike the null-metrics crash I'm reporting separately, this one does not depend on the data. It is on the plain `aggregate` → `build_aggregated_results` → `update_config_object` path, so it fires on every invocation. As far as I can tell that means `aggregate` cannot currently succeed for anyone on `main`.

A side effect worth noting: because the read never succeeds, the cluster-config params never reach the aggregated result at all.

### To reproduce

```
opensearch-benchmark aggregate --test-runs=<id1>,<id2>
```

Any two compatible test runs will do.

If the runs contain an operation with null metrics, that crash happens first during metric reduction — which is why this one only becomes visible once that is fixed.

### Expected behavior

`aggregate` reads `test_run.cluster_config_params` and completes, and the cluster-config params reach the aggregated result. With that one name corrected, my aggregates carry `cluster-config-instance-params: {"heap_size": "6g"}` as intended.

### Host / Environment

- OpenSearch Benchmark `main` at `a5880562` (2.4.0), installed from source, Python 3.12
- macOS 15 (arm64)
- Reproducible with any two test runs

### Additional context

Why the test suite doesn't catch it, since that seems worth flagging for the project generally: `tests/aggregator_test.py` never constructs a real `TestRun` — every test passes a bare `Mock()`, and a `Mock` returns a fresh `Mock` for any attribute name asked of it, so the read succeeds regardless of the name. `main` is at 1422 passed, 5 skipped while `aggregate` cannot run once.

`Mock(spec=TestRun)` does not help either: `spec` inspects the class, which knows nothing about attributes assigned in `__init__`, so it raises on `cluster_config` first and points at the wrong line. A regression test for this has to construct a real `TestRun`.

On the storage side, for context: `metrics.py` writes the legacy hyphenated JSON key `cluster-config-instance-params` (`:1409`, `:1441`), and `from_dict` maps it back onto `cluster_config_params`. So the on-disk key, the attribute, and the name being read are three different spellings, and only the third one has no definition behind it.

I hit this in Apache solr-orbit, which is a Python port of OSB — its `aggregator.py` is byte-identical to this one apart from the import module name, so this is inherited rather than port-specific. Reported there as apache/solr-orbit#56.

### Relevant log output

```shell
  File ".../osbenchmark/aggregator.py", line 193, in build_aggregated_results
    self.update_config_object(test_run)
  File ".../osbenchmark/aggregator.py", line 169, in update_config_object
    "cluster_config.params", test_run.cluster_config_instance_params)
AttributeError: 'TestRun' object has no attribute 'cluster_config_instance_params'.
  Did you mean: 'cluster_config_instance'?
```

serhiy-bzhezytskyy:
### Description

`update_config_object` read `test_run.cluster_config_instance_params`, which no test run has, so `aggregate` failed on every invocation:

```
[ERROR] ❌ Cannot aggregate. 'TestRun' object has no attribute 'cluster_config_instance_params'.
```

`TestRun` defines `cluster_config_params` (`osbenchmark/metrics.py:1343`), and the name being read occurs nowhere else in the repository — nothing assigns it. #882 (`2c85d505`) renamed the attribute on `TestRun` to `cluster_config_params` but renamed this read site to `cluster_config_instance_params`; #939 (`e10342d1`) later corrected the *config key* on that same line, from `cluster_config_instance.params` to `cluster_config.params`, and left the attribute name as it was. This one-word change reads the name `TestRun` actually defines.

Unlike the null-metrics crash (#1093), this does not depend on the data — it is on the plain `aggregate` → `build_aggregated_results` → `update_config_object` path. As far as I can tell it means `aggregate` cannot currently succeed for anyone on `main`.

A side effect of the fix: because the read never succeeded, the cluster-config params never reached the aggregated result. They now do.

### Issues Resolved

Resolves #1094

### Testing

- One new test, `test_update_config_object_reads_attributes_that_test_runs_actually_have`, which constructs a **real** `metrics.TestRun` and asserts the `cluster_config.params` value reaches `config.add`.
- Verified red without the fix: with `main`'s `aggregator.py` restored under the new test it fails with `AttributeError: 'TestRun' object has no attribute 'cluster_config_instance_params'`; with the fix it passes.
- Full suite: `1423 passed, 5 skipped` (baseline on `main` is 1422). `pylint` clean.
- End-to-end through the real CLI: `aggregate` reports `✅ SUCCESS` and the aggregated result now carries `cluster-config-instance-params: {'heap_size': '6g'}`, which was being silently lost.

The reason the existing suite doesn't catch this is worth flagging on its own: `tests/aggregator_test.py` never constructs a real `TestRun` — every test passes a bare `Mock()`, and a `Mock` returns a fresh `Mock` for any attribute name asked of it, so a wrong name reads fine. `main` sits at 1422 passed while `aggregate` cannot complete once. `Mock(spec=TestRun)` does not fix it either, because `spec` inspects the class, which knows nothing about attributes assigned in `__init__`; it raises on `cluster_config` first and points at the wrong line. A real `TestRun` is the only thing that catches it, which is why the new test builds one.

### Notes

Found while using Apache solr-orbit, a Python port of OSB whose `aggregator.py` is byte-identical to this one apart from the import module name, to run a multi-configuration benchmark campaign. The same fix is proposed there as apache/solr-orbit#59.

This was the second of three defects blocking `aggregate`; it only became visible after the null-metrics crash was fixed, because that one happens earlier on the same path.

### DIFF
--- a/osbenchmark/aggregator.py
+++ b/osbenchmark/aggregator.py
@@ -166,7 +166,7 @@ class Aggregator:
         self.config.add(config.Scope.applicationOverride, "test_run", "pipeline", test_run.pipeline)
         self.config.add(config.Scope.applicationOverride, "workload", "params", test_run.workload_params)
         self.config.add(config.Scope.applicationOverride, "builder",
-                        "cluster_config.params", test_run.cluster_config_instance_params)
+                        "cluster_config.params", test_run.cluster_config_params)
         self.config.add(config.Scope.applicationOverride, "builder", "plugin.params", test_run.plugin_params)
         self.config.add(config.Scope.applicationOverride, "workload", "latency.percentiles", test_run.latency_percentiles)
         self.config.add(config.Scope.applicationOverride, "workload", "throughput.percentiles", test_run.throughput_percentiles)
--- a/tests/aggregator_test.py
+++ b/tests/aggregator_test.py
@@ -1,6 +1,7 @@
 from unittest.mock import Mock
 import pytest
 from osbenchmark import config
+from osbenchmark import metrics
 from osbenchmark.aggregator import Aggregator, AggregatedResults
 
 @pytest.fixture
@@ -119,6 +120,21 @@ def test_calculate_weighted_average(aggregator):
     assert result["throughput"] == 160  # (100*2 + 200*3) / (2+3)
     assert result["latency"]["avg"] == 16  # (10*2 + 20*3) / (2+3)
     assert result["latency"]["unit"] == "ms"
+
+def test_update_config_object_reads_attributes_that_test_runs_actually_have(aggregator):
+    # a real TestRun, not a Mock: a Mock creates whatever attribute is asked of it, so it cannot
+    # tell us whether update_config_object reads attributes that a test run really carries
+    test_run = metrics.TestRun(
+        benchmark_version="1.0.0", benchmark_revision="abc123", environment_name="unit-test",
+        test_run_id="test1", test_run_timestamp="20250101T000000Z", pipeline="benchmark-only",
+        user_tags={}, workload="workload1", workload_params={}, test_procedure="test_proc1",
+        cluster_config=["external"], cluster_config_params={"heap_size": "6g"}, plugin_params={},
+        meta_data={})
+
+    aggregator.update_config_object(test_run)
+
+    aggregator.config.add.assert_any_call(config.Scope.applicationOverride, "builder",
+                                          "cluster_config.params", {"heap_size": "6g"})
 
 def test_calculate_rsd(aggregator):
     values = [1, 2, 3, 4, 5]


OVI3D0 (approved):
LGTM, thanks @serhiy-bzhezytskyy !

PR #1098, opened 2026-07-25
Let aggregate load a workload from --workload-path

### LINKED ISSUE #1095: aggregate has no --workload-path, so runs made with --workload-path can't be aggregated

`run` can take its workload either from a repository or from a local path — `add_workload_source()` (`osbenchmark/benchmark.py:78`) adds a mutually-exclusive `--workload-repository` / `--workload-path` pair, and it is applied to `list` (`:132`), `info` (`:135`) and the test-run parser (`:541`).

`aggregate` isn't in that list. It defines only its own `--workload-repository` (`:324`), so there is no way to point it at a local workload directory, even though `aggregate()` does need to load the workload — it reads the test procedure's schedule to get per-task iteration counts, via `workload.load_workload` and `repository.name`.

The effect is that a workflow which is fully supported for a test run stops working at aggregate time:

```
$ opensearch-benchmark run --workload-path=/path/to/my-workload ...               # works
$ opensearch-benchmark aggregate --test-runs=<id1>,<id2> --workload-path=/path/to/my-workload
opensearch-benchmark: error: unrecognized arguments: --workload-path=/path/to/my-workload
```

So a workload developed locally and benchmarked with `--workload-path` — the "create your own workload" path in the docs — can be run but its results can't be aggregated. Cloning the `default` workloads repository works fine and is the available workaround, so this is an inconvenience rather than a blocker.

Call `add_workload_source(aggregate_parser)` and drop the hand-rolled `--workload-repository`, since the shared helper already provides it plus `--workload-revision`. That makes `aggregate` accept a workload source the same way `run` does, keeps the two options mutually exclusive, and lets it reuse `configure_workload_params` for path normalisation and validation.

One behavioural detail: a path sets `workload.path`, which is what makes the loader choose `SimpleWorkloadRepository`, so `aggregate()` should skip setting `repository.name` when a path was given — otherwise the loader goes looking for the workload inside that repository instead. That guard is load-bearing, since argparse's default for `--workload-repository` is the string `"default"` rather than `None`.

Going further: a test run already records the workload it used, so `aggregate` could resolve the spec from the stored run and not need a workload-source argument at all. That is arguably the better shape, but it is a larger change and would need a decision about what to do when the stored workload is no longer available. Starting with the parity fix matches the existing pattern.

### Additional context

Came up while running a multi-configuration campaign, before I worked out that the documented `default` repository clone satisfies `aggregate` on its own.

I hit this in Apache solr-orbit, which is a Python port of OSB with the same argparse structure, so it is inherited rather than port-specific. Reported there as apache/solr-orbit#57.

serhiy-bzhezytskyy:
### Description

`add_workload_source()` offers a workload either as a repository or as a local path, and `run`, `list` and `info` all use it. `aggregate` defined only its own `--workload-repository`, so a test run made against a workload on disk could not be aggregated afterwards — the aggregator loads the workload to read its schedule, and a repository lookup fails for a workload that only exists as a path:

```
$ opensearch-benchmark aggregate --test-runs=<id1>,<id2> --workload-path=/path/to/workload
opensearch-benchmark: error: unrecognized arguments: --workload-path=/path/to/workload
```

The aggregate parser now uses the same `add_workload_source` helper as `run`, so the two options stay mutually exclusive and `--workload-revision` comes along with them. Dispatch calls `configure_workload_params` with `command_requires_workload=False`, matching how `list` already handles a command that takes a workload source but no `--workload` — `aggregate` gets the workload name from the stored test run.

One behavioural detail: a path sets `workload.path`, which is what makes the loader choose `SimpleWorkloadRepository`. So when a path was given, `aggregate()` no longer names a repository — doing both would send the loader looking for the workload inside that repository instead. That guard is load-bearing, since argparse's default for `--workload-repository` is the string `"default"` rather than `None`.

### Issues Resolved

Resolves #1095

### Testing

- Two new tests: `test_aggregate_names_the_workload_repository` (unchanged behaviour when a repository is used) and `test_aggregate_leaves_a_workload_path_alone` (no `repository.name` is configured when a path is given).
- The shared `mock_args` fixture previously supplied no `workload_path`, so the aggregator saw a `Mock` attribute that happens to be truthy. It is now explicit — that fixture change is what makes the first test meaningful rather than accidentally passing.
- CLI verified by hand: `aggregate --workload-path=<dir>` is accepted where it previously errored, aggregation succeeds end to end, and combining it with `--workload-revision` is correctly rejected.
- Full suite: `1424 passed, 5 skipped` (baseline on `main` is 1422). `pylint` clean.

### Notes

Found while using Apache solr-orbit, a Python port of OSB with the same argparse structure, to run a multi-configuration benchmark campaign. The same change is proposed there as apache/solr-orbit#60. Third of three `aggregate` gaps I hit, and the only one that is a parity gap rather than a bug.

There is a case for going further — a test run already records the workload it used, so `aggregate` could resolve the spec from the stored run and need no workload-source argument at all. I've gone with the parity fix because it is the smaller change and matches the existing pattern, but I'm open to the other shape if you'd rather have it.

### DIFF
--- a/osbenchmark/aggregator.py
+++ b/osbenchmark/aggregator.py
@@ -292,7 +292,10 @@ class Aggregator:
         if self.test_run_compatibility_check():
             self.test_run = self.test_store.find_by_test_run_id(list(self.test_runs.keys())[0])
             self.test_procedure_name = self.test_run.test_procedure
-            self.config.add(config.Scope.applicationOverride, "workload", "repository.name", self.args.workload_repository)
+            # a workload given as a path is already configured; naming a repository too would send the
+            # loader looking for the workload in that repository instead
+            if not self.args.workload_path:
+                self.config.add(config.Scope.applicationOverride, "workload", "repository.name", self.args.workload_repository)
             self.config.add(config.Scope.applicationOverride, "workload", "workload.name", self.test_run.workload)
             self.loaded_workload = workload.load_workload(self.config)
             for id in self.test_runs.keys():
--- a/osbenchmark/benchmark.py
+++ b/osbenchmark/benchmark.py
@@ -321,10 +321,7 @@ def create_arg_parser():
         "--results-file",
         help="Write the aggregated results to the provided file.",
         default="")
-    aggregate_parser.add_argument(
-        "--workload-repository",
-        help="Define the repository from where OSB will load workloads (default: default).",
-        default="default")
+    add_workload_source(aggregate_parser)
 
     download_parser = subparsers.add_parser("download", help="Downloads an artifact")
     download_parser.add_argument(
@@ -1184,6 +1181,7 @@ def dispatch_sub_command(arg_parser, args, cfg):
             cfg.add(config.Scope.applicationOverride, "reporting", "percentiles", args.percentiles)
             publisher.compare(cfg, args.baseline, args.contender)
         elif sub_command == "aggregate":
+            configure_workload_params(arg_parser, args, cfg, command_requires_workload=False)
             test_runs_dict = prepare_test_runs_dict(args, cfg)
             aggregator_instance = aggregator.Aggregator(cfg, test_runs_dict, args)
             aggregator_instance.aggregate()
--- a/tests/aggregator_test.py
+++ b/tests/aggregator_test.py
@@ -1,4 +1,4 @@
-from unittest.mock import Mock
+from unittest.mock import Mock, patch
 import pytest
 from osbenchmark import config
 from osbenchmark import metrics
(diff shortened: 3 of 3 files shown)


OVI3D0 (approved):
LGTM but there are conflicts from the other PR(s) so we'll just need to do another quick rebase before i merge. Thank you @serhiy-bzhezytskyy !

OVI3D0:
@serhiy-bzhezytskyy this will need one more rebase because of #1096 :)

serhiy-bzhezytskyy:
Rebased onto main with #1096 in.

PR #1100, opened 2026-08-03
Revert to execute-test / test_execution terminology (3.0)

OVI3D0:
### Description

First of the stacked OpenSearch Benchmark 3.0 PRs. This reverts the 2.x-era terminology back to the 3.0 (and original 1.x) vocabulary, with full back-compat so existing automation and historical data keep working.

**CLI**
- `execute-test` is the primary subcommand again; `run` / `execute` remain as deprecated aliases (they warn, then dispatch).
- `--test-execution-id` is primary; `--test-run-id` remains as a legacy alias.

**Internal rename**
- `test_run` / `TestRun` → `test_execution` / `TestExecution` across metrics, aggregator, worker_coordinator, builder, paths, publisher, and the orchestrator (module `test_run_orchestrator.py` → `test_execution_orchestrator.py`).
- On disk: `test-runs/` → `test-executions/`; `test_run.json` → `test_execution.json`.
- Datastore (OpenSearch): indices `benchmark-test-runs-*` → `benchmark-test-executions-*`; stored field `test-run-id` → `test-execution-id`.

**Back-compat — reads accept BOTH, writes emit the new vocabulary**
- `TestExecution.from_dict` accepts either `test-execution-id`/`test-run-id` and `test-execution-timestamp`/`test-run-timestamp`.
- The node-stats CPU-window query and `list` accept both old and new names.
- **Seamless 1.x → 3.0 migration:** the reader also accepts genuine 1.x artifacts — the pre-2.x `provision-config-instance` (+ `-params`, + `cluster.provision-config-revision`) field names as fallbacks for the reverted `cluster-config-*` names, and the 1.x on-disk `test_executions/` (underscore) directory layout. Without this, a 1.x record silently dropped from `list` on the file path and crashed `list` on the OpenSearch-datastore path (both funnel through `from_dict`). Reads only — 3.0 always writes the new vocabulary.

No behavior change beyond naming and widened back-compat reads. Multi-engine support and the CloudWatch datastore build on this foundation in subsequent stacked PRs.

**Note for reviewers**
- `version.txt` is intentionally left at 2.4.0; the 3.0.0 bump lands in its own dedicated PR at release time (matching repo convention, e.g. #1090).
- Pre-existing minor flag inconsistency, not introduced here: `aggregate` uses `--test-executions-id` (plural) while `execute-test` uses `--test-execution-id` (singular); both have working legacy aliases. Can be smoothed here or left.

### Issues Resolved

Restores the original 1.x `execute-test` / `test_execution` vocabulary that 2.x had renamed.

### Testing

- **Unit:** full suite passes (1435 passed / 5 skipped; engine + metrics_stores suites belong to later stacked PRs). Includes 13+ back-compat assertions exercising the legacy `test-run-id` key, plus two new tests for genuine 1.x artifacts (`from_dict` parsing the `provision-config-*` vocab, and reading the 1.x underscore on-disk layout).
- **Live e2e** on real AWS infra (OpenSearch 2.19.1): full `execute-test` write→read round-trip confirming the new `test-executions/` dir, `test_execution.json`, and `test-execution-id`/`test-execution-timestamp` fields with no legacy keys emitted; OpenSearch-datastore path creating `benchmark-{metrics,results,test-executions}-*` indices with all metric docs queryable by `test-execution-id`; read-both proven on a live index (synthetic legacy `test-run-id` doc matched by the back-compat query); `list` / `compare` / `aggregate` / `visualize` exercised via both new and legacy flags; a full (non-test-mode) benchmark producing real throughput/latency.
- **1.x → 3.0 migration proven live:** genuine 1.x artifacts produced by OpenSearch Benchmark 1.18.0 (on-disk `test_executions/` underscore dir + a datastore run doc carrying `provision-config-instance`) were read by this branch — `list`, `find`, and `compare` all succeed on the on-disk run, and the datastore `list` reads the 1.x doc without the previous `from_dict` crash.

---

### DIFF
--- a/osbenchmark/builder/builder.py
+++ b/osbenchmark/builder/builder.py
@@ -98,7 +98,8 @@ def install(cfg):
 
 def start(cfg):
     root_path = paths.install_root(cfg)
-    test_run_id = cfg.opts("system", "test_run.id")
+    test_execution_id = cfg.opts("system", "test_execution.id", mandatory=False) \
+        or cfg.opts("system", "test_run.id")
     # avoid double-launching - we expect that the node file is absent
     with contextlib.suppress(FileNotFoundError):
         _load_node_file(root_path)
@@ -115,7 +116,7 @@ def start(cfg):
     else:
         raise exceptions.SystemSetupError("Unknown build type [{}]".format(node_config.build_type))
     nodes = node_launcher.start([node_config])
-    _store_node_file(root_path, (nodes, test_run_id))
+    _store_node_file(root_path, (nodes, test_execution_id))
 
 
 def stop(cfg):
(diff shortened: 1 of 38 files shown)


rishabh6788 (commented):
LGTM, hold the merge for now.

rishabh6788 (approved):

osbenchmark/builder/builder.py
```diff
@@ -98,7 +98,8 @@ def install(cfg):
 
 def start(cfg):
     root_path = paths.install_root(cfg)
-    test_run_id = cfg.opts("system", "test_run.id")
+    test_execution_id = cfg.opts("system", "test_execution.id", mandatory=False) \
```
rishabh6788, line 101: Can you explain what is happening here? Shouldn't `test-execution-id` be mandatory?

OVI3D0: yeah it is still mandatory, but `mandatory=False` is incase the lookup fails so we can fall back to the old `test_run_id` key

PR #1113, opened 2026-08-27
Bump base image to python:3.11-slim-trixie to clear sqlite3 CVEs

### LINKED ISSUE #1112: [Bug]: Base image python:3.11.2-slim ships already-public sqlite3 CVEs with no fix on its Debian release
### Describe the bug

These are already-public CVEs in a base image dependency, not a newly discovered flaw in opensearch-benchmark itself, so I am raising it as a public issue per the security policy's exception for already-disclosed issues.

The published opensearch-benchmark container image is based on `python:3.11.2-slim`, which resolves to Debian 11 (bullseye). The `sqlite3` package on bullseye is affected by CVE-2025-6965 and CVE-2025-7458. Neither has a released fix on that Debian version, so a container image scanner flags the base layer and the findings cannot be cleared by patching in place.

Moving both `FROM` stages to a current Debian 13 (trixie) Python base clears both. I verified `python:3.11-slim-trixie` with Trivy:

- **19** total findings
- **3** fixable HIGH
- **0** CRITICAL
- the two `sqlite3` CVEs above no longer present

The image scanned was:

```
python:3.11-slim-trixie@sha256:be1575ed968de893bd54f4c56315ff7c4736ce522c1bca08fd521731aafc0d76
```

(Python `3.11.16` on this image.) Python stays at 3.11 and `sqlite3` moves to `3.46.1-7+deb13u1`, which carries the fixes.

No urgency on my personal use case. We already built our own image internally on a different base for our own use. Raising it mainly in case it helps the broader user base. 

### To reproduce

1. Pull the published image: `docker pull opensearchproject/opensearch-benchmark:latest` (or build from the repo Dockerfile).
2. Scan it with Trivy: `trivy image opensearchproject/opensearch-benchmark:latest`
3. Observe HIGH findings against `sqlite3` for `CVE-2025-6965` and `CVE-2025-7458` with no fixed version available on the base image's Debian release.

### Expected behavior

The published image ships a base layer free of known, already-fixed HIGH or CRITICAL CVEs. In particular, the bundled `sqlite3` is a version that already includes the fixes for CVE-2025-6965 and CVE-2025-7458, so pulling and running the published image does not carry these known vulnerabilities.

### Host / Environment

Image: `opensearchproject/opensearch-benchmark:latest` (also published to the public ECR gallery).
Base layer: `python:3.11.2-slim`, which resolves to Debian 11 (bullseye).
Scanner: Trivy (image scan). The finding is on the base layer, so it reproduces on any host that can pull and scan the image, independent of the runtime OS.

### Additional context

Suggested fix: move both `FROM` lines to `python:3.11-slim-trixie`. Using the floating trixie tag lets routine rebuilds keep picking up base patches. If reproducible builds are preferred, it can be pinned by digest instead, ideally paired with Dependabot so the pin still advances.

Optional process follow ups, separate from the base bump, that would keep this from recurring:

1. A Trivy or equivalent image scan gate in CI on the published image.
2. Dependabot on the base image tag.
3. Publishing an SBOM with each release.

### Relevant log output

```shell
Current published image (base: python:3.11.2-slim, Debian 11 bullseye)

libsqlite3-0  CVE-2025-6965  HIGH  affected (no fix available on bullseye)
libsqlite3-0  CVE-2025-7458  HIGH  affected (no fix available on bullseye)

After moving to python:3.11-slim-trixie
(python:3.11-slim-trixie@sha256:be1575ed968de893bd54f4c56315ff7c4736ce522c1bca08fd521731aafc0d76)

Total: 19 (HIGH fixable: 3, CRITICAL: 0)
sqlite3 upgraded to 3.46.1-7+deb13u1; CVE-2025-6965 and CVE-2025-7458 no longer present.
Remaining 3 fixable HIGH are in openssl, jaraco.context, and wheel (unrelated to sqlite3).
```

chaowatson:
## Description

Moves both build stages from `python:3.11.2-slim` (Debian 11 bullseye) to `python:3.11-slim-trixie` (Debian 13). The bullseye base ships a `sqlite3` affected by `CVE-2025-6965` and `CVE-2025-7458` with no fix available on that Debian release. The trixie base carries `sqlite3` 3.46.1-7+deb13u1, which includes the fixes for both. Python stays at 3.11.

Fixes #1112.

## Verification

Scanned the trixie based image with Trivy: **19 total findings, 3 fixable HIGH, 0 CRITICAL**, and the two `sqlite3` CVEs no longer present. The remaining 3 fixable HIGH findings are in openssl, jaraco.context, and wheel, unrelated to this change. Image scanned:

```
python:3.11-slim-trixie@sha256:be1575ed968de893bd54f4c56315ff7c4736ce522c1bca08fd521731aafc0d76
```

## Notes

Using the floating trixie tag so routine rebuilds keep picking up base patches. If reproducible builds are preferred, this can be pinned by digest instead, ideally paired with Dependabot so the pin still advances. Happy to switch it to a pinned digest if maintainers would rather.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,7 +5,7 @@
 #
 # Stage 1: build packages and compile where needed
 #
-FROM python:3.11.2-slim AS build-stage
+FROM python:3.11-slim-trixie AS build-stage
 
 RUN apt-get -y update && \
     apt-get install -y curl git gcc lbzip2 pbzip2 pigz make jq && \
@@ -22,7 +22,7 @@ RUN cd opensearch-benchmark; \
 #
 # Stage 2: create image
 #
-FROM python:3.11.2-slim AS image-stage
+FROM python:3.11-slim-trixie AS image-stage
 ENV BENCHMARK_RUNNING_IN_DOCKER=True
 
 RUN groupadd --gid 1000 opensearch-benchmark && \


chaowatson:
The Code-Diff-Analyzer is hard-blocking it, but under a blanket rule that flags any FROM directive change, so there is no author side edit that clears it (it fired the same on the floating tag, the pinned patch tag, everything). It looks like it needs a maintainer to review and add the skip-diff-analyzer label, then re-run.

docker/Dockerfile
```diff
@@ -5,7 +5,7 @@
 #
 # Stage 1: build packages and compile where needed
 #
-FROM python:3.11.2-slim AS build-stage
+FROM python:3.11.16-slim-trixie AS build-stage
```
OVI3D0: Should we just use the minor-version floating tag in both stages instead:
```
FROM python:3.11-slim-trixie AS build-stage
FROM python:3.11-slim-trixie AS image-stage
```

so we continue to get the benefit of any future patch releases?

OVI3D0:
thanks @chaowatson ! just left one comment.

docker/Dockerfile
chaowatson: Good question. Quick context first: I actually started with the floating `python:3.11-slim-trixie` and switched to the pinned `3.11.16-slim-trixie` to try to get past the Code-Diff-Analyzer. That did not help. The analyzer flags any FROM change regardless of how it is pinned (floating, patch-pinned, and a full digest all trip it), so the PR is still blocked and will need a maintainer to review and apply the `skip-diff-analyzer` label. So the pinning choice and the block are independent, and we can pick whatever is best on the merits.

Float vs fixed:

Floating minor tag (`python:3.11-slim-trixie`)
- Pro: every rebuild picks up the latest 3.11 patch and Debian point-release security fixes automatically, which is exactly the benefit you mention and the spirit of this change.
- Con: builds are not reproducible, and a scan result can drift from what actually ships.

Fixed patch tag (`python:3.11.16-slim-trixie`)
- Pro: explicit and reproducible.
- Con: it freezes the Python patch, and once 3.11.17 ships this tag generally stops being rebuilt, so it slowly goes stale, which is the same class of problem this PR is fixing.

(A full digest pin is the most reproducible option but the least readable and needs Renovate/Dependabot to advance, so I would not push that here.)

Recommendation: since the goal is to stay patched and the project builds and publishes its own image, I agree with you, let's go with `python:3.11-slim-trixie` in both stages. If you later want reproducibility as well, the cleaner way is a CI image scan plus a published SBOM (I noted both as optional follow-ups in #1112) rather than freezing the tag. I will switch the PR back to the floating minor tag in both stages.

docker/Dockerfile
```diff
@@ -5,7 +5,7 @@
 #
 # Stage 1: build packages and compile where needed
 #
-FROM python:3.11.2-slim AS build-stage
+FROM python:3.11-slim-trixie AS build-stage
```
rishabh6788, line 8: can you please update it to `python:3.11.16-slim` to get the latest image on 3.11 version?

chaowatson: Sure, let me work on it!
Thanks for reviewing.

rishabh6788 (approved):

